PR #11, opened 2023-02-14
docs: boilerplate

supproduction:
adds boilerplate documentation

closes [HRZ-2151](https://spryker.atlassian.net/browse/HRZ-2151)

[HRZ-2151]: https://spryker.atlassian.net/browse/HRZ-2151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/drafts/orchestration/boilerplate.md
```diff
@@ -0,0 +1,211 @@
+# Storefront Boilerplate
+
```
tobi-or-not-tobi: In general we need to provide more explanation to the topic. 

I'd start with an introduction to explain what boilerplate is. 
> Boilerplate refers to the code that developers need to write in order to set up and configure the application before they can start building custom code. This includes the creating of the project directory structure, setup the dependencies and build the application entry point(s), such as index.html (CSR app) and index.js (SSR app).

Then we should outline the approach/vision of Oryx to boilerplate:
- Oryx tries to avoid boilerplate, as it is template code that cannot be easily updated to newer versions of Oryx
- All application logic is provided in packages which are distributed on npm. We should provide a link to the versioning strategy
- The feature sets, theming and configuration of the app is provided to the app builder
- Opinionated features are provided in presets. Presets can be added with a single line of code, and are provided to quickly setup a standard (demo) app. As soon as you start building a custom app, you might want to opt-out of the standard presets. Presets contain feature sets, themes and ...

We should provide the _ways_ to create the boilerplate code:
- manually, following steps below (I'm not too sure if we should do this, but we can. it means more maintenance on the docs)
- clone a repo (let's discuss in catch up on what repo we'll offer)
- (going forward we will provide a CLI to generate)

Then we describe what the boilerplate of Oryx entails and why:
- the bare minimum code: index.html, app.ts, ssr entry point (optional), ...

We could have a section for each of the main files and detail how they work, what they include.

tobi-or-not-tobi (dismissed):

AndreyYevtukhov:
1. PR title is not correct
2. Some checks failed
3. Conversations should be resolved

docs/boilerplate.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Introduction
+
+Boilerplate refers to the _template_ code that is used to generate an application code that can be further customized. Changing boilerplate is convenient at first sight as the code is at hand, generated in your project repository. However, when upgrading to newer versions of the original code, it becomes challenging. If you've customize the boilerplate code and want to reapply a new version of the boilerplate, you have to merge the customizations with the boilerplate during every update to a new version of the application. This is a time-consuming and error-prone process that can slow down your development process and increase the risk of bugs.
+
+## How we prevent boilerplate code
+
+In Oryx, we use a few tactics to prevent boilerplate code. These tactics include:
+
+1. bootstrap the application from npm packages rather than source.
```
tolerants, line 13:
```suggestion
1. Bootstrap the application from npm packages rather than source.
```

docs/boilerplate.md
```diff
@@ -0,0 +1,126 @@
...
+2. Expose an appBuilder function to setup the [app orchestrator](./app-orchestrator.md) conveniently.
+3. Provide [presets](./presets.md) for the standard application setup, including the feature sets and UI themes.
+4. Provide the source code in a public repository ([github.com/spryker/oryx](https://github.com/spryker/oryx)) to enable developers to customize the code. The code is implemented with TypeScript and the types and interfaces are documented where ever needed to provide a good developer experience.
```
tolerants, line 16: Maybe it's worth to describe how exactly it helps in preventing boilerplate code?

docs/boilerplate.md
```diff
@@ -0,0 +1,126 @@
...
+5. Allow for configuration of business logic and component behavior.
+6. Allow for customizations by [dependency injection](./dependency-injection.md).
+
+Most of the details around these tactics can be found elsewhere in the documentation. In the following sections you read more details on the boilerplate files that are needed.
+
+### NPM Packages
+
+One of the biggest sources of boilerplate code in any project is the application logic itself. In Oryx, we've separated out all the application logic into individual packages that are distributed on npm. These packages include components, business logic and integrations to the Spryker APIs.
+
+By separating out the application logic in this way, we've eliminated the boilerplate code from your projects. Instead of writing code to handle basic functionality, you can simply install the appropriate package and use it directly in your code. And because the packages are distributed on npm and published with Semantic Versioning, you can be confident that upgrading to a new version of the framework won't break your code.
```
tolerants, line 26:
```suggestion
By separating out the application logic in this way, we've eliminated the boilerplate code from your projects. Instead of writing code to handle basic functionality, you can simply install the appropriate package and use it directly in your code. And because the packages are distributed on npm and published with [Semantic Versioning](https://semver.org/), you can be confident that upgrading to a new version of the framework won't break your code.
```

docs/boilerplate.md
```diff
@@ -0,0 +1,126 @@
...
+│   └──server.ts
+```
+
+Here's a breakdown of the different bits of boilerplate code:
+
+### `package.json`
+
+The package.json file contains all the dependencies of a node js project. To simplify the dependency management, all dependencies are pulled through a single preset package. The preset package contains dependencies to all available oryx packages. This might not be the most optimal setup for your project over time, as it might contain a lot of "dead code", but for the start of your project it is very convenient. While you're getting experienced with Oryx over time, you could consider to build your own list of dependencies to get more fine-grained control. That being said, having unused dependencies in your project will not affect the build time or run time of your project. It is only an overhead during the npm install process.
+
+The bare minimum package.json is setup with the following package.json:
```
tolerants, line 62: Suggest to add a note that `vite` is not required and can be replaced with another frontend builder/bundler and added as an example here.

tolerants: It looks like we are missing vite configuration.

docs/boilerplate.md
```diff
@@ -0,0 +1,126 @@
...
+
+```ts
+import { appBuilder } from '@spryker-oryx/core';
+import { b2cFeatures } from '@spryker-oryx/presets';
+import { storefrontTheme } from '@spryker-oryx/themes';
+
+export const app = appBuilder()
+  .withFeature(b2cFeatures)
+  .withTheme(storefrontTheme)
+  .withEnvironment(import.meta.env)
```
tolerants, line 110: But there is no example and description for our default environment variable, where you wanted to put this part?

docs/boilerplate.md
```diff
@@ -0,0 +1,126 @@
...
+Here's a breakdown of the different bits of boilerplate code:
+
+### `package.json`
+
+The package.json file contains all the dependencies of a node js project. To simplify the dependency management, all dependencies are pulled through a single preset package. The preset package contains dependencies to all available oryx packages. This might not be the most optimal setup for your project over time, as it might contain a lot of "dead code", but for the start of your project it is very convenient. While you're getting experienced with Oryx over time, you could consider to build your own list of dependencies to get more fine-grained control. That being said, having unused dependencies in your project will not affect the build time or run time of your project. It is only an overhead during the npm install process.
+
+The bare minimum package.json is setup with the following package.json:
+
+```json
+{
```
tolerants, line 65: Maybe it's worth to add a command to run the server?

docs/boilerplate.md
```diff
@@ -0,0 +1,126 @@
...
+  .create();
+```
+
+In the boilerplate we see the following things:
+
+- the usage of a standard _feature set_, which is provided by the `presets` package
+- the usage of a standard theme, that applies a global color scheme and other theme design tokens to the application
+- the usage of environment variables, to provide the backend base URLs to the application
+
+### `server/render.ts` (optional)
```
tolerants, line 120: Hm, we are describing the file but not even showing the content, what is the purpose of having it?

docs/boilerplate.md
tobi-or-not-tobi: I'll give it a try, but not sure if it helps.

docs/boilerplate.md
tobi-or-not-tobi: I'd expect that to be covered in the getting started section.

docs/boilerplate.md
tobi-or-not-tobi: as discussed in setup guide PR, we'll create it in environment docs

docs/boilerplate.md
tobi-or-not-tobi: the point was to only show the dependencies of the package.json. I've added a few dots before and after to ensure this is not seen as the full file.

tobi-or-not-tobi:
Closing as we moved this documentation to the docs repo, see https://github.com/spryker/spryker-docs/pull/1684

PR #14, opened 2023-02-15
fix: error handling

galitskye:
closes [HRZ-2130](https://spryker.atlassian.net/browse/HRZ-2130)


libs/domain/site/src/services/error-handling/error-handler.ts
```diff
@@ -3,17 +3,46 @@ import { inject } from '@spryker-oryx/di';
 import { Types } from '@spryker-oryx/ui/notification';
 import { NotificationService } from '../notification';
 
+interface Error {
```
dunqan: let;s rename it to API error

dunqan (approved):

PR #16, opened 2023-02-15
feat: add netlify.toml file with ssr config

AndreyYevtukhov:
Fixes SSR build in Netlify

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,3 @@
+[functions.ssr]
+  node_bundler = "esbuild"
+  included_files = ["dist/apps/storefront/**"]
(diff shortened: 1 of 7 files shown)


tolerants (dismissed):

netlify.toml
```diff
@@ -0,0 +1,3 @@
+[functions.ssr]
+  node_bundler = "esbuild"
+  included_files = ["dist/apps/storefront/**"]
```
tobi-or-not-tobi, line 3: not enough:
- redirect is needed to ensure all deep links redirect to index.html
- security headers would be nice to have here, as it will help us to understand potential integration issues upfront while we're building features

tobi-or-not-tobi (changes requested):

netlify.toml
AndreyYevtukhov: Added missed files and package

tobi-or-not-tobi (approved):

PR #18, opened 2023-02-15
fix: upgrade lit to ^2.6.0

aleespryker:
Closes [HRZ-2225](https://spryker.atlassian.net/browse/HRZ-2225)


libs/base/ui/overlays/drawer/src/component.ts
```diff
@@ -4,7 +4,7 @@ import { TAG_NAME } from './tag';
 export const drawerComponent = componentDef({
   name: TAG_NAME,
   impl: () =>
-    window.HTMLDialogElement
```
dunqan, line 7: let's make it explicit by adding `isServer` check to the mix (but keep current behavior)

libs/base/ui/overlays/modal/src/component.ts
```diff
@@ -3,7 +3,7 @@ import { componentDef } from '@spryker-oryx/core';
 export const modalComponent = componentDef({
   name: 'oryx-modal',
   impl: () =>
-    window.HTMLDialogElement
```
dunqan, line 6: the same as above

dunqan (dismissed):

libs/base/ui/overlays/modal/src/component.ts
aleespryker: Using `globalThis` for now as discussed because of unit tests. Solution may appear later in another ticket.

dunqan (approved):

PR #20, opened 2023-02-15
fix: sorting hydration 

nsilva-spryker:
Closes [HRZ-2229](https://spryker.atlassian.net/browse/HRZ-2229)

### DIFF
--- a/libs/domain/search/sort/src/sort.component.ts
+++ b/libs/domain/search/sort/src/sort.component.ts
@@ -1,5 +1,4 @@
 import { resolve } from '@spryker-oryx/di';
-import { ProductListSort } from '@spryker-oryx/product';
 import { RouterService } from '@spryker-oryx/router';
 import {
   asyncState,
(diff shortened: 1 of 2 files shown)


libs/domain/search/sort/src/sort.component.ts
```diff
@@ -29,6 +30,8 @@ export class SortComponent extends LitElement {
   protected renderSorting(
     sort: ProductListSort | undefined | null
```
galitskye: My suggestion to make check before invoke `renderSorting` method. To avoid `undefined | null`

galitskye (changes requested):

libs/domain/search/sort/src/sort.component.ts
nsilva-spryker: I keep it even more simple by using the sortingOptions directly in the method

libs/domain/search/sort/src/sort.component.ts
```diff
@@ -24,13 +23,11 @@ export class SortComponent extends LitElement {
...
-    return this.renderSorting(this.sortingOptions);
+    return this.renderSorting();
   }
 
-  protected renderSorting(
-    sort: ProductListSort | undefined | null
-  ): TemplateResult {
-    const hasOptions = !!sort?.sortValues.length;
+  protected renderSorting(): TemplateResult {
+    const hasOptions = !!this.sortingOptions?.sortValues.length;
```
galitskye: If I correct understand `sortValues` may be undefined or array. By trying take length from `undefined` will be an error.

nsilva-spryker: The expression is checking whether sortingOptions is true and whether it has a non-zero sortValues array length.
If the sortValues array is empty or null, the expression returns 0. The current code could be sufficient

galitskye (approved):

dunqan (approved):

PR #25, opened 2023-02-15
chore: video UI element

tobi-or-not-tobi:
Move video element down to the UI library, so that it is directly reusable in other components.


libs/base/ui/graphical/video/src/video.component.ts
```diff
@@ -0,0 +1,85 @@
...
+  @property({ type: Boolean }) muted?: boolean;
+  @property({ type: Boolean }) playsInline?: boolean;
+  @property() preload?: VideoPreload;
+  @property() set aspectRatio(value: VideoAspectRatio) {
+    this.style.setProperty('--aspect-ratio', value);
+    this.style.setProperty('height', value ? '' : 'var(--height)');
+  }
+
+  protected override render(): TemplateResult | void {
+    console.log('muted', this.muted);
```
Sinucid: console.log

libs/base/ui/graphical/video/src/video.component.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  protected isYoutube(): boolean {
+    return !!this.url && /youtu.be|youtube/g.test(this.url);
+  }
+
+  protected resolveYoutubeUrl(): string | undefined {
+    if (!this.url) return;
+
+    const resolveId = (url: string): string | null | undefined => {
```
Sinucid: This func returns `string | null | undefined` and then its result uses in template string that build the url. `null` and `undefined` is converting to `'null' `and `'undefined'`, is this an expected behavior?

libs/base/ui/graphical/video/src/video.component.spec.ts
```diff
@@ -0,0 +1,209 @@
+import { fixture } from '@open-wc/testing-helpers';
+import { useComponent } from '@spryker-oryx/core/utilities';
+import { html } from 'lit';
+import VideoComponent from './video.component';
+import { videoComponent } from './video.def';
+
+describe('VideoComponent', () => {
```
Sinucid, line 7: I believe that we need also test that changing of ascpectRatio prop sets the css properties

Sinucid (changes requested):

Sinucid (dismissed):

Sinucid (approved):

PR #27, opened 2023-02-16
fix: more detailed field validation

Sinucid:
Pattern validation feature for required text-like inputs

closes to [HRZ-2227](https://spryker.atlassian.net/browse/HRZ-2227)

[HRZ-2227]: https://spryker.atlassian.net/browse/HRZ-2227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/form/src/models/form.model.ts
```diff
@@ -25,7 +25,6 @@ export const enum FormFieldType {
 export interface FormFieldDefinition<K = string> {
```
supproduction: extends Partial<FieldValidationPattern>

libs/base/form/src/renderers/default-form.renderer.ts
```diff
@@ -5,13 +5,25 @@ import { classMap, ClassMapDirective } from 'lit/directives/class-map.js';
...
 import { FormRenderer } from './form.renderer';
 import { FormFieldRenderer } from './renderer';
 
 export class DefaultFormRenderer implements FormRenderer {
   constructor(protected injector = inject(INJECTOR)) {}
 
+  fieldValidationPattern(field: FormFieldDefinition): FieldValidationPattern {
+    return {
+      pattern: field.pattern ?? '.*\\S+.*',
+      title: field.title ?? 'at least one character',
```
supproduction: Do we need this `at least one` if field is not required?

libs/base/form/src/renderers/default-form.renderer.ts
```diff
@@ -5,13 +5,25 @@ import { classMap, ClassMapDirective } from 'lit/directives/class-map.js';
...
+  FormFieldType,
+  FormValues,
+} from '../models';
 import { FormRenderer } from './form.renderer';
 import { FormFieldRenderer } from './renderer';
 
 export class DefaultFormRenderer implements FormRenderer {
   constructor(protected injector = inject(INJECTOR)) {}
 
+  fieldValidationPattern(field: FormFieldDefinition): FieldValidationPattern {
```
supproduction, line 21: Also do we want to expose this method as public API?

supproduction (changes requested):

libs/base/form/src/renderers/default-form.renderer.ts
Sinucid: Optional field should not also contain spaces only

libs/base/form/src/renderers/default-form.renderer.ts
Sinucid: Yeah, I believe so. For cases when we are building custom fields like country select in address form component.

libs/base/form/src/renderers/default-form.renderer.ts
supproduction: So basically we have validation for optional field. Is it expected behavior? cc @tobi-or-not-tobi

supproduction (dismissed):

libs/base/form/src/renderers/default-form.renderer.ts
tobi-or-not-tobi: agreed, if its a required field we rather stick to standard validation message provided by required field -  if thats what we mean here

tobi-or-not-tobi (changes requested):
would also be nice to get ride of the gh warnings

libs/base/form/src/renderers/default-form.renderer.ts
Sinucid: I didn't get. Validation for required text fields works only when no value provided, in other case (when has value) - pattern check will work

The same for optional fields. If it has value - pattern will check it. 

Is this a correct behavior?

AndreyYevtukhov:
1. `dev` should be merged here
2. No approve

libs/base/form/src/renderers/default-form.renderer.ts
tobi-or-not-tobi: Ok, lets make the call: we add the validation for required fields only.

libs/base/form/src/renderers/default-form.renderer.ts
```diff
@@ -5,13 +5,26 @@ import { classMap, ClassMapDirective } from 'lit/directives/class-map.js';
...
+    const { required, pattern, title } = field;
+    return {
+      pattern: pattern ?? (required ? '.*\\S+.*' : undefined),
+      title: title ?? (required ? 'at least one character' : undefined),
```
tobi-or-not-tobi:
```suggestion
      title: title ?? (required ? i18n('form.validation.invalid-empty-text') : undefined),
```

libs/base/form/src/mixins/form.mixin.ts
```diff
@@ -19,6 +19,7 @@ export declare class FormMixinInterface<FormValues = unknown>
 export const FormMixin = <
   // TODO: remove any and pass generic type to mixin
   // when we fix issue with types inference of mixins chain
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
tobi-or-not-tobi, line 22: can we make it `unknown`?

tobi-or-not-tobi (changes requested):

libs/base/form/src/mixins/form.mixin.ts
Sinucid: Unfortunately no, because `unknown` required providing some type to be able to access to the typed fields and providing some specific type for the parameter brake the mixin classes chain. We have such issue and it described above.

tobi-or-not-tobi (dismissed):

tobi-or-not-tobi (approved):

PR #29, opened 2023-02-16
feat: Release workflow

tolerants:
- Fixed existing release workflow
- Created new release workflow from packages
- Fixed release script and lerna config
- Added a check if build is successful for test workflow

closes https://spryker.atlassian.net/browse/HRZ-2223


.github/workflows/test.yml
```diff
@@ -102,3 +102,17 @@ jobs:
           cmd: npx nx run storybook:visual-regression-branch
           affectedCmd: npx nx affected --target=visual-regression-pr
           isDev: ${{ env.IS_DEV }}
+  test-build:
```
tolerants, line 105: It's a suggestion, as long as for now we are not checking that build is working.

AndreyYevtukhov: good idea, let's have it. We can also add it in branch config to be a required check in the PR

AndreyYevtukhov (dismissed):

libs/LICENSE
```diff
@@ -0,0 +1,40 @@
+SPRYKER SYSTEMS GMBH EVALUATION LICENSE AGREEMENT
```
gund: Maybe we can move this file to the root so it's clear to everyone what license this repo is under?

tolerants: There is already LICENSE file with the same content in the root, but I wanted to separate them initially, so release process of `libs` will rely on the file in `libs` folder.

gund: But the point is to have one license for all and have it in root so ppl can easily assess the license and have no surprises that the license in packages is different from the license in the root.

tolerants: Ok, good point. Removed it.

gund (approved):

PR #36, opened 2023-02-17
feat: unsafeStatic hack

aleespryker:
Closes HRZ-2280

Still WIP; don't merge until HRZ-2225 is merged.


AndreyYevtukhov:
@aleespryker conflicts with `dev`

aleespryker:
> @aleespryker conflicts with `dev`

Fixed

libs/domain/auth/src/services/components.provider.ts
```diff
@@ -1,24 +1,7 @@
...
-    },
-    ['auth-logout']: {
-      template: (uid: string, layoutClasses?: string) =>
-        html`<auth-logout
-          uid="${uid}"
-          class=${ifDefined(layoutClasses)}
-        ></auth-logout>`,
-    },
-  },
+  useValue: true,
```
dunqan: we should just drop it

libs/domain/cart/src/services/components.provider.ts
```diff
@@ -1,38 +1,7 @@
...
-    },
-    ['cart-summary']: {
-      template: (uid: string, layoutClasses?: string) =>
-        html`<oryx-cart-summary
-          uid="${uid}"
-          class=${ifDefined(layoutClasses)}
-        ></oryx-cart-summary>`,
-    },
-  },
+  useValue: true,
```
dunqan: Let's remove all those providers

libs/platform/experience/src/services/registry/default-components-registry.service.ts
```diff
@@ -32,7 +32,14 @@ export class DefaultComponentsRegistryService
     uid: string,
     styleClasses?: string
   ): TemplateResult | undefined {
-    const component = this.componentMapping?.[type];
+    let component;
```
dunqan: component mappig should have precedence, and only if there is no explicit mapping defined, we want to reach to component plugin, to check, if it provides this component.

dunqan: but actually, let's simplify:
- don't check component plugin at all (bring back protected method)
- just render the type as tag always, even if we don't have a mapping

dunqan (changes requested):

libs/platform/core/src/orchestration/components/components-loader.ts
```diff
@@ -42,7 +42,7 @@ export class ComponentsLoader {
     }
   }
 
-  protected findComponentDefBy(name: string): ComponentDef {
+  findComponentDefBy(name: string): ComponentDef {
```
dunqan: can we revert this?

libs/template/presets/src/features/b2c/experience/header.ts
```diff
@@ -65,9 +65,8 @@ export const HeaderTemplate: StaticComponent = {
           },
         },
         { type: 'oryx-user-summary' },
-        { type: 'auth-logout' },
```
dunqan, line 68: is it deliberate?

libs/platform/experience/src/services/registry/default-components-registry.service.ts
```diff
@@ -13,13 +13,12 @@ export class DefaultComponentsRegistryService
   protected componentMapping?: ComponentMapping;
 
   constructor(
-    registeredComponents = inject(ComponentMapping),
+    registeredComponents = inject(ComponentMapping, []),
```
dunqan: let's use `[] as ...` to fix typings instead

dunqan (changes requested):

libs/template/presets/src/features/b2c/experience/header.ts
aleespryker: We are using `oryx-auth-logout` now. Also noticed that `oryx-user-summary` includes the logout component anyways, so auth-logout is not needed. I asked @tobi-or-not-tobi about it earlier.

dunqan (approved):

PR #38, opened 2023-02-17
fix: video muted state behaviour

Sinucid:
Regarding w3 specification toggling of muted attribute doesn't change the property. It sets the default value only. Due to this need to toggle property manually


libs/base/ui/graphical/video/src/video.component.ts
```diff
@@ -24,6 +24,17 @@ export default class VideoComponent
...
+  protected updated(_changedProperties: PropertyValues): void {
+    super.updated(_changedProperties);
+
+    const video = this.renderRoot.querySelector<HTMLVideoElement>('video');
+
+    if (video) {
+      video.toggleAttribute('muted', !!this.muted);
+      video.muted = !!this.muted;
+    }
+  }
```
tobi-or-not-tobi: but now it requires hydration of this component, right?

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #41, opened 2023-02-18
feat: add context fallback

supproduction:
adds context fallback feature 

closes [HRZ-2248](https://spryker.atlassian.net/browse/HRZ-2248)


workspace.json
```diff
@@ -12,7 +12,6 @@
     "form": "libs/base/form",
     "i18n": "libs/platform/i18n",
     "indexed-db": "libs/platform/indexed-db",
-    "labs": "libs/template/labs",
```
dunqan: I think it can be reverted now, unless there is a another reason for it.

libs/platform/core/src/services/context/default-context.service.ts
```diff
@@ -100,6 +102,22 @@ export class DefaultContextService implements ContextService {
...
   }
 
+  protected contextFallback<T>(
+    token: string
+  ): (observable$: Observable<T>) => Observable<T> {
+    return (observable$) =>
+      observable$.pipe(
+        switchMap((value) =>
+          value === undefined
+            ? resolve<Observable<T>>(
```
dunqan: We should use injector here.

dunqan (changes requested):

dunqan (approved):

PR #45, opened 2023-02-20
[Docs] App orchestrator

gund:
Closes [HRZ-2150](https://spryker.atlassian.net/browse/HRZ-2150)

Also simplified AppBuilder interfaces.

NOTE: Link to Providers docs is not provided as there is no docs for it yet.

https://github.com/spryker/oryx/blob/687c8ee091d5f218fdaefedcecd6f010c9f503d7/docs/drafts/app-orchestrator/index.md?plain=1#L1-L39

[HRZ-2150]: https://spryker.atlassian.net/browse/HRZ-2150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/drafts/app-orchestrator/app-builder.md
+++ b/docs/drafts/app-orchestrator/app-builder.md
@@ -1,0 +1,34 @@
+# AppBuilder
+
+`AppBuilder` allows to compose different pieces into a single Oryx application.
+
+It is a pluggable builder which supports the following built-ins:
+
+- Options (`withAppOptions`/`withOptions`)
+- AppFeatures (`withFeature`)
+- Components (`withComponents`)
+- Providers (`withProviders`)
+- Themes (`withTheme`)
+- Environment (`withEnvironment`)
+- Resources (`withResources`)
+- Plugins (`with`)
+
+All of the above are built-in plugins but you can also
+add your custom plugins by using `with` API:
+
+```ts
+import { appBuilder, AppPlugin } from '@spryker-oryx/core';
+
+appBuilder().with(class MyPlugin implements AppPlugin {});
+```
+
+After all of the pieces have been registered you can create an app
+by using `create` API which will return you a promise of the app:
+
+```ts
+import { appBuilder } from '@spryker-oryx/core';
+
+const app = appBuilder().create().catch(console.error);
+```
+
+See [`App`](/docs/drafts/app-orchestrator/app.md) to learn more about it's usage.
--- a/docs/drafts/app-orchestrator/app.md
+++ b/docs/drafts/app-orchestrator/app.md
@@ -1,0 +1,9 @@
+# App
+
+`App` represents a running Oryx application.
+
+It allows you to:
+
+- Interact with registered plugins during the construction phase (`findPlugin`/`requirePlugin`)
+- Wait for application ready state (`whenReady`)
+- Destroy and cleanup application (`destroy`)
--- a/docs/drafts/app-orchestrator/index.md
+++ b/docs/drafts/app-orchestrator/index.md
@@ -1,0 +1,39 @@
+# App orchestrator
+
+Any Oryx application starts with the orchestrator.
+It allows to compose your application from reusable bits and pieces such as
+features, components, providers, themes, etc.
+
+To use app orchestrator import `AppBuilder` from the `@spryker-oryx/core` and
+start composing your own set of features with it:
+
+```ts
+import { appBuilder } from '@spryker-oryx/core';
+import { b2cFeatures, b2cTheme } from '@spryker-oryx/presets';
+
+const app = appBuilder()
+  .withFeature(b2cFeatures)
+  .withTheme(b2cTheme)
+  .withEnvironment(import.meta.env);
+```
+
+Once you are satisfied with your app configuration it's time to create it:
+
+```ts
+app.create().catch(console.error);
+```
+
+Now you have fully up and running Oryx app tailored to your needs!
+
+---
+
+To learn more about details of app orchestrator see:
+
+- [AppBuilder](/docs/drafts/app-orchestrator/app-builder.md)
+- [App](/docs/drafts/app-orchestrator/app.md)
+- [AppFeature](/docs/drafts/app-orchestrator/app-feature.md)
+- [AppEnvironment](/docs/drafts/app-orchestrator/app-environment.md)
+- [Components](/docs/drafts/components/index.md)
+- [Providers](TODO: Link to providers)
+- [Resources](/docs/drafts/app-orchestrator/resources.md)
+- [Theme](/docs/drafts/theme/configuration.md)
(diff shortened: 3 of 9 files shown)


docs/drafts/app-orchestrator/app-builder.md
```diff
@@ -0,0 +1,34 @@
+# AppBuilder
+
+`AppBuilder` allows to compose different pieces into a single Oryx application.
+
+It is a pluggable builder which supports the following built-ins:
+
+- Options (`withAppOptions`/`withOptions`)
```
supproduction, line 7:
```suggestion
- Options (`withAppOptions`)
- FeatureOptions (`withOptions`)
```

libs/platform/core/src/orchestration/app/modular-app-builder.ts
```diff
@@ -106,21 +102,19 @@ export class ModularAppBuilder extends SimpleAppBuilder<AppBuilderWithModules> {
...
     if (this.providers.length) {
-      this.plugins.unshift(
-        new InjectionPlugin(this.providers, this.options?.injector)
-      );
+      this.with(new InjectionPlugin(this.providers, this.options?.injector));
     }
 
     if (this.resources) {
-      this.plugins.unshift(new ResourcePlugin(this.resources));
+      this.with(new ResourcePlugin(this.resources));
```
supproduction, line 109: We wanted that these plugins have first order of initialization

supproduction (changes requested):

docs/drafts/app-orchestrator/index.md
```diff
@@ -0,0 +1,39 @@
+# App orchestrator
+
+Any Oryx application starts with the orchestrator.
```
tobi-or-not-tobi, line 3: I'd like to have a more general introduction into this topic before diving into the code. Maybe in general we do this:
1. what is it? a one liner or a small section that explains the concept
2. why do I need it? A section that describes the various use cases.  This also describes for who this is. 
3. when do I need it? This could be part of 2 I guess, but in this case it's definitely nice to have as separate section.
4. how do I use it? here we can provide a bit of code
5. what is there more to learn? how does it relate to other topics?

Ad 2 – Let's put a bit more explanation on why we need an orchestrator. Customisations?decide what to load when at runtime? ability to change the app at runtime? It should help to understand:
- the app Builder script is the minimum boilerplate that Oryx provides to ensure easy upgrades to later versions
- provides a flexible setup to configure standard with customisations
- the effect on the build, what pieces are part of a the app, despite the features available in the packages
- the effect on runtime behaviour; not all bundled chunks are loaded, since features are lazily loaded
- how the environment is associated to the app
- how a global theme is applied to the components

Ad 3 – app orchestrations is used to setup the project, and will typically remain fairly stable during development. It is recommended to create your own feature set, so that you can extend features over time without changing the orchestration setup. 

Ad 5 – this is where I'd add links to other related sections

docs/drafts/app-orchestrator/index.md
```diff
@@ -0,0 +1,39 @@
...
+const app = appBuilder()
+  .withFeature(b2cFeatures)
+  .withTheme(b2cTheme)
+  .withEnvironment(import.meta.env);
+```
+
+Once you are satisfied with your app configuration it's time to create it:
+
+```ts
+app.create().catch(console.error);
```
tobi-or-not-tobi, line 23: we we really want to use the catch here? What does it bring? Would you keep it in production?

docs/drafts/app-orchestrator/index.md
```diff
@@ -0,0 +1,39 @@
+# App orchestrator
+
+Any Oryx application starts with the orchestrator.
+It allows to compose your application from reusable bits and pieces such as
+features, components, providers, themes, etc.
+
+To use app orchestrator import `AppBuilder` from the `@spryker-oryx/core` and
```
tobi-or-not-tobi, line 7: I'd emphasize this introduction on a description on why we have this thing and what it does without the need for code. Then, for sure, we can have a section that shows how to "Setup an application"

docs/drafts/app-orchestrator/index.md
```diff
@@ -0,0 +1,39 @@
+# App orchestrator
```
tobi-or-not-tobi, line 1: what about: 

Orchestrate an Oryx Application

docs/drafts/app-orchestrator/index.md
tobi-or-not-tobi: we could then link to the `appBuilder` section for more info.

docs/drafts/app-orchestrator/index.md
```diff
@@ -0,0 +1,39 @@
...
+To learn more about details of app orchestrator see:
+
+- [AppBuilder](/docs/drafts/app-orchestrator/app-builder.md)
+- [App](/docs/drafts/app-orchestrator/app.md)
+- [AppFeature](/docs/drafts/app-orchestrator/app-feature.md)
+- [AppEnvironment](/docs/drafts/app-orchestrator/app-environment.md)
+- [Components](/docs/drafts/components/index.md)
+- [Providers](TODO: Link to providers)
+- [Resources](/docs/drafts/app-orchestrator/resources.md)
+- [Theme](/docs/drafts/theme/configuration.md)
```
tobi-or-not-tobi, line 39: I'd make them more actionable rather then provide dry knowledge, e.g.:
- [Configure feature set](/docs/drafts/app-orchestrator/app-feature.md)
- [Setup the environment](/docs/drafts/app-orchestrator/app-environment.md)
- [Configure custom components](/docs/drafts/components/index.md)
- etc

docs/drafts/app-orchestrator/app.md
```diff
@@ -0,0 +1,9 @@
+# App
+
+`App` represents a running Oryx application.
```
tobi-or-not-tobi, line 3: Can we get an example on when this is useful? Are we using this ourselves?

docs/drafts/app-orchestrator/app-builder.md
```diff
@@ -0,0 +1,34 @@
+# AppBuilder
+
+`AppBuilder` allows to compose different pieces into a single Oryx application.
```
tobi-or-not-tobi, line 3: can we introduce the changing concept here?

tobi-or-not-tobi (changes requested):

libs/platform/core/src/orchestration/app/modular-app-builder.ts
gund: But it's not correct to rely on the order of plugins registration, we have special hooks for that (before/after apply) which should be used instead to establish correct order.
I will have a look into those plugins to correct the order but we should not couple registration order with execution order.

docs/drafts/app-orchestrator/app-builder.md
gund: What do you mean by "changing concept"?

tobi-or-not-tobi: I meant _chaining_

AndreyYevtukhov:
1. PR title is not valid
2. Conversations should be resolved
3. Some check fail (development should be merged here)

gund:
This PR is moved to https://github.com/spryker/spryker-docs/pull/1723

PR #47, opened 2023-02-21
fix: re-render of address list

Sinucid:
Checked attribute sets only the default state and doesn't affect on property, but LIT is able to manages it under the hood and boolean checked attribute changes the prop. But for some cases, like deletion of currently checked radio button, this alignments doesn't work. To overcome this issue `keyed` directive is using for address list component, to trigger re-render and align attribute and property correctly 

closes to [HRZ-2237](https://spryker.atlassian.net/browse/HRZ-2237)


libs/domain/user/address-list/src/address-list.component.ts
```diff
@@ -46,9 +49,17 @@ export class AddressListComponent extends ComponentMixin<AddressListItemOptions>
     distinctUntilChanged()
   );
 
-  protected addresses$ = combineLatest([
-    resolve(AddressService).getAddresses().pipe(distinctUntilChanged()),
-    new ContentController(this).getOptions(),
+  protected addresses$ = this.addressService
+    .getAddresses()
+    .pipe(distinctUntilChanged());
```
supproduction: `distinctUntilChanged` doesn't make sense for object

libs/domain/user/address-list/src/address-list.component.ts
```diff
@@ -46,9 +49,17 @@ export class AddressListComponent extends ComponentMixin<AddressListItemOptions>
     distinctUntilChanged()
```
supproduction: Why do we need `distinctUntilChanged` here?

supproduction (changes requested):

libs/domain/user/address-list/src/address-list.component.ts
Sinucid: Yeah, my bad. I refactored it but didn't pay attention on observables logic

libs/domain/user/address-list/src/address-list.component.ts
Sinucid: Not necessary, dropped

supproduction (approved):

tobi-or-not-tobi (approved):

PR #55, opened 2023-02-21
fix: show all order entries when number of items under threshold

aleespryker:
Closes [HRZ-2300](https://spryker.atlassian.net/browse/HRZ-2300)


libs/domain/order/entries/src/entries.component.ts
```diff
@@ -35,11 +35,13 @@ export class OrderEntriesComponent extends OrderMixin(
 
   protected hasThreshold$ = combineLatest([this.order$, this.options$]).pipe(
     map(([order, options]) => {
-      if (!order || !options.limit || !options.threshold) {
+      if (!order || !options.limit || typeof options.threshold !== 'number') {
```
Sinucid: I didn't get why do you need to check the type here? If take into account the model, threshold can be number or undefined. I believe that `threshold = 0` is equal `threshold = undefined` and `!options.threshold` should be enough

libs/domain/order/entries/src/entries.component.ts
```diff
@@ -35,11 +35,13 @@ export class OrderEntriesComponent extends OrderMixin(
...
         return false;
       }
-
-      return order.items.length < options.limit + options.threshold;
+      const hasThreshold =
+        order.items.length < options.limit + options.threshold;
+      this.showAllEntries = hasThreshold ? true : this.showAllEntries;
```
Sinucid: not necessary to change this property here. Just change the check:

```
         this.hasThreshold || this.showAllEntries
              ? this.order.items.length
              : this.componentOptions?.limit`
```

Sinucid (changes requested):

libs/domain/order/entries/src/entries.component.ts
aleespryker: Because we do actually consider 0 to be a valid threshold value, and need the calculation after it or else it will show the button when it shouldn't. Can happen in Storybook.

Sinucid: `!isNaN(options.threshold)`

aleespryker: Ended up with `isNaN(options.threshold as any)`.

We still want to continue if it is a number.

Sinucid (approved):

PR #56, opened 2023-02-21
fixes: hydration mismatch issue

supproduction:
fixes hydration mismatch issue

closes [HRZ-2306](https://spryker.atlassian.net/browse/HRZ-2306)

[HRZ-2306]: https://spryker.atlassian.net/browse/HRZ-2306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/utilities/src/decorators/hydratable/hydratable.ts
```diff
@@ -101,5 +109,13 @@ function hydratableClass<T extends Type<HTMLElement>>(
       this.removeAttribute('defer-hydration');
       prototype.connectedCallback.call(this);
     }
+
+    render(): TemplateResult {
+      if (!state) {
```
dunqan: we would like to also detect, if there is a ssr render available. If not, we don't have to wait.

libs/base/utilities/src/decorators/hydratable/hydratable.ts
```diff
@@ -101,5 +109,13 @@ function hydratableClass<T extends Type<HTMLElement>>(
...
+        return super.render();
+      }
+
+      return html`${when(this[state as keyof this], () => super.render())}`;
```
dunqan: We probably should use our small custom directive, with explicit `noChange` return,

libs/base/utilities/src/decorators/hydratable/hydratable.ts
```diff
@@ -101,5 +105,37 @@ function hydratableClass<T extends Type<HTMLElement>>(
...
+        (Boolean(this.renderRoot) || this.safariRoot()) && states && !isServer;
+
+      if (hasSsr) {
+        return html`${whenState(
+          states === JSON.stringify(this[asyncStates]),
+          () => super.render()
+        )}`;
+      }
+
+      return html`${whenState(true, () => super.render())}`;
```
dunqan: can't we just return `super.render()`?

libs/base/utilities/src/decorators/async-state/async-state.controller.ts
```diff
@@ -72,6 +79,16 @@ export class AsyncStateController implements ReactiveController {
 
     data.subscription = data.observable$.subscribe((value) => {
       const oldValue = data.value;
+
+      this.host[asyncStates]![key] = true;
+
+      if (isServer) {
+        this.host.setAttribute(
```
dunqan: if all attrs has to be true, let's try to drop this attribute

dunqan (dismissed):

dunqan (approved):

PR #57, opened 2023-02-22
chore: simplify address list implementation

tobi-or-not-tobi:
drop RxJS complexity from components


libs/domain/user/address-list/src/address-list.component.ts
```diff
@@ -28,85 +29,35 @@ export class AddressListComponent extends ContentMixin<AddressListItemOptions>(
...
-      if (address) {
-        this.dispatchEvent(
-          new CustomEvent(SELECT_EVENT, {
-            bubbles: true,
-            composed: true,
-            detail: { address },
-          })
+  protected addressesSubscription = this.addressService.getAddresses().pipe(
+    tap((addresses) => {
+      if (!this.componentOptions?.selectable) return;
```
Sinucid: Much better, but still. This subscription will trigger only when addresses change, the same situation if for some reason addresses will be fetched earlier then options, or if options change after, it has no effect. Unfortunately need to combine them to keep stable behaviour

Sinucid (changes requested):

libs/domain/user/address-list/src/address-list.component.ts
```diff
@@ -9,138 +9,130 @@ import {
...
+    return html`<slot name="empty">
       <oryx-icon type="location" size="large"></oryx-icon>
       ${i18n('user.address.no-addresses')}
     </slot>`;
   }
 
-  protected createAddressesKey(): string | void {
-    return this.addresses?.map(({ id }) => id).join('-');
+  protected select(): void {
+    const hasSelectedAddress = this.addresses?.find(
```
Sinucid:
```
if ( this.addresses?.find(
      (address) => address.id === this.selectedAddressId)) return;

this.selectedAddressId =
        this.addresses?.find((a) => this.isDefault(a))?.id ??
        this.addresses?.[0].id;
```

Sinucid (changes requested):

Sinucid (dismissed):

Sinucid (approved):

PR #62, opened 2023-02-23
feat: refactor product loading with queries

dunqan:
- Introduces base Query and Command functionality
- Refactor product services with Query usages


libs/domain/product/src/services/adapter/default-product-list.adapter.ts
```diff
@@ -55,13 +55,20 @@ export class DefaultProductListAdapter implements ProductListAdapter {
...
 
   get(qualifier: ProductListQualifier): Observable<ProductList> {
+    const include = [
+      ApiProductModel.Includes.AbstractProducts,
+      ApiProductModel.Includes.ConcreteProducts,
+      ApiProductModel.Includes.ConcreteProductImageSets,
+      ApiProductModel.Includes.ConcreteProductPrices,
+      ApiProductModel.Includes.ConcreteProductAvailabilities,
+      ApiProductModel.Includes.Labels,
+    ].filter((type, index, arr) => arr.indexOf(type) === index);
```
supproduction: I am wondering, why do we need this filter?

libs/platform/core/src/services/query/core/core-query.ts
```diff
@@ -0,0 +1,233 @@
...
+    protected destroyNotifier$?: Observable<undefined>
+  ) {}
+
+  protected getKey(qualifier: Qualifier): string {
+    return qualifier
+      ? JSON.stringify(
+          Object.entries(qualifier)
+            .filter((x) => x[1])
+            .sort()
+            .flat()
```
supproduction, line 55: mb we should change order?

libs/platform/core/src/services/query/core/core-query.ts
```diff
@@ -0,0 +1,233 @@
...
+  }: {
+    data: ValueType;
+    qualifier: Qualifier;
+    optimistic?: boolean;
+  }): void {
+    const subject$ = this.queryCache.get(this.prepareKey(qualifier))!.subject$;
+    subject$.next({
+      error: subject$.value.error,
+      loading: subject$.value.loading,
+      stale: optimistic ? true : subject$.value.stale,
```
supproduction, line 191:
```suggestion
      stale: optimistic || subject$.value.stale,
```

libs/platform/core/src/services/query/default-query.service.ts
```diff
@@ -0,0 +1,68 @@
...
+  QueryService,
+  QUERY_TOKEN,
+} from './query.service';
+
+export class DefaultQueryService
+  extends CoreQueryService
+  implements QueryService, OnDestroy
+{
+  constructor(
+    effects = inject(EFFECT_TOKEN, []),
```
supproduction: Should we mark it as protected?

libs/platform/core/src/services/query/default-query.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    protected injector = inject(INJECTOR)
+  ) {
+    super();
+    this.initEffects(effects);
+  }
+
+  protected initEffects(effects: any[]): void {
```
supproduction: Can omit any?

supproduction (changes requested):

libs/domain/product/src/services/adapter/default-product-list.adapter.ts
dunqan: thanks!

libs/platform/core/src/services/query/default-query.service.ts
dunqan: 👍

libs/platform/core/src/services/query/default-query.service.ts
dunqan: 👍

libs/platform/core/src/services/query/core/core-query.ts
dunqan: no, we want to sort parameters of qualifier before faltting values. Flatting it uset to make key string more compact.

libs/platform/core/src/services/query/core/core-query.ts
dunqan: Hmm... 
I like the first one for readibility, and hope that optimizer will optimize it. But not strong opinion :)

supproduction (approved):

PR #64, opened 2023-02-23
feat: product videos

tobi-or-not-tobi:
Render product video's when they're part of the product media (aka "image" set). 

We've also done some heavy refactor on this component:
- use asyncState, schema, naming convention
- leverage `oryx-layout` rather than custom layout inside the product images component


libs/domain/product/images/src/images.component.spec.ts
```diff
@@ -41,7 +41,9 @@ describe('ProductImagesComponent', () => {
         },
       ],
     });
-    element = await fixture(html`<product-images sku="1"></product-images>`);
```
Sinucid, line 44: I can't leave comment for unaffected parts, but replace this above:
`getOptions = ({ uid = '' }): Observable<any> => of({});`
by
`getOptions = vi.fn().mockReturnValue(of({}))`

libs/domain/product/images/src/images.component.ts
```diff
@@ -1,248 +1,186 @@
...
-      style += `--oryx-grid-items-per-column: ${options.gridItemsPerColumn}`;
+  protected override render(): TemplateResult | void {
+    if (this.product) {
+      const media = this.resolveImages();
+      const main = this.renderMainLayout(media);
+      const navigation = this.renderNavigationLayout(media);
+
+      return html`
+        <oryx-layout
+          navigation=${this.componentOptions?.navigationPosition ||
```
Sinucid, line 53: You have default options, why do you need fallback values then?

libs/domain/product/images/src/images.component.ts
```diff
@@ -1,248 +1,186 @@
...
-            mediaIndex: i,
-            containerSize: ProductMediaContainerSize.Detail,
-            alt: `preview${i + 1}`,
-            loading: active !== i ? 'lazy' : undefined,
-          }}
-          ?active=${active === i}
-        ></oryx-product-media>`
+    return html`<oryx-layout
+      class="main"
+      .layout=${layout || ProductImagesMainLayout.Carousel}
```
Sinucid, line 87: The same, you have default options

libs/domain/product/images/src/images.component.ts
```diff
@@ -1,248 +1,186 @@
...
+    if (media.length < 2 || display === ProductImagesNavigationDisplay.None) {
+      return;
     }
 
-    return html` <fieldset class="nav">
-      ${images.map(
-        (image: ProductMedia, i: number) => html`
+    return html`<oryx-layout
+      class="navigation"
+      .layout=${layout || NavigationLayout.Carousel}
```
Sinucid, line 125: ^^

libs/domain/product/images/src/images.component.ts
```diff
@@ -1,248 +1,186 @@
...
-  /**
-   * Resets the thumbnail navigation. This is useful when the image data
-   * is change (i.e. during page navigation) or when the options are changing.
-   */
-  protected reset<T>(source$: Observable<T>): Observable<T> {
-    return source$.pipe(
-      tap(() => this.setActive(0, ProductImagesScrollBehavior.Disable))
-    );
+  protected resolveImages(): ProductMedia[] {
+    const set = !this.componentOptions?.mediaSet
```
Sinucid:
```
return (!this.componentOptions?.mediaSet
      ? this.product?.mediaSet?.[0]
      : this.product?.mediaSet?.find((set) => set.name === this.componentOptions?.mediaSet))?.media ?? []
```

libs/domain/product/images/src/images.model.ts
```diff
@@ -95,16 +97,10 @@ export interface ProductImagesComponentOptions {
    *
    * @default copies the `thumbHeight`
```
Sinucid, line 98: What does it mean?

libs/domain/product/images/src/images.styles.ts
```diff
@@ -0,0 +1,100 @@
...
+export const productImageStyles = css`
+  :host {
+    --gap: 15px;
+
+    display: block;
+  }
+
+  :host > oryx-layout {
+    display: grid;
+    gap: var(--gap);
```
Sinucid, line 12:
```suggestion
    gap: var(--gap, 15px);
```
And drop css var definition above

libs/domain/product/media/src/media.component.ts
```diff
@@ -32,14 +32,34 @@ export class ProductMediaComponent extends ProductMixin(
       containerSize
     );
 
+    if (sources[0]?.url) {
```
Sinucid:
```
if (!sources[0]?.url)  {return html``};

if (this.isVideo(sources[0]?.url)) {
        return this.renderVideo(sources[0]?.url);
}

return this.renderImage(sources[0]?.url, this.getSrcSet(sources));
```
 
And another thing. I noticed that you call service inside render and try to get the source.
```
const sources = this.imageService.resolveSources(
      productMedia,
      containerSize
);
```

It will call only during re-render, and if for some reason service resolves the source after last render it doesn't affect on template, because this is not reactive

libs/platform/experience/src/services/experience/layout/default-layout.builder.ts
```diff
@@ -120,7 +120,7 @@ export class DefaultLayoutBuilder implements LayoutBuilder {
           if (!options?.omitUnit) {
             value = addUnit(value, options?.unit);
           } else if (value === 0 && !options?.addEmpty) {
-            value = '';
+            // value = '';
```
Sinucid: ?

libs/template/presets/src/features/b2c/app.ts
```diff
@@ -58,7 +57,7 @@ export const b2cFeatures: AppFeature[] = [
   {
     resources: b2cResources,
   },
-  StaticExperienceFeature,
+  // StaticExperienceFeature,
```
Sinucid: Need to uncomment I guess

Sinucid (changes requested):

libs/domain/product/images/src/images.component.ts
tobi-or-not-tobi: The problem is currently with EB. If you clear a value, it creates an empty string rather than an undefined, which is why we need to use `||` operator.

libs/domain/product/images/src/images.component.ts
tobi-or-not-tobi: see above

libs/domain/product/images/src/images.component.ts
tobi-or-not-tobi: yes i understand ;)

libs/domain/product/images/src/images.styles.ts
tobi-or-not-tobi: That was not working, but I've changed a small rule in the layout styles to allow for this.

libs/domain/product/media/src/media.component.ts
tobi-or-not-tobi: Good point, i'm changing to update lifecycle

libs/domain/product/images/src/images.component.ts
Sinucid: But in this case should it use the defaultOptions? I believe we will have at least something, EB options or default ones, but still

Sinucid (dismissed):

libs/domain/product/images/src/images.component.ts
tobi-or-not-tobi: debatable. I believe it's a EB bug, it should clear the option in case the user clears a select, so it's a bit of workaround.

PR #77, opened 2023-02-27
chore: heading small screen appearance

tobi-or-not-tobi:
We added appearance styles for small screens.

As part of this PR, we change the configuration of the screen sizes; screen sizes no longer will flow into larger screens, but are limited to the min AND max config of the breakpoint. As a result, we've been refactoring a few component styles which depends on the small screen size rules that was inherited by larger screen sizes originally.


Sinucid (dismissed):

libs/domain/search/box/src/styles/base.styles.ts
```diff
@@ -116,7 +116,7 @@ export const baseStyles = css`
 `;
 
 const mediumScreen = css`
-  :host([stretched]) {
+  /* :host([stretched]) {
```
Sinucid: medium screen styles still relevant?

Sinucid (approved):

PR #81, opened 2023-02-27
feat: add static data into eb

supproduction:
adds static data into eb

closes [HRZ-2310](https://spryker.atlassian.net/browse/HRZ-2310)

[HRZ-2310]: https://spryker.atlassian.net/browse/HRZ-2310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/experience/src/services/experience/data-client/data-client.model.ts
+++ b/libs/platform/experience/src/services/experience/data-client/data-client.model.ts
@@ -1,5 +1,6 @@
 import { ResourceGraphic } from '@spryker-oryx/core';
 import { ContentComponentSchema } from '../../../models';
+import { StaticComponent } from '../static-data';
 
 export const enum MessageType {
   Graphics = 'oryx.graphics',
@@ -7,6 +8,7 @@ export const enum MessageType {
   Products = 'oryx.products',
   Query = 'oryx.query',
   Model = 'oryx.component-model',
+  Static = 'oryx.static',
   ComponentType = 'oryx.component-type',
   Schemas = 'oryx.component-schemas',
 }
@@ -29,6 +31,8 @@ export type ExperienceMessageData<T> = {
     ? string
     : T extends MessageType.Schemas
     ? ContentComponentSchema[] | undefined
+    : T extends MessageType.Static
+    ? StaticComponent[]
     : never;
   [key: string]: any;
 };
--- a/libs/platform/experience/src/services/experience/static-data/default-experience-static.service.ts
+++ b/libs/platform/experience/src/services/experience/static-data/default-experience-static.service.ts
@@ -16,7 +16,7 @@ export class DefaultExperienceStaticService implements ExperienceStaticService {
 
   protected processData(): Component[] {
     return this.staticData.flat().map((component) => {
-      component.id = component.id ?? this.getAutoId();
+      component.id ??= this.getAutoId();
 
       this.processComponent(component as Component);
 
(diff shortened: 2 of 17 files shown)


libs/platform/experience/src/services/experience/static-data/default-experience-static.service.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.staticComponents = this.processData();
+  }
+
+  getData(): Component[] {
+    return this.staticComponents;
+  }
+
+  protected processData(): Component[] {
+    return this.staticData.flat().map((component) => {
+      component.id = component.id ?? this.getAutoId();
```
Sinucid:
```suggestion
      if (!component.id) {
        component.id = this.getAutoId();
      }
```

libs/platform/experience/src/services/experience/static-data/default-experience-static.service.ts
```diff
@@ -0,0 +1,40 @@
...
+      this.processComponent(component as Component);
+
+      return component as Component;
+    });
+  }
+
+  protected processComponent(component: Component): void {
+    const components = component?.components || [];
+
+    component.id = component.id ?? this.getAutoId();
```
Sinucid: Duplication?

libs/platform/experience/src/services/experience/static-data/default-experience-static.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    components.forEach((component: Component) => {
```
Sinucid:
```suggestion
    component?.components?.forEach((component: Component) => {
```

Sinucid (changes requested):

libs/template/application/root-app/src/root-app.component.ts
```diff
@@ -15,8 +15,16 @@ export class RootAppComponent extends LitElement {
   static styles = styles;
 
   override render(): TemplateResult {
-    return html`<experience-composition uid="header"></experience-composition>
-      ${this._router.outlet()}
-      <experience-composition uid="footer"></experience-composition>`;
+    return html`
+      ${asyncValue(
+        // Uses as a trigger for dropping hydration
```
dunqan: let's mark it as TODO, as we want to have better solution here...

libs/platform/experience/src/services/experience/static-data/default-experience-static.service.ts
```diff
@@ -0,0 +1,40 @@
+import { inject } from '@spryker-oryx/di';
+import { Component } from '../models';
+import { ExperienceStaticData, ExperienceStaticService } from './static-data';
+
+export class DefaultExperienceStaticService implements ExperienceStaticService {
```
dunqan, line 5: I'm not against this service, but I don't really like we have to iterate through all components recursively twice, because we moved autoId here 🤔 
(and without autoid functionality, overhead with separate service is probably not justified)

libs/platform/experience/src/services/experience/static-data/default-experience-static.service.ts
supproduction: another level

Sinucid (approved):

dunqan (approved):

libs/platform/experience/src/services/experience/data-client/data-client.model.ts
```diff
@@ -1,12 +1,14 @@
...
 import { ContentComponentSchema } from '../../../models';
+import { StaticComponent } from '../static-data';
 
 export const enum MessageType {
   Graphics = 'oryx.graphics',
   Options = 'oryx.options',
   Products = 'oryx.products',
   Query = 'oryx.query',
   Model = 'oryx.component-model',
+  Static = 'oryx.static',
```
tobi-or-not-tobi, line 11: not directly related, but maybe we can go with

```ts
ComponentType = 'oryx.component-type',
ComponentModel = 'oryx.component-model',
ComponentSchemas = 'oryx.component-schemas',

```

PR #84, opened 2023-02-28
fix(auth): refactor auth

gund:
- [x] Unify auth package to use Oauth
- [x] Add support for Authorization Code and Password Grant flows
- [x] Setup auth features for SAPI and BAPI in `application` package
- [x] Fix build/tests/ci

Labs PR: https://github.com/spryker/oryx-labs/pull/8


libs/platform/auth/package.json
```diff
@@ -1,33 +1,40 @@
 {
   "name": "@spryker-oryx/auth",
-  "version": "0.1.0-rc.0",
+  "version": "1.0.0-alpha.0",
```
tolerants: Can you please check? :)

libs/platform/auth/project.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "../../../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "libs/platform/router/src",
```
tolerants:
```suggestion
  "sourceRoot": "libs/platform/auth/src",
```

libs/platform/auth/project.json
```diff
@@ -0,0 +1,24 @@
...
+      "executor": "./tools/executors/components-library:build"
+    },
+    "paths": {
+      "executor": "./tools/executors/components-library:paths"
+    },
+    "lint": {
+      "executor": "@nrwl/linter:eslint",
+      "outputs": ["{options.outputFile}"],
+      "options": {
+        "lintFilePatterns": ["libs/platform/router/**/*.ts"]
```
tolerants:
```suggestion
        "lintFilePatterns": ["libs/platform/auth/**/*.ts"]
```

libs/platform/auth/src/components/login/login.component.ts
```diff
@@ -0,0 +1,317 @@
...
+  @property() heading?: DirectiveResult | string;
+  @property() isLoading?: boolean;
+  @property() hasError?: boolean;
+  @property() emailName = 'email';
+  @property() passwordName = 'password';
+  @property() rememberMeName = 'rememberme';
+
+  @state() protected isDisabled = false;
+
+  #internals =
```
tolerants: But is it working in Safari?

gund: yes
image

libs/platform/auth/package.json
gund: Fixed to 0.1.0 as in all other packages.

libs/platform/auth/project.json
gund: Fixed

libs/platform/auth/project.json
gund: Fixed

libs/platform/auth/src/components/login/login.component.ts
gund: Actually I had to drop it because storybook does not support this syntax in decorated classes 🤷

libs/platform/auth/package.json
```diff
@@ -2,32 +2,39 @@
   "name": "@spryker-oryx/auth",
   "version": "0.1.0",
   "type": "module",
-  "description": "Authentication features for Oryx applications",
-  "author": "Spryker Systems GmbH",
-  "license": "See license in LICENSE.md",
+  "description": "Authentication functionality for Oryx framework",
+  "author": "Spryker",
```
tolerants: Can you please revert this change?

libs/platform/auth/package.json
```diff
@@ -2,32 +2,39 @@
...
+  "license": "MIT",
```
tolerants: Can you please revert this change?

libs/platform/auth/package.json
gund: Fixed

libs/platform/auth/package.json
gund: Fixed

libs/platform/auth/src/components/logout/logout.def.ts
```diff
@@ -9,7 +9,7 @@ declare global {
 
 export const authLogoutComponent = componentDef({
   name: 'oryx-auth-logout',
-  impl: () => import('./logout.component').then((m) => m.AuthLogoutComponent),
+  impl: () => import('./logout.component').then((m) => m.default),
```
dunqan: Why `default`?

dunqan (dismissed):

tolerants (dismissed):

libs/platform/auth/src/components/logout/logout.def.ts
gund: Actually I would like to propose to use default for our dynamic imports to reduce amount of code needed to be written, and to add default extraction to the loader itself, so we can simply keep `import('./logout.component')` and no `.then()` chain is required.
So this was just a preparation for that stage =)

dunqan (dismissed):

libs/platform/auth/src/components/logout/logout.def.ts
dunqan: yeah, we should discuss it first then, instead of sneaking it into code base :P

tobi-or-not-tobi: Not a big deal to merge this PR now, but I agree we need to discuss this one. 

I wonder if there are any consequences. For example, in some cases we might have multiple exports in the component fil.

tobi-or-not-tobi: Anyway, let's rol it back as it's not related to auth topic

tobi-or-not-tobi: (i'm doing it so we can resolve this one and merge)

dunqan (approved):

PR #89, opened 2023-02-28
fix: render fallback image in case of unresolved source image

aleespryker:
Closes [HRZ-2331](https://spryker.atlassian.net/browse/HRZ-2331)

[HRZ-2331]: https://spryker.atlassian.net/browse/HRZ-2331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AndreyYevtukhov:
Screenshot 2023-02-28 at 19 47 05

@aleespryker the image is there, but it is not centred in the block -> can you also change it pls? We have done it before somewhere (if I'm not mistaken) - I'm wondering how it is possible that it appeared again..

Do we have 2 images that are the same or I'm just lost...)

aleespryker:
> Screenshot 2023-02-28 at 19 47 05
> 
> @aleespryker the image is there, but it is not centred in the block -> can you also change it pls? We have done it before somewhere (if I'm not mistaken) - I'm wondering how it is possible that it appeared again..
> 
> Do we have 2 images that are the same or I'm just lost...)

We replaced the image completely because it wasn't centered elsewhere and the source image had extra white space in the bottom corner, and didn't match the version on Figma to begin with. So I think whatever css we're using here we can just drop so that it's properly centered anyways.

aleespryker:
> Screenshot 2023-02-28 at 19 47 05
> 
> @aleespryker the image is there, but it is not centred in the block -> can you also change it pls? We have done it before somewhere (if I'm not mistaken) - I'm wondering how it is possible that it appeared again..
> 
> Do we have 2 images that are the same or I'm just lost...)

Okay, I found out what the issue was. There are actually two icons, one in the b2c/storefront icon set and the other is in backoffice. It seems we grabbed the backoffice version of this icon and used it not knowing about the extra spacing in it, and it was recently replaced with the version from Figma, which visually looks the same, but doesn't have the extra spacing at the bottom, so components using this version of the icon are properly centered without requiring extra CSS to do it. But if you look at storybook without using any query parameters, you will allways see the backoffice version of the icon. To see the new version of the icon, you have to use `&globals=oryx-thene:b2c`. This also applies to some components in the checkout or order library where we had to add new icons that didn't previously exist in backoffice. The icon will not show up until you add that query parameter in storybook. I did ask @tobi-or-not-tobi about this earlier, and he said it was alright if the icons didn't show up.

tobi-or-not-tobi (approved):

PR #90, opened 2023-03-01
fix: HRZ-2278 post release fixes

tolerants:
PR for release 0.1.0
Fixed import issues in search package.
Moved CatogoryX interface to the group-item-variants-generator

closes: https://spryker.atlassian.net/browse/HRZ-2278


dunqan (dismissed):

AndreyYevtukhov (dismissed):

libs/domain/search/src/mocks/src/suggestion/mock-suggestion.service.ts
```diff
@@ -1,9 +1,6 @@
-import {
-  Suggestion,
-  SuggestionQualifier,
-  SuggestionService,
-} from '@spryker-oryx/search';
 import { Observable, of } from 'rxjs';
+import { Suggestion, SuggestionQualifier } from '../../../models';
```
supproduction: It's ok for mocks use alias

supproduction (changes requested):

libs/domain/search/src/mocks/src/suggestion/mock-suggestion.service.ts
tolerants: Agree, but was trying if it can break. Gonna revert.

supproduction (approved):

PR #94, opened 2023-03-01
fix: product images vertical overflow

Sinucid:
closes to [HRZ-2337](https://spryker.atlassian.net/browse/HRZ-2337)

[HRZ-2337]: https://spryker.atlassian.net/browse/HRZ-2337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/product/images/src/images.styles.ts
+++ b/libs/domain/product/images/src/images.styles.ts
@@ -30,6 +30,8 @@ export const productImageStyles = css`
 
   .main {
     --cols: 1;
+
+    overflow-y: hidden;
   }
 
   .navigation {
--- a/libs/domain/product/images/src/images.styles.ts
+++ b/libs/domain/product/images/src/images.styles.ts
@@ -30,8 +30,6 @@ export const productImageStyles = css`
 
   .main {
     --cols: 1;
-
-    overflow-y: hidden;
   }
 
   .navigation {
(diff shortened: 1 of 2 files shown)


libs/domain/product/images/src/images.styles.ts
```diff
@@ -30,6 +30,8 @@ export const productImageStyles = css`
 
   .main {
     --cols: 1;
+
+    overflow-y: hidden;
```
supproduction: Is it break scrolling in case we have vertical navigation?

supproduction (changes requested):

libs/domain/product/images/src/images.styles.ts
Sinucid: This is for main layout only, not for navigation

supproduction (dismissed):

libs/domain/product/images/src/images.styles.ts
tobi-or-not-tobi: Can we fix this inside layout styles please? We should not have vertical scrolling on horizontal carousel and on vertical we should not be able scroll horizontal

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #96, opened 2023-03-01
feat(e2e): add sf e2e tests

AndreyYevtukhov:
Adds Cypress E2E tests for Storefront, closes https://spryker.atlassian.net/browse/HRZ-2358


apps/storefront-e2e/src/types/product.type.ts
```diff
@@ -0,0 +1,8 @@
+export type Product = {
```
supproduction: Why we can't use types from packages?

apps/storefront-e2e/project.json
```diff
@@ -0,0 +1,36 @@
...
+      "configurations": {
+        "headless": {
+          "watch": false
+        },
+        "headless-ci": {
+          "cypressConfig": "apps/storefront-e2e/cypress.ci.config.js",
+          "watch": false,
+          "record": true
+        }
+      }
```
supproduction, line 24: I would suggest use it as separate targets, cause we do like this everywhere

supproduction: And command shorter then

package.json
```diff
@@ -12,6 +12,10 @@
     "prepare": "husky install",
     "build-storybook": "nx run storybook:build",
     "build:executors": "tsc --project ./tools/tsconfig.tools.json",
+    "sf:e2e:open": "npx concurrently --kill-others --success first \"npx nx run storefront:serve-prod\" \"wait-on -i=2000 http://localhost:3001 && npx nx run storefront-e2e:e2e\"",
+    "sf:e2e:headless": "npx concurrently --kill-others --success first \"npx nx run storefront:serve-ssr\" \"wait-on -i=2000 http://localhost:3001 && npx nx run storefront-e2e:e2e:headless\"",
+    "sf:e2e:headless:ci": "npx concurrently --kill-others --success first \"npx nx run storefront:serve-prod\" \"wait-on -i=2000 http://localhost:3001 && npx nx run storefront-e2e:e2e:headless-ci\"",
+    "sf:e2e:headless:ci:affected": "npx nx print-affected --select=projects | grep -E 'storefront' && npm run sf:e2e:headless:ci",
```
supproduction, line 18: Why do we need `npx` here?

supproduction: you can use `run-p` instead of `concurrently` and delete this new dep

supproduction (changes requested):

apps/storefront-e2e/project.json
AndreyYevtukhov: but in the case of separate targets, we will have a lot of copy-pasted code (as we do now)
I had an issue with that 3 times so far, and decided to do it NX way (which actually should be used everywhere) :)

package.json
AndreyYevtukhov: > Why do we need npx here? 

I use it everywhere to run packages - why should we NOT use it?)

AndreyYevtukhov: What is `run-p`? (I see that it was last updated 7 years ago, so I'm not quite sure that it is usable - https://npmtrends.com/concurrently-vs-npm-run-all-vs-run-p-vs-run-parallel-vs-wsrun)

---

And why should we replace the `concurrently` package? 

It does everything we need here: 

- parallel tasks execution
- nice useful logs
- correct process killing and handling
- the ability to manage exit codes (which is used in CI) etc. 

It is actively supported and does not have security issues - so I can't understand what is the problem here and why it should be changed (and why we should spend hours again debugging and configuring this flow with a new tool like `run-p` or something if we already have a working solution) :)

apps/storefront-e2e/src/types/product.type.ts
AndreyYevtukhov: I'm not quite sure what you mean here(( 

These types are used only for **E2E test data** and are not related to frontend types or API responses

I can change the name to `ProductTestData` or something to make it more precise. Should I do it?

supproduction: Better change it

apps/storefront-e2e/project.json
supproduction: I am not against that to change it everywhere. I am against that this one will be not the same as others. If we have an issue we need ticket to fix it

package.json
supproduction: You need just script which will run processes in parallel. We already have this, why do we need new dep?

apps/storefront-e2e/project.json
AndreyYevtukhov: https://spryker.atlassian.net/browse/HRZ-2363 - refactoring ticket

package.json
AndreyYevtukhov: https://spryker.atlassian.net/browse/HRZ-2364 - refactoring ticket

apps/storefront-e2e/src/types/product.type.ts
AndreyYevtukhov: Done

supproduction (dismissed):

supproduction (approved):

PR #104, opened 2023-03-06
fix: improve collapsible component's keyboard navigation

Sinucid:
Unfortunately we can't overcome the accessibility issue with nested interactive elements. Interactive elements can't be ignored, and accessibility checker tools will always recognise nesting. But, despite of that issue has a 'serious' status it does not guarantee that screen readers will ignore that content. I've checked this in 2 screen readers plugins for FF and chrome, and they recognise nested interactive content. Also built in accessibility tools for FF and chrome are also getting all labels, even from nested interactives. In result, we have much more benefits from keeping the structures nested. This allows to build much more convenient user experience and keyboard navigation. I would suggest to make some exception regarding this accessibility issue

closes to [HRZ-2242](https://spryker.atlassian.net/browse/HRZ-2242)

[HRZ-2242]: https://spryker.atlassian.net/browse/HRZ-2242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/search/facet-value-navigation/src/facet-value-navigation.component.ts
```diff
@@ -45,7 +45,7 @@ export class SearchFacetValueNavigationComponent
   }
 
   protected override render(): TemplateResult {
-    return html` <oryx-collapsible ?open=${this.open}>
+    return html` <oryx-collapsible ?open=${this.open} nonTabbable>
```
tobi-or-not-tobi: I like this solution! What about `nonTabbable=${onlyWhenTheClearButtonIsAvailable}` ?

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #111, opened 2023-03-06
feat: migate picking-lists to picking domain

tarashaliuk:
closes: [HRZ-2377](https://spryker.atlassian.net/browse/HRZ-2377)

[HRZ-2377]: https://spryker.atlassian.net/browse/HRZ-2377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/components/customer-note-modal/customer-note-modal.styles.ts
```diff
@@ -0,0 +1,16 @@
...
+    --oryx-card-body-padding: 18px 30px;
+  }
+
+  [slot='heading'] {
+    font-size: 18px;
+  }
+
+  oryx-button {
+    width: 100%;
+  }
```
tobi-or-not-tobi: We should not have any PA specifics here. We're [asking for a solution at design](https://spryker.slack.com/archives/C03NTPT5ZTQ/p1678351425008989) if these can be become part of mobile variation of modal.

libs/domain/picking/src/components/customer-note-modal/customer-note-modal.styles.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const styles = css`
+  oryx-modal {
+    --oryx-card-header-padding: 18px 30px;
+    --oryx-card-body-padding: 18px 30px;
+  }
+
+  [slot='heading'] {
+    font-size: 18px;
+  }
```
tobi-or-not-tobi: We should not have PA specifics here. Can we leverage the heading component to accomplish the font size?

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
```diff
@@ -1,24 +1,38 @@
...
+      ${asyncValue(
+        this.pickingLists$,
+        (pickingLists) => this.renderPickingLists(pickingLists),
+        () => html`<p>Loading picking lists...</p>`
+      )}
+
+      <oryx-customer-note-modal
+        customerNote=${ifDefined(this.customerNote)}
+        @close=${this.closeCustomerNoteModal}
+      ></oryx-customer-note-modal>
```
tobi-or-not-tobi: @Sinucid can you check this one and see if we can drop the modal component all together? I like to use the service to create the modal only when needed, and only provide the content of the modal here.

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
```diff
@@ -1,24 +1,38 @@
...
 
-  protected override render(): unknown {
-    return html`${asyncValue(
-      this.pickingLists$,
-      (pickingLists) => this.renderPickingLists(pickingLists),
-      () => html`<p>Loading picking lists...</p>`
-    )}`;
+  protected override render(): TemplateResult {
+    return html`
+      ${asyncValue(
```
tobi-or-not-tobi: `asyncValue` is outdated, lets use `asyncState` (and stay tuned for signals which will come very soon)

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
```diff
@@ -32,12 +46,23 @@ export class PickingListsComponent extends LitElement {
...
-        () => html`<p>No picking lists found!</p>`
+        this.renderFallback
       )}
     `;
   }
-}
 
-export default PickingListsComponent;
+  protected renderFallback(): TemplateResult {
+    return html`<p>No picking lists found!</p>`;
```
tobi-or-not-tobi: `i18n` please

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
tobi-or-not-tobi: Let's consider a mixin for the domain. Speak to @Sinucid or @dunqan if you need help.

libs/domain/picking/src/components/picking-lists/component.ts
```diff
@@ -1,6 +1,7 @@
 import { componentDef } from '@spryker-oryx/core';
 
-export const pickingListListComponent = componentDef({
+export const pickingListsComponent = componentDef({
```
tobi-or-not-tobi: the def file name should be `picking-list.def.ts`

libs/domain/picking/src/components/picking-list-card/component.ts
```diff
@@ -2,5 +2,8 @@ import { componentDef } from '@spryker-oryx/core';
 
 export const pickingListCardComponent = componentDef({
   name: 'oryx-picking-list-card',
-  impl: () => import('./picking-list-card.component').then((m) => m.default),
+  impl: () =>
```
tobi-or-not-tobi: the filename should be `picking-list-item.def.ts`

I'd name the feature picking-list-item or picking-item, to not lock it to the UI flavour (card).

libs/base/ui/structure/card/src/card.component.ts
```diff
@@ -12,7 +12,7 @@ export class CardComponent extends LitElement {
 
   protected override render(): TemplateResult {
     return html`
-      <slot name="heading">
+      <slot name="heading" part="heading">
```
Sinucid: Do we really need to set this part attribute on the slot? Can we achieve the same by adding necessary styles to the slotted content?

libs/domain/picking/src/components/picking-list-card/picking-list-card.component.ts
```diff
@@ -1,16 +1,98 @@
...
+  static styles = styles;
 
-  protected override render(): unknown {
-    return html`<p>
-      PickingList ${this.pickingList?.id} in status ${this.pickingList?.status}
-      with ${this.pickingList?.items.length} item(s)
-    </p>`;
+  protected pickingListService = resolve(PickingListService);
+
+  @property({ type: Object }) pickingList?: PickingList;
```
tobi-or-not-tobi: We do not like to blob in the pricking list. I'd add a property for the picking list item ID (uid, code, ..., ?) and then use a service to resolve the item. Additionally we could use a mixin, controller and context, but we can introduce this later on.

libs/domain/picking/src/components/picking-list-card/picking-list-card.styles.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  oryx-icon[type='info'] {
+    color: var(--oryx-color-primary-300);
+  }
+
+  oryx-button {
+    margin-block: 0;
+    width: 100%;
+  }
```
tobi-or-not-tobi: we need to have very good arguments to introduce custom styles here. We like to leverage the design system as much as possible. If the designs diverge, we should first talk to design team to understand why. We've seen many cases where more alignment is needed.

libs/domain/picking/src/components/customer-note-modal/customer-note-modal.component.ts
```diff
@@ -0,0 +1,32 @@
...
+      @oryx.close=${this.closeCustomerNoteModal}
+    >
+      <span slot="heading">Customer note</span>
+
+      ${this.customerNote}
+
+      <oryx-button slot="footer" type="primary" size="small">
+        <button @click=${this.closeCustomerNoteModal}>
+          <oryx-icon type="checkMark"></oryx-icon>
+          Got it
```
Sinucid: use `i18n` please

libs/domain/picking/src/components/index.ts
```diff
@@ -1,2 +1,3 @@
+export * from './customer-note-modal/component';
```
tobi-or-not-tobi: component must not be exported to not break lazy loading

tobi-or-not-tobi (changes requested):

libs/domain/picking/src/components/picking-list-card/picking-list-card.component.ts
```diff
@@ -1,16 +1,98 @@
...
+    this.isDisabled = false;
   }
-}
 
-export default PickingListCardComponent;
+  protected showCustomerNote(): void {
+    this.dispatchEvent(
+      new CustomEvent('showCustomerNote', {
+        detail: {
+          customerNote: this.pickingList?.cartNote,
```
Sinucid: Do we need to show the note in case we don't have `this.pickingList?.cartNote`?

I believe need to dispatch the event in case if we have something to show

libs/domain/picking/src/components/picking-list-card/picking-list-card.component.ts
```diff
@@ -1,16 +1,98 @@
...
+        },
+      })
+    );
+  }
+
+  protected override render(): TemplateResult {
+    return html`
+      <oryx-card>
+        <div slot="heading" class="qqq">qqq</div>
```
Sinucid: Left over?

libs/domain/picking/src/components/picking-list-card/picking-list-card.component.ts
```diff
@@ -1,16 +1,98 @@
...
+        <div slot="heading" class="time">
+          ${this.formatTime(this.pickingList?.createdAt)}
+        </div>
+        <div slot="heading" class="identifier">${this.pickingList?.id}</div>
+
+        <div class="total">
+          <oryx-icon type="cart"></oryx-icon>
+          <span :data-e2e="order-total-label-${this.pickingList?.id}"
```
Sinucid: Do we need e2e identifier?

libs/domain/picking/src/components/picking-list-card/picking-list-card.component.ts
```diff
@@ -1,16 +1,98 @@
...
+          >
+            Start picking
+          </button>
+        </oryx-button>
+      </oryx-card>
+    `;
+  }
+
+  protected formatTime(time?: Date): string {
+    if (!time) {
```
Sinucid: We've already have an issue related to optional time: https://spryker.atlassian.net/browse/HRZ-2382

I believe here will be the same: render the formatted time or nothing. Don't forgot to update styles if need

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
```diff
@@ -1,24 +1,38 @@
...
-      this.pickingLists$,
-      (pickingLists) => this.renderPickingLists(pickingLists),
-      () => html`<p>Loading picking lists...</p>`
-    )}`;
+  protected override render(): TemplateResult {
+    return html`
+      ${asyncValue(
+        this.pickingLists$,
+        (pickingLists) => this.renderPickingLists(pickingLists),
+        () => html`<p>Loading picking lists...</p>`
```
Sinucid: i18n

libs/domain/picking/src/components/customer-note-modal/customer-note-modal.component.ts
```diff
@@ -0,0 +1,32 @@
...
+import { property } from 'lit/decorators.js';
+import { styles } from './customer-note-modal.styles';
+
+export class CustomerNoteModal extends LitElement {
+  static styles = styles;
+
+  @property() customerNote?: string;
+
+  protected closeCustomerNoteModal(): void {
+    this.dispatchEvent(new Event('close'));
```
Sinucid: Like a public event it should have name like: `oryx.[...]`.

And could you double check, please. `oryx.close` event that is emitted by oryx-modal should bubble up and you can listen to it outside, instead `close` event

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
Sinucid: @tobi-or-not-tobi Not sure. Here is also an event listener for close event. From developers perspective it is overhead. What benefits will it bring to create modal by service?

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
```diff
@@ -32,12 +46,23 @@ export class PickingListsComponent extends LitElement {
             (pl) =>
               html`<oryx-picking-list-card
                 .pickingList=${pl}
+                @showCustomerNote=${this.openCustomerNoteModal}
```
Sinucid: `@oryx.show-note`

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
```diff
@@ -1,24 +1,38 @@
...
-    )}`;
+  protected override render(): TemplateResult {
+    return html`
+      ${asyncValue(
+        this.pickingLists$,
+        (pickingLists) => this.renderPickingLists(pickingLists),
+        () => html`<p>Loading picking lists...</p>`
+      )}
+
+      <oryx-customer-note-modal
```
Sinucid: And I'm wondering, why do we need this modal as separate component? It is not so complex inside, and to decrease complexity of the list will be better just use oryx-modal here

Sinucid (changes requested):

libs/domain/picking/src/components/picking-list-card/picking-list-card.component.ts
tarashaliuk: just migrated it from PA

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
tarashaliuk: The customer note modal looks like a separate component and has different responsibilities than picking lists component. Those are reasons why it should be a separate component.

Sinucid: This should be true in case if:
1) Component will be reused inside same template or even in another components
2) To decrease complexity of parent component and make it much more readable

Seems to be both cases weren't matched. It won't be reused and complexity even grew.

In result customer note modal contains a few rows of template, useless event dispatcher and doesn't contain any other specific stuff. Even it's styles should be aligned on app level (excepts paddings of the card). I would strongly suggest you just move it's template part to the picking card directly


```
    return html`
      ${this.renderPickingLists()}

       <oryx-modal 
         ?open=${this.customerNote}
          enableFooter
          @oryx.close=${this.closeCustomerNoteModal}
       >
          <span slot="heading">${i18n('picking.customer-note.heading')}</span/>
          ${this.customerNote}
          <oryx-button slot="footer" type="primary" size="small">
            <button value="cancel">
              <oryx-icon type="checkMark"></oryx-icon>
              ${i18n('picking.customer-note.close')}
            </button>
          </oryx-button>
        </oryx-modal>
    `;
  }
```

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
Sinucid: I suggested dropping just the separate component for modal implementation, the story is complex enough, lets not increase its complexity now

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
Sinucid: We have a sync and decided not to create a mixin now, due to there is almost nothing to move to it and there are no more components that will reuse it yet

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  @observe()
+  protected pickingListId$ = new BehaviorSubject(this.pickingListId);
+
+  @state() isDisabled?: boolean;
+
+  protected pickingList$ = this.pickingListId$
+    .pipe(distinctUntilChanged())
+    .pipe(
+      filter((pickingListId) => !!pickingListId),
```
supproduction: So it means if id will be `0` we will not have emission?

libs/domain/picking/src/components/picking-list-item/picking-list-item.styles.ts
```diff
@@ -0,0 +1,64 @@
+import { css } from 'lit';
+
+export const styles = css`
+  oryx-card {
+    --oryx-card-header-padding: 16px 10px 12px;
+    --oryx-card-body-padding: 9px 10px 12px;
+    --oryx-card-footer-padding: 0 10px 12px;
```
supproduction: Assume should be fixed on design side

libs/domain/picking/src/components/picking-lists/picking-lists.component.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+          },
+        ],
+      });
+
+      element = await fixture(html`<oryx-picking-lists></oryx-picking-lists>`);
+    });
+
+    afterEach(() => {
+      vi.clearAllMocks();
+      destroyInjector();
```
supproduction: It's not necessary to duplicate `afterEach`

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+      element = await fixture(
+        html`<oryx-picking-list-item
+          .pickingList=${pickingListProp}
+        ></oryx-picking-list-item>`
+      );
+    });
+
+    afterEach(() => {
+      vi.clearAllMocks();
+      destroyInjector();
```
supproduction: Not needed duplication

libs/domain/picking/src/components/picking-lists/picking-lists.styles.ts
```diff
@@ -0,0 +1,24 @@
...
+  :host {
+    padding: 24px 16px 0;
+    width: 100%;
+    display: grid;
+    gap: 34px;
+  }
+
+  oryx-modal {
+    --oryx-card-header-padding: 18px 30px;
+    --oryx-card-body-padding: 18px 30px;
```
supproduction: Do we really need new paddings here?

libs/domain/picking/src/services/picking-list-default.service.ts
```diff
@@ -1,14 +1,59 @@
...
 
 export class PickingListDefaultService implements PickingListService {
+  protected pickingListsQualifier$ = new BehaviorSubject<PickingListQualifier>(
+    {}
+  );
+  protected pickingListsState$ = new BehaviorSubject<PickingList[]>([]);
+
+  protected pickingListsLoading$ = this.pickingListsQualifier$.pipe(
+    switchMap((qualifier) => {
+      console.log('qualifier', qualifier);
```
supproduction: Do we need this log?

supproduction (changes requested):

libs/domain/picking/src/components/picking-list-item/picking-list-item.styles.ts
tarashaliuk: Started a conversation with the design team here https://spryker.slack.com/archives/C03NTPT5ZTQ/p1678827556664729

libs/domain/picking/src/components/picking-list-card/picking-list-card.styles.ts
tarashaliuk: Started a conversation with the design team https://spryker.slack.com/archives/C03NTPT5ZTQ/p1678827556664729

supproduction (dismissed):

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
```diff
@@ -0,0 +1,129 @@
...
+              </button>
+            </oryx-icon-button>
+          `
+        )}
+
+        <oryx-button
+          slot="footer"
+          type=${ButtonType.Primary}
+          size=${Size.Lg}
+          @click=${this.startPicking}
```
Sinucid: click handling should be on the button element

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  protected pickingListService = resolve(PickingListService);
+
+  @property() pickingListId?: string;
+
+  @observe()
+  protected pickingListId$ = new BehaviorSubject(this.pickingListId);
+
+  @state()
+  protected isDisabled?: boolean;
```
Sinucid, line 36: Leave a comment please, that this should be refactored when implement start picking logic and loading state should be area of responsibility of the service.

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
```diff
@@ -0,0 +1,129 @@
...
+        >
+          <button :disabled=${this.isDisabled}>
+            ${i18n('picking.picking-list-item.start-picking')}
+          </button>
+        </oryx-button>
+      </oryx-card>
+    `;
+  }
+
+  protected formatTime(time: Date): string {
```
Sinucid: Looks like an utility

Sinucid (changes requested):

Sinucid (dismissed):

libs/domain/picking/src/components/picking-lists/picking-lists.styles.ts
```diff
@@ -0,0 +1,19 @@
+import { css } from 'lit';
+
+export const styles = css`
+  :host {
+    padding: 24px 16px 0;
+    width: 100%;
```
tobi-or-not-tobi: The layout seems off, probably we either drop 100% width or add the `box-sizing: border-box`

libs/domain/picking/src/components/picking-lists/picking-lists.styles.ts
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  oryx-modal [slot='heading'] {
+    font-size: 18px;
+    font-weight: 600;
+  }
+
+  oryx-modal oryx-button {
+    width: 100%;
+  }
```
tobi-or-not-tobi, line 12: we seem to have 2 options for rendering buttons inside modal footers. either 100% or auto. I've tried to find the pattern with UX team, but have not found it yet. Can we introduce a modal property that dictates to render footer buttons 100% or not? rather than adding the styles here?

libs/domain/picking/src/components/picking-lists/picking-lists.styles.ts
```diff
@@ -0,0 +1,19 @@
...
+  :host {
+    padding: 24px 16px 0;
+    width: 100%;
+    display: grid;
+    gap: 34px;
+  }
+
+  oryx-modal [slot='heading'] {
+    font-size: 18px;
+    font-weight: 600;
```
tobi-or-not-tobi: it should be a certain heading (h3, h4, etc) rather than using custom font sizing here.

libs/domain/picking/src/components/picking-lists/picking-lists.def.ts
```diff
@@ -0,0 +1,7 @@
+import { componentDef } from '@spryker-oryx/core';
+
+export const pickingListsComponent = componentDef({
+  name: 'oryx-picking-lists',
+  impl: () =>
+    import('./picking-lists.component').then((m) => m.PickingListsComponent),
```
tobi-or-not-tobi, line 6: can we add a schema so we can find it in EB? The options can be empty if there are none.

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
```diff
@@ -1,43 +1,90 @@
...
-              html`<oryx-picking-list-card
-                .pickingList=${pl}
-              ></oryx-picking-list-card>`
+              html`<oryx-picking-list-item
+                .pickingListId=${pl.id}
+                @oryx.show-note=${this.openCustomerNoteModal}
+              ></oryx-picking-list-item>`
           )}`,
-        () => html`<p>No picking lists found!</p>`
+        this.renderFallback
```
tobi-or-not-tobi:
```ts
if (this.pickingLists?.length) {
    return html`${this.pickingLists.map(
      (list) =>
        html`<oryx-picking-list-item
          .pickingListId=${list.id}
          @oryx.show-note=${this.openCustomerNoteModal}
        ></oryx-picking-list-item>`
    )}`;
  } else {
    return this.renderFallback();
  }
```

Now that we no longer use `asyncValue` we can use standard ts to conditionally render. 
Not sure if repeat would give us anything here, so feel free to change.

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
```diff
@@ -1,43 +1,90 @@
...
+          <button @click=${this.closeCustomerNoteModal}>
+            <oryx-icon type=${IconTypes.CheckMark}></oryx-icon>
+            ${i18n('picking.customer-note.close')}
+          </button>
+        </oryx-button>
+      </oryx-modal>
+    `;
+  }
+
+  protected renderFallback(): TemplateResult {
```
tobi-or-not-tobi:
```suggestion
  protected renderEmptyLists(): TemplateResult {
```

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
tobi-or-not-tobi, line 1: We're missing a barrel file to export the component and styles

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
tobi-or-not-tobi, line 1: We're missing a barrel file to export the component and styles

libs/domain/picking/src/utilities.ts
```diff
@@ -0,0 +1,9 @@
+export const formatTime = (time: Date): string => {
+  return time
+    .toLocaleString('en-US', {
```
tobi-or-not-tobi, line 3: It is important to not hardcode the locale. We have a local service that can be used. Have a look a the `DefaultPricingService`. Also, the [Intl api](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) might be useful to look at.

tobi-or-not-tobi: The idea of using the current locale would be that we can use this app cross locales without coding changes

libs/domain/picking/src/components/picking-list-card/picking-list-card.styles.ts
tobi-or-not-tobi: I've not seen the changes in the css that I'd expect. we still have css variables for the padding.

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
```diff
@@ -0,0 +1,113 @@
...
+                @click=${this.showCustomerNote}
+              >
+                <oryx-icon type=${IconTypes.Info}></oryx-icon>
+              </button>
+            </oryx-icon-button>
+          `
+        )}
+
+        <oryx-button slot="footer" type=${ButtonType.Primary} size=${Size.Lg}>
+          <button :disabled=${this.isDisabled} @click=${this.startPicking}>
```
tobi-or-not-tobi:
```suggestion
          <button ?disabled=${this.isDisabled} @click=${this.startPicking}>
```

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
```diff
@@ -0,0 +1,113 @@
...
+          note: this.pickingList?.cartNote,
+        },
+      })
+    );
+  }
+
+  protected override render(): TemplateResult {
+    return html`
+      <oryx-card>
+        <div slot="heading">
```
tobi-or-not-tobi: can we use `oryx-heading` here?

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
```diff
@@ -0,0 +1,113 @@
...
+    this.dispatchEvent(
+      new CustomEvent('oryx.show-note', {
+        detail: {
+          note: this.pickingList?.cartNote,
+        },
+      })
+    );
+  }
+
+  protected override render(): TemplateResult {
```
tobi-or-not-tobi:
```suggestion
  protected override render(): TemplateResult | void {
    if (!this.pickingList) return;
```
We could simplify some logic by adding this. no need to optional chaining and we can just use `<div class="time">${formatTime(this.pickingList.createdAt)}</div>`

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
```diff
@@ -0,0 +1,113 @@
...
+import { BehaviorSubject, distinctUntilChanged, filter, switchMap } from 'rxjs';
+import { formatTime } from '../../utilities';
+import { styles } from './picking-list-item.styles';
+
+export class PickingListItemComponent extends LitElement {
+  static styles = styles;
+
+  protected pickingListService = resolve(PickingListService);
+
+  @property() pickingListId?: string;
```
tobi-or-not-tobi, line 29:
```suggestion
  @property() id?: string;
```

not need to repeat as it's clear from the context

tobi-or-not-tobi (changes requested):

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
Sinucid: `id` is a standard attribute. Not so clear =)

tarashaliuk: Agree with @Sinucid. That’s why I used `pickingListId` instead of `id` to avoid attributes conflict

libs/domain/picking/src/components/picking-lists/picking-lists.styles.ts
tarashaliuk: It looks like a separate task that is not related to migrating picking lists to picking domain. I guess we can create a new subtask in the current story to avoid blocking Alvina and Pavlo PRs with this PR

libs/domain/picking/src/utilities.ts
tarashaliuk: Good idea! I think creating `DateTimeService` can also be reused in EB. It looks like a new feature and I think it can be separated into new subtask to avoid blocking Alvina and Pavel with the current PR

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
tobi-or-not-tobi: good point

tobi-or-not-tobi (dismissed):

Sinucid (dismissed):

Sinucid (approved):

PR #113, opened 2023-03-06
feat: add toggle color mode component

supproduction:
adds toggle color mode component

closes [HRZ-2244](https://spryker.atlassian.net/browse/HRZ-2244)

[HRZ-2244]: https://spryker.atlassian.net/browse/HRZ-2244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/site/mode-selector/src/mode-selector.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  static styles = styles;
+
+  protected darkMode = 'mode-dark';
+  protected lightMode = 'mode-light';
+  protected root =
+    resolve(AppRef).findPlugin(ComponentsPlugin)?.getRoot() ?? 'body';
+  protected storage = resolve(StorageService);
+
+  @subscribe()
+  protected mode$ = this.storage.get<string>(modeStorageKey).pipe(
```
dunqan: just to be sure, this one currently will emit only once, it's not observing changes to the storage. We may improve it in the future, but now this subscription is technically dead, if first emission is undefined.

libs/domain/site/mode-selector/src/mode-selector.def.ts
```diff
@@ -0,0 +1,11 @@
+import { componentDef } from '@spryker-oryx/core';
+
+export const siteModeSelectorComponent = componentDef({
+  name: 'oryx-site-mode-selector',
```
tobi-or-not-tobi:
```suggestion
  name: 'oryx-site-color-mode-selector',
```

Think we should rename the feature and all files, names, etc. to include color.

libs/domain/site/mode-selector/src/mode-selector.styles.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  button:hover {
+    background-color: transparent;
+    box-shadow: none;
+    border-color: var(--oryx-color-primary-300);
+  }
+
+  button:focus-visible {
+    outline: solid 1px blue;
+  }
```
tobi-or-not-tobi: I'd drop all styles except for maybe the color. It's not great, but it clearly shows that there's work left on the design system to make it look good.

libs/domain/site/mode-selector/src/mode-selector.component.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  protected override render(): TemplateResult {
+    return html`
+      <oryx-icon-button>
+        <button
+          type="button"
+          aria-label="${i18n('site.change-color-mode')}"
+          @click=${this.triggerEvent}
+        >
+          <oryx-icon type="${this.mode}"></oryx-icon>
```
tobi-or-not-tobi: when the mode is light, it should show the moon and visa versa, as it's an action button. If it would be just showing the state, then it would have been fine.

tobi-or-not-tobi (changes requested):
- When there's no mode selector in the storefront, the feature doesn't bubble down from EB; that's what we'd like to have

libs/domain/site/mode-selector/src/mode-selector.component.ts
supproduction: basically it's just showing icon state

libs/base/ui/action/color-mode-selector/src/color-mode-selector.component.ts
```diff
@@ -0,0 +1,87 @@
...
+import { styles } from './color-mode-selector.styles';
+import { ModeEvent, toggleMode } from './utilities';
+
+export const EVENT_TOGGLE_COLOR = 'oryx.toggle-mode';
+
+export class ColorModeSelectorComponent extends LitElement {
+  static styles = [styles];
+
+  protected darkMode = 'mode-dark';
+  protected lightMode = 'mode-light';
```
tobi-or-not-tobi: we have a few other locations where we need to distinquish dark vs light mode. can we centralise this model inside ui or experience package? We have some UI components that could benefit from shared model, as well as the color generation stuff that would. Might be something for a follow up when we move theme/color out of core.

tobi-or-not-tobi (changes requested):

libs/template/presets/src/features/b2c/experience/header.ts
```diff
@@ -19,19 +19,31 @@ export const HeaderTemplate: StaticComponent = {
...
+          options: {
+            data: {
+              rules: [
+                {
+                  margin: '0 0 0 auto',
+                },
+              ],
+            },
+          },
+        },
```
tobi-or-not-tobi: I'd drop it from storefront header actually, as it's not common to have.

tobi-or-not-tobi (changes requested):

libs/base/ui/action/color-mode-selector/src/color-mode-selector.component.ts
supproduction: If we have in plans move out `theme` from core I'll move them into it

tobi-or-not-tobi (dismissed):

tobi-or-not-tobi (approved):

PR #122, opened 2023-03-07
fix: hydration support

supproduction:
enables hydration
adds force hydration
adds hydration initializers
adds default hydration service

closes [HRZ-2351](https://spryker.atlassian.net/browse/HRZ-2351)

[HRZ-2351]: https://spryker.atlassian.net/browse/HRZ-2351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/structure/text/src/text.component.spec.ts
```diff
@@ -30,6 +25,17 @@ const text = `Lorem Ipsum is simply dummy text of the printing and typesetting
             with desktop publishing software like Aldus PageMaker including
             versions of Lorem Ipsum.`;
 
+vi.mock('@spryker-oryx/utilities', async () => {
```
dunqan, line 28: I guess we need to do it because isServer is true in unit tests?

libs/base/ui/structure/text/src/text.component.ts
```diff
@@ -57,7 +58,7 @@ export class TextComponent extends LitElement implements TextProperties {
           </oryx-icon-button>
         </slot>`
       )}
-      ${prehydrate(truncateFix, 'oryx-text')}
+      ${prehydrate(truncateFix, this.tagName.toLowerCase())}
```
dunqan, line 61: did you check if we have it on SSR?

libs/domain/site/locale-selector/src/locale-selector.component.ts
```diff
@@ -54,6 +55,12 @@ export class SiteLocaleSelectorComponent extends ContentMixin<SiteLocaleSelector
   }
 
   protected onClick(locale: string): void {
+    document.dispatchEvent(
```
dunqan: it should be side effect of language change, not click of this selector

libs/platform/core/src/directives/prehydrate.ts
```diff
@@ -24,7 +24,15 @@ class PrehydrateDirective extends Directive {
       return html``;
     }
     if (components[tag]) {
-      return html`${unsafeHTML(`<script>__dryFn('${tag}');</script>`)}`;
+      return html`${unsafeHTML(
```
dunqan, line 27: can you "minimize" it, as it can be replaced in the ouput html multiple times?

libs/platform/core/src/services/hydrate/default-hydrate.service.ts
```diff
@@ -0,0 +1,117 @@
...
+  rootInjectable,
+} from '@spryker-oryx/utilities';
+import { isServer, LitElement } from 'lit';
+import { AppRef, ComponentsPlugin } from '../../orchestration';
+import { HydrateService } from './hydrate.service';
+
+export const locationHydrationEvent = 'oryx.location';
+export const hydrateAllEvent = 'oryx.force-all';
+
+export class DefaultHydrateService implements HydrateService {
```
dunqan: I think core hydration logic, like initHydrateHooks, should be implemented and should work independently from the injector and provided services.

Also, if we are using service for initiating full rehydrate, why can't we just inject service an call some method?

libs/platform/experience/src/plugins/preview-plugin.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  beforeApply(): void {
+    const root = rootInjectable.get();
+
+    if (!isServer && root === 'body') {
+      return;
+    }
+
+    if (!isServer) {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
dunqan: shouldn't we check, if there is some pre-rendered content before rewritting element?

libs/base/ui/structure/text/src/text.component.spec.ts
supproduction: Yes

libs/base/ui/structure/text/src/text.component.ts
supproduction: Yes, we wrap default components with our additional logic on orchestration side 
image

libs/platform/core/src/services/hydrate/default-hydrate.service.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  protected initialize(): void {
+    const initializers = this.injector.inject(HydrateInitializer, null);
+
+    if (!initializers) {
+      return;
+    }
+
+    for (const initializer of initializers) {
+      this.subscription.add(initializer(this, this.injector).subscribe());
```
dunqan: I'd improve this one a bit. We should consider either:
- creating generic app initializers (then logic would stay the same)
- or create specific hydration initializers, we can react to here (so they won't have to do it in a initializer)

Also we sohuld not have to pass injector, as using factory and `inject` is more proper way to inject stuff.

libs/platform/experience/src/plugins/preview-plugin.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    if (!isServer && root === 'body') {
+      return;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const rootElement = document.querySelector(root)!;
+
+    if (!isServer && rootElement.shadowRoot) {
+      rootElement.outerHTML = `<${root}></${root}>`;
```
dunqan, line 27: Wondering if this one couldn't be solved by our hydration initializers ;)

libs/base/utilities/src/decorators/hydratable/hydratable.ts
```diff
@@ -11,62 +11,71 @@ const DEFER_HYDRATION = Symbol('deferHydration');
 const HYDRATION_CALLS = Symbol('hydrationCalls');
 export const HYDRATE_ON_DEMAND = '$__HYDRATE_ON_DEMAND';
 export const HYDRATING = '$__HYDRATING';
+export const hydratableAttribute = 'hydratable';
 
 export interface PatchableLitElement extends LitElement {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-misused-new
   new (...args: any[]): PatchableLitElement;
   _$needsHydration?: boolean;
+  hasSsr: boolean;
```
dunqan: I'd not pollute componet class with this arbitrary variable. We could use symbol, but actually I'm wondering if we really need it, and can get away withuot having it on the class?

libs/base/utilities/src/decorators/hydratable/hydratable.ts
```diff
@@ -111,14 +124,19 @@ function hydratableClass<T extends Type<HTMLElement>>(
     render(): TemplateResult {
       const states = this[asyncStates];
 
+      setTimeout(() => {
+        if (!this.useRealRender) this.useRealRender = true;
```
dunqan, line 130: I'm still wondering, if this one is fully bullet proof, as we are not triggering next render explicitely?

libs/platform/core/src/services/hydrate/hydrate.service.ts
```diff
@@ -0,0 +1,18 @@
+import { Observable } from 'rxjs';
+
+export const HydrateService = 'oryx.HydrateService';
+export const HydrateInitializer = 'oryx.HydrateInitializer*';
+
+export interface HydrateService {
```
dunqan: Can we rename to `HydrationService`?

libs/platform/core/src/services/hydrate/hydrate.service.ts
```diff
@@ -0,0 +1,18 @@
...
+  hydrateOnDemand(element: HTMLElement, skipMissMatch?: boolean): Promise<void>;
+  initHydrateHooks(immediate?: boolean): void;
+}
+
+export type HydrateInitializer = Observable<HTMLElement | unknown>;
```
dunqan: What about `HydrationTrigger`?

dunqan (approved):

PR #125, opened 2023-03-08
feat: fulfillment presets

Sinucid:
closes to [HRZ-2369](https://spryker.atlassian.net/browse/HRZ-2369)

[HRZ-2393]: https://spryker.atlassian.net/browse/HRZ-2393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HRZ-2369]: https://spryker.atlassian.net/browse/HRZ-2369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/themes/src/fulfillment.theme.ts
```diff
@@ -0,0 +1,12 @@
+import { Theme } from '@spryker-oryx/core';
+import { defaultBreakpoints } from '@spryker-oryx/themes/breakpoints';
+import { fulfillmentIcons } from '@spryker-oryx/themes/icons';
+
+// TODO: This should be dropped after HRZ-2239
```
supproduction, line 5: Do we need this comment?

supproduction (approved):

libs/template/themes/src/fulfillment.theme.ts
Sinucid: Nope, I guess, I've just restored the removed theme, but I believe after cleaning of the mess with icon for diff themes we will have only 2 of them, and this one will be removed

PR #126, opened 2023-03-08
feat: migrate push-notification package

aleespryker:
Closes [HRZ-2389](https://spryker.atlassian.net/browse/HRZ-2389), [HRZ-2391](https://spryker.atlassian.net/browse/HRZ-2391)

[HRZ-2389]: https://spryker.atlassian.net/browse/HRZ-2389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-2391]: https://spryker.atlassian.net/browse/HRZ-2391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/push-notification/src/services/default-push.service.ts
```diff
@@ -0,0 +1,34 @@
...
+    return this.provider
+      .getSubscription()
+      .pipe(switchMap((token) => this.sendToken(token)));
+  }
+
+  unsubscribe(): Observable<boolean> {
+    return this.provider.deleteSubscription();
+  }
+
+  protected sendToken(token: any): Observable<any> {
```
Sinucid: Can we specify a type here?

libs/domain/push-notification/src/services/providers/web-push/web-push.ts
```diff
@@ -0,0 +1,80 @@
...
+      ),
+      map((subscription) => subscription.toJSON())
+    );
+  }
+
+  deleteSubscription(): Observable<boolean> {
+    return this.getExistingSubscription().pipe(
+      switchMap((subscription) => {
+        if (subscription) {
+          return subscription.unsubscribe();
```
Sinucid: `subscription ? subscription.unsubscribe() : of(true)`

And what is the purpose of `of(true)`?

Sinucid (changes requested):

libs/domain/push-notification/src/services/default-push.service.ts
aleespryker: It depends on the provider. Firebase returns a different type than web-push. And it only seems possible to define the provider type (`TSubscription`) when calling the constructor unless there's a nicer way to do it.

And this service should probably be agnostic to the provider type anyways.

libs/domain/push-notification/src/services/providers/web-push/web-push.ts
aleespryker: The function should return a boolean. In the original, it also returns true if there is no subscription.

Sinucid (dismissed):

apps/storefront/dependencies.sh
```diff
@@ -1 +1 @@
-DEPENDENCIES="apps/storefront/* libs/base/di/* libs/base/form/* libs/base/ui/* libs/base/utilities/* libs/domain/cart/* libs/domain/checkout/* libs/domain/content/* libs/domain/order/* libs/domain/picking/* libs/domain/product/* libs/domain/search/* libs/domain/site/* libs/domain/user/* libs/platform/auth/* libs/platform/core/* libs/platform/experience/* libs/platform/i18n/* libs/platform/indexed-db/* libs/platform/offline/* libs/platform/router/* libs/template/application/* libs/template/labs/* libs/template/presets/* libs/template/themes/*"
\ No newline at end of file
+DEPENDENCIES="apps/storefront/* libs/base/di/* libs/base/form/* libs/base/ui/* libs/base/utilities/* libs/domain/cart/* libs/domain/checkout/* libs/domain/content/* libs/domain/order/* libs/domain/picking/* libs/domain/product/* libs/domain/push-notification/* libs/domain/search/* libs/domain/site/* libs/domain/user/* libs/platform/auth/* libs/platform/core/* libs/platform/experience/* libs/platform/i18n/* libs/platform/indexed-db/* libs/platform/offline/* libs/platform/router/* libs/template/application/* libs/template/labs/* libs/template/presets/* libs/template/themes/*"
```
gund: I don't think storefront is depending on the push-notifications package so this should probably be reverted.

libs/domain/push-notification/package.json
```diff
@@ -0,0 +1,28 @@
+{
```
gund: `push-notifications` package should be on a platform layer which is a lower level then the domain layer.

libs/domain/push-notification/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "name": "@spryker-oryx/push-notification",
+  "version": "0.0.1",
```
gund: Missing description, could be something like: "Provides low level capabilities to manage push subscriptions for Oryx".

libs/domain/push-notification/project.json
```diff
@@ -0,0 +1,36 @@
...
+      }
+    },
+    "lint": {
+      "executor": "@nrwl/linter:eslint",
+      "options": {
+        "lintFilePatterns": ["libs/domain/push-notification/**/*.ts"]
+      }
+    }
+  },
+  "tags": ["layer:domain", "root"]
```
gund: Layer must be "platform":
```suggestion
  "tags": ["layer:platform", "root"]
```

Also I'm not sure what is "root" tag actually for?

libs/domain/push-notification/src/services/providers/web-push/web-push.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+const mockServiceWorker = {
+  ready: new Promise<any>((resolve) =>
+    resolve({ pushManager: new PushManagerMock() })
+  ),
+};
+
+Object.defineProperty(navigator, 'serviceWorker', {
+  value: mockServiceWorker,
+  configurable: true,
+});
```
gund: Please use `spyOn()` in beforeEath block instead of direct property override to add mocks to global stuff only for this test suite.
Otherwise code in other places may be broken because of this.

libs/domain/push-notification/src/services/providers/web-push/web-push.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+      pushManager.getSubscription.mockReturnValue(of(undefined));
+
+      provider.getSubscription().subscribe((subscription) => {
+        expect(pushManager.subscribe).toHaveBeenCalledWith({
+          userVisibleOnly: true,
+        });
+        expect(subscription).toEqual(mockSubscription);
+        // Should be a different object then original subscription
+        expect(subscription).not.toEqual(pushManager.subscription);
+      });
```
gund: Please do not use `expect` inside of the callbacks like `.subscribe()` because if the callback is not invoked none of the expectations will be ever triggered.
You can use a spy and then check that it was called with specific arguments using `.toHabeenCalledWith()` API.

libs/domain/push-notification/src/services/providers/web-push/web-push.ts
```diff
@@ -0,0 +1,77 @@
...
+        subscription ? of(subscription) : this.createSubscription()
+      ),
+      map((subscription) => subscription.toJSON())
+    );
+  }
+
+  deleteSubscription(): Observable<boolean> {
+    return this.getExistingSubscription().pipe(
+      switchMap((subscription) =>
+        subscription ? subscription.unsubscribe() : of(true)
```
gund:
```suggestion
        subscription?.unsubscribe() ?? of(true)
```

libs/domain/push-notification/src/services/providers/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './push-provider.service';
+export * from './web-push';
```
gund: Push providers should be separated in their own secondary entry points, so they are not part of the main package API but should be imported explicitly, for example `@spryker-oryx/push-notifications/web` for web-push provider.

libs/domain/push-notification/src/services/default-push.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  ) {}
+
+  init(): Observable<void> {
+    return this.provider.init();
+  }
+
+  subscribe(): Observable<any> {
+    return this.provider
+      .getSubscription()
+      .pipe(switchMap((token) => this.sendToken(token)));
```
gund: This service should not make any http requests, we will implementing that in another story and another package.

libs/domain/push-notification/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "main": "./src/index.js",
+  "typings": "./src/index.d.ts",
+  "author": "Spryker Sytems GmbH",
+  "license": "See license in LICENSE.md",
+  "peerDependencies": {
+    "@spryker-oryx/core": "*",
+    "@spryker-oryx/di": "*",
+    "@spryker-oryx/router": "*",
+    "@spryker-oryx/ui": "*",
```
gund: I don't see dependencies to `router` and `ui` in this package so we should remove those.

libs/template/application/src/push-notification/bapi/feature.ts
```diff
@@ -0,0 +1,24 @@
+import { AppFeature, injectEnv } from '@spryker-oryx/core';
+import {
+  PushNotificationFeature,
+  PushServiceConfig,
+} from '@spryker-oryx/push-notification';
+
+export class BapiPushNotificationFeature
```
gund: We do not need this feature at all.
Simply integrate PushNotificationsFeature in the FulfillmentFeature.

libs/domain/push-notification/src/services/push-notification.providers.ts
```diff
@@ -0,0 +1,12 @@
...
+import { DefaultPushService } from './default-push.service';
+import { webPushProviders } from './providers';
+import { PushService } from './push.service';
+
+export const pushNotificationProviders: Provider[] = [
+  {
+    provide: PushService,
+    useClass: DefaultPushService,
+  },
+  ...webPushProviders,
```
gund: PushNotifications feature should only provide PushService and no other push providers.
And another more specific features should extend PushNotificationFeature and provide their specific push providers.

libs/domain/push-notification/src/services/push.service.ts
```diff
@@ -0,0 +1,23 @@
+import { Observable } from 'rxjs';
+
+export interface PushService {
+  init(): Observable<void>;
+  subscribe(): Observable<any>;
```
gund: Please use `unknown` instead of `any`.

gund (changes requested):

apps/storefront/dependencies.sh
aleespryker: `nx storefront:deps` complains about it if it's not there.

libs/domain/push-notification/src/services/providers/web-push/web-push.spec.ts
aleespryker: `spyOn` requires a function, so I ended up using `stubGlobal` instead. You can unstub it in `afterEach` as well.

libs/domain/push-notification/src/services/providers/web-push/web-push.spec.ts
aleespryker: Well I originally wanted to do it like this:

```
const callback = vi.rn();

provider.getSubscription().subscribe(callback);

expect(callback).toHaveBeenCalledWith(...)
```

But that didn't work because there's some promises in there and the callback doesn't get called at the time of the `expect`. So then we'd either have to also use something like `nextTick` in the original tests or use rxjs `lastValueFrom`. Or any other suggestions that work.

libs/domain/push-notification/src/services/default-push.service.ts
aleespryker: So what do we return for now, just the token? An adapter?

gund: Just the subscription.

libs/domain/push-notification/src/services/providers/web-push/web-push.spec.ts
gund: I think the simplest solution is to wait for next tick in the test or if it's promises only then `await Promise.resolve()` should be even better and slightly faster as you are going to execute the test in the single tick still.

aleespryker: Nevermind, got it to work with just observables. Mock `of` instead of using a real Promise.

libs/platform/push-notification/src/services/default-push.service.spec.ts
```diff
@@ -61,18 +47,16 @@ describe('PushService', () => {
 
       service.subscribe().subscribe();
       expect(provider.getSubscription).toHaveBeenCalled();
-      expect(http.url).toBe(`${apiUrl}/push-token`);
     });
 
-    it('should send subscription with merchant ref to api client', async () => {
-      const token = '1234';
+    it('should send subscription', async () => {
```
gund:
```suggestion
    it('should return subscription', async () => {
```

libs/platform/push-notification/web/feature.ts
```diff
@@ -0,0 +1,9 @@
+import { PushNotificationFeature } from '@spryker-oryx/push-notification';
+import { webPushProviders } from './web-push.providers';
+
+export class WebPushFeature extends PushNotificationFeature {
+  constructor() {
+    super();
+    this.providers = [...this.providers, ...webPushProviders];
+  }
```
gund:
```suggestion
  providers = [...super.providers, ...webPushProviders];
```

libs/platform/push-notification/web/feature.ts
```diff
@@ -0,0 +1,9 @@
+import { PushNotificationFeature } from '@spryker-oryx/push-notification';
+import { webPushProviders } from './web-push.providers';
+
+export class WebPushFeature extends PushNotificationFeature {
```
gund:
```suggestion
export class WebPushNotificationFeature extends PushNotificationFeature {
```

libs/template/presets/src/features/fulfillment/app.ts
```diff
@@ -31,6 +32,7 @@ export function fulfillmentFeatures(
     {
       resources: fulfillmentResources,
     },
+    new PushNotificationFeature(),
```
gund: We should use WebPushNotification feature in the fulfillment app
 ```suggestion
    new WebPushNotificationFeature(),
```

gund (changes requested):

gund (approved):

PR #133, opened 2023-03-10
feat: optional STORE

tobi-or-not-tobi:
We make the `STORE` environment variable optional. 

When there’s no var available, we’ll default to the first available store from the store response. 

closes [HRZ-2404](https://spryker.atlassian.net/browse/HRZ-2404)

[HRZ-2404]: https://spryker.atlassian.net/browse/HRZ-2404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/site/src/services/store/default-store.service.ts
+++ b/libs/domain/site/src/services/store/default-store.service.ts
@@ -31,10 +31,14 @@ export class DefaultStoreService implements StoreService {
       switchMap((store) =>
         !store
           ? this.getAll().pipe(
-              map(
-                (stores) =>
-                  stores.find((store) => store.id === this.store) ?? null
-              )
+              map((stores) => {
+                const store = !this.store
+                  ? stores?.[0]
+                  : stores.find((store) => store.id === this.store) ??
+                    stores?.[0];
+                this.store$.next(store);
+                return this.store$.value;
+              })
             )
           : of(store)
       )
--- a/libs/domain/site/src/services/store/default-store.service.ts
+++ b/libs/domain/site/src/services/store/default-store.service.ts
@@ -14,11 +14,10 @@ export class DefaultStoreService implements StoreService {
   protected activeStore$ = new BehaviorSubject<string | undefined>(this.store);
   protected stores$ = this.adapter.get().pipe(shareReplay(1));
   protected store$ = combineLatest([this.stores$, this.activeStore$]).pipe(
-    map(
-      ([stores, activeStore]) =>
-        (activeStore
-          ? stores?.find((store) => store.id === activeStore)
-          : undefined) ?? stores?.[0]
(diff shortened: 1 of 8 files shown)


libs/domain/site/src/services/store/default-store.service.ts
```diff
@@ -31,10 +31,14 @@ export class DefaultStoreService implements StoreService {
...
-                (stores) =>
-                  stores.find((store) => store.id === this.store) ?? null
-              )
+              map((stores) => {
+                const store = !this.store
+                  ? stores?.[0]
+                  : stores.find((store) => store.id === this.store) ??
+                    stores?.[0];
+                this.store$.next(store);
+                return this.store$.value;
```
dunqan: It would be safer to `return EMPTY` here I believe, as `this.store$.next(store);` will trigger the whole stream anyway.

libs/domain/site/src/services/store/default-store.service.ts
```diff
@@ -1,21 +1,26 @@
...
   protected stores$ = this.adapter.get().pipe(shareReplay(1));
-  protected store$ = new BehaviorSubject<Store | null>(null);
-  protected subscription = new Subscription();
+  protected store$ = combineLatest([this.stores$, this.activeStore$]).pipe(
+    map(([stores, activeStore]) =>
+      activeStore
+        ? stores?.find((store) => store.id === activeStore)
+        : stores?.[0]
+    ),
+    shareReplay(1)
```
supproduction:
```suggestion
    shareReplay({refCount: true, bufferSize: 1})
```

supproduction (changes requested):

supproduction (dismissed):

dunqan (dismissed):

libs/domain/site/src/services/store/default-store.service.ts
```diff
@@ -1,21 +1,26 @@
...
   protected stores$ = this.adapter.get().pipe(shareReplay(1));
-  protected store$ = new BehaviorSubject<Store | null>(null);
-  protected subscription = new Subscription();
+  protected store$ = combineLatest([this.stores$, this.activeStore$]).pipe(
+    map(([stores, activeStore]) =>
+      activeStore
+        ? stores?.find((store) => store.id === activeStore)
+        : stores?.[0]
+    ),
+    shareReplay({refCount: true, bufferSize: 1})
```
supproduction:
```suggestion
    shareReplay({ refCount: true, bufferSize: 1 })
```

supproduction (changes requested):

dunqan (approved):

supproduction (approved):

PR #135, opened 2023-03-10
tests: added e2e tests for guest checkout

Ivan-Vidiaiev:
closes https://spryker.atlassian.net/browse/HRZ-2370 and https://spryker.atlassian.net/browse/HRZ-2437


AndreyYevtukhov (dismissed):

libs/domain/order/summary/src/summary.component.ts
```diff
@@ -78,11 +78,6 @@ export class OrderSummaryComponent extends OrderMixin(
       )}
       <div class="title">${i18n('order.payment')}:</div>
       <div>${this.order?.payments[0].paymentProvider}</div>
-      </div>
-      <div class="title">${i18n('order.email')}:</div>
-      <div>${this.order?.billingAddress?.email}</div>
```
Sinucid, line 83: Our US address schema contain this field, but we don't have this country in supported list. I would make this field at least optional for rendering to keep the data consistent
Screenshot 2023-03-15 at 09 43 30

Sinucid (approved):

libs/domain/order/summary/src/summary.component.ts
AndreyYevtukhov: It might be, but we should not display emails on `/order` page till we will have authorization and clear rules for it

PR #136, opened 2023-03-10
chore: product title refactor

tobi-or-not-tobi:
- use schema
- leverage heading styles
- complete stories


libs/domain/product/title/src/title.component.ts
```diff
@@ -14,12 +14,17 @@ export class ProductTitleComponent extends ProductMixin(
...
+  protected override render(): TemplateResult | void {
+    const options = this.componentOptions ?? {};
+    const { tag, as, asLg, asMd, asSm, maxLines } = this.componentOptions ?? {};
 
     return html`<oryx-heading
-      tag=${ifDefined(options?.tag)}
-      maxLines=${ifDefined(options?.maxLines)}
+      .tag=${tag}
+      .maxLines=${maxLines}
+      .as=${as}
```
Sinucid, line 24: This looks really like overhead. Mess of attributes which purpose is unclear. I would at least rename this attributes, for example: `level, levelSm, levelMd, ... size, sizeSm, ... , scale, scaleSm, ...`

And what about change the approach? Define bunch of attributes not for the host and for its slotted content, for example:

```
<oryx-heading>
  <h1 size="h2" sizeSm="h3">Heading</h1>
</oryx-heading>
```

or even more complex case, with different heading for different screen sizes:

```
<oryx-heading>
  <h1 hideSm>Heading</h1>
  <h1 showSm>Heading for SM</h1>
</oryx-heading>
```

Sinucid: Uh, now I realised that this could be difficult to configure in EB side. Never mind, what about just find better names for the attribures?

Sinucid (approved):

PR #137, opened 2023-03-13
docs: HRZ-2279 Release process docs

tolerants:
Updated release process docs

closes https://spryker.atlassian.net/browse/HRZ-2279

### DIFF
--- a/docs/internal/release/release-process.md
+++ b/docs/internal/release/release-process.md
@@ -1,0 +1,38 @@
+# Release process.
+
+## Regular release.
+
+This is a regular flow, when you are just following our usual versioning and there no additional conditions.   
+
+For a `next` release, run Release workflow from Github actions panel on `development` branch.
+For a stable release, do the following:
+
+1. Create branch from `development` for next minor/major as `release/v1.2.x`.
+2. Update `libs/lerna.json` version to a new minor/major version as `1.2.0`.
+3. Push the release branch.
+4. Trigger Release workflow from Github actions panel on release branch.
+   This creates an RC release.
+5. On `development` branch update `libs/lerna.json` version to a next minor version as `1.3.0`
+
+For a stable release on a release branch, do the following:
+
+1. Update `libs/lerna.json` version to remove RC tag `1.2.0-rc.4` -> `1.2.0`.
+2. Push the version to the release branch.
+3. Trigger Release workflow from Github actions panel on release branch.
+
+For a patch on a release branch, do the following:
+
+1. Push fixes to the release branch.
+2. Trigger Release workflow from Github actions panel on release branch.
+
+## Release from packages workflow.
+
+This workflow should be used in case regular workflow 
+
+1. Create branch from `development` for next minor/major as `release/v1.2.x`
+2. Set the version of every package in `libs` manually (ex. `1.2.0`).
+3. Update every `@spryker-oryx` `peerDependency` in every package in `libs` manually (ex. `1.2.0`).
+4. Update `libs/lerna.json` version `1.2.0`.
+5. Push the release branch.
+6. Trigger Release from packages workflow from Github actions panel on release branch.
+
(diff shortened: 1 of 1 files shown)


docs/internal/release/release-process.md
```diff
@@ -0,0 +1,38 @@
...
+
+This is a regular flow, when you are just following our usual versioning and there no additional conditions.   
+
+For a `next` release, run Release workflow from Github actions panel on `development` branch.
+For a stable release, do the following:
+
+1. Create branch from `development` for next minor/major as `release/v1.2.x`.
+2. Update `libs/lerna.json` version to a new minor/major version as `1.2.0`.
+3. Push the release branch.
+4. Trigger Release workflow from Github actions panel on release branch.
```
AndreyYevtukhov, line 13: Should it be done manually? 

We can have a rule like in https://github.com/spryker/oryx/blob/bb526e9cf991a7b751b920ec26ac484e690a45d5/.github/workflows/test.yml#L3-L7 to make it automatically on branches starting with `release/` to make it automatic

docs/internal/release/release-process.md
```diff
@@ -0,0 +1,37 @@
...
+
+For a stable release on a release branch, do the following:
+
+1. Update `libs/lerna.json` version to remove RC tag `1.2.0-rc.4` -> `1.2.0`.
+2. Push the version to the release branch.
+3. Trigger Release workflow from Github actions panel on release branch.
+
+For a patch on a release branch, do the following:
+
+1. Push fixes to the release branch.
```
AndreyYevtukhov, line 25: Will the version be automatically updated in this case? (1.2.0 -> 1.2.1 etc.)

docs/internal/release/release-process.md
```diff
@@ -0,0 +1,37 @@
...
+2. Trigger Release workflow from Github actions panel on release branch.
+
+## Release from packages workflow.
+
+This workflow should be used in case regular workflow
+
+1. Create branch from `development` for next minor/major as `release/v1.2.x`
+2. Set the version of every package in `libs` manually (ex. `1.2.0`).
```
AndreyYevtukhov, line 33: What about having a small script (bash/node) that will update everything for us?))) 

```
npm run bump-version 1.2.0
```

or something?)

AndreyYevtukhov: there are too many manual steps that should be done in many files - and it might lead to missed updates or unexpected errors

tolerants: The idea is that we are going to use regular flow where lerna taking care about it, so this flow is an exceptional case, when we have to do something specific and don't want to use regular flow.

docs/internal/release/release-process.md
tolerants: Yep.

docs/internal/release/release-process.md
tolerants: Yep, probably we can do it. Don't see any issues for now.

AndreyYevtukhov (dismissed):

docs/internal/release/release-process.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Regular release.
+
+This is a regular flow, when you are just following our usual versioning and there no additional conditions.
+
+For a `next` release, run Release workflow from Github actions panel on `development` branch.
+For a stable release, do the following:
+
+1. Create branch from `development` for next minor/major as `release/v1.2.x`.
+2. Update `libs/lerna.json` version to a new minor/major version as `1.2.0`.
```
dunqan: Shouldn't be ?
```suggestion
2. Update `libs/lerna.json` version to a new minor/major version as `1.2.0-rc.1`.
```

tolerants: But it's for stable release, not for release candidate.

AndreyYevtukhov (approved):

PR #157, opened 2023-03-15
feat: picking page tabs migration

aleespryker:
Closes [HRZ-2417](https://spryker.atlassian.net/browse/HRZ-2417)

[HRZ-2417]: https://spryker.atlassian.net/browse/HRZ-2417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -0,0 +1,114 @@
...
+        <oryx-chip dense>${tab.items?.length ?? '0'}</oryx-chip>
+      </oryx-tab>`
+    )}`;
+  }
+
+  protected renderTabContents(): TemplateResult {
+    return html`${repeat(
+      this.tabs,
+      (tab) => html`<div slot="panels" id="tab-${tab.id}">
+        ${tab.items && tab.items?.length > 0
```
Sinucid: `${tab.items?.length`

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -0,0 +1,114 @@
...
+    { id: ItemsFilters.Picked, title: 'picked', items: this.pickedItems },
+    {
+      id: ItemsFilters.NotFound,
+      title: 'not-found',
+      items: this.notFoundItems,
+    },
+  ];
+
+  protected override render(): TemplateResult {
+    this.tabs[0].items = this.notPickedItems;
```
Sinucid: Not a good decision to update something inside render method. It will perform this on each re-render, even if nothing changes in the data. Due to each tab contains just filtered items, you can perform that filtering just in tabs configuration:

```
protected pickingList$ = this.pickingId$.pipe(
    distinctUntilChanged(),
    switchMap((id) => this.pickingListService.getById(id ?? ''))
);

protected tabs$ = this.pickingList$.pipe(map(list => ([
    {
      id: ItemsFilters.NotPicked,
      title: 'not-picked',
      items: list?.items.filter((item) => item.status === ItemsFilters.NotPicked),
    },
    { id: ItemsFilters.Picked, 
       title: 'picked', 
       items: list?.items.filter(item => item.numberOfPicked && item.status === ItemsFilters.Picked),
    },
    {
      id: ItemsFilters.NotFound,
      title: 'not-found',
      filter: list?.items.filter(item => item.numberOfNotPicked || (item.numberOfPicked < item.quantity && item.status === ItemsFilters.NotFound)),
    },
  ])));

@asyncState()
protected tabs = valueType(this.tabs$);
```

Sinucid (changes requested):

libs/domain/picking/src/components/picking/picking.component.ts
Sinucid: Shame on me, I didn't test my suggestion. If something is unclear or broken - let me know

libs/domain/picking/src/services/picking-list-default.service.ts
```diff
@@ -1,27 +1,48 @@
...
+    new BehaviorSubject<PickingListQualifier | null>(null);
+  protected pickingListsState$ = new BehaviorSubject<PickingList[]>([]);
 
-  protected pickingLists$ = this.pickingListsQualifier$.pipe(
-    switchMap((qualifier) => this.adapter.get(qualifier)),
-    shareReplay({ bufferSize: 1, refCount: true })
+  protected pickingListsLoading$ = this.pickingListsQualifier$.pipe(
+    switchMap((qualifier) =>
+      this.adapter.get(
+        qualifier ?? {
```
Sinucid: Not sure regarding this. Why do we need to set default qualifier here? How can we fetch all orders then?

Sinucid (changes requested):

libs/domain/picking/src/services/picking-list-default.service.ts
aleespryker: Sorry, did the wrong revert. Should be fixed now. I never meant to touch that file.

Sinucid (dismissed):

PR #164, opened 2023-03-16
feat: add store interceptor

supproduction:
adds store interceptor

closes [HRZ-2450](https://spryker.atlassian.net/browse/HRZ-2450)

[HRZ-2450]: https://spryker.atlassian.net/browse/HRZ-2450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/site/src/services/store/store.interceptor.ts
+++ b/libs/domain/site/src/services/store/store.interceptor.ts
@@ -4,16 +4,15 @@ import {
   RequestOptions,
 } from '@spryker-oryx/core';
 import { inject, INJECTOR } from '@spryker-oryx/di';
-import { map, Observable, switchMap, take } from 'rxjs';
-import { LocaleService } from './locale.service';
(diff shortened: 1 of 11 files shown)


libs/domain/site/src/services/store/store.interceptor.ts
```diff
@@ -0,0 +1,57 @@
...
+        )
+      );
+  }
+
+  protected addStoreHeader(options: RequestOptions): RequestOptions {
+    return {
+      ...options,
+      headers: {
+        ...options.headers,
+        'X-Store': this.STORE,
```
dunqan: I think StoreService should be a source of truth to the interceptor, not env variable directly

dunqan (approved):

PR #171, opened 2023-03-20
Feature/hrz 2386 sb stories for migrated piking lists

tarashaliuk:
closes: [HRZ-2386](https://spryker.atlassian.net/browse/HRZ-2386)

[HRZ-2386]: https://spryker.atlassian.net/browse/HRZ-2386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/components/picking-list-item/stories/demo.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+  argTypes: {
+    pickingListId: {
+      options: pickingListIds,
+      control: { type: 'select' },
+    },
+  },
+} as Meta;
+
+interface Props {
+  pickingListId: string;
```
Sinucid: Create an interface for `PickingListItemComponent` please, and reuse it here as a type for props.

Sinucid (changes requested):

Sinucid (approved):

PR #172, opened 2023-03-20
feat: navigation item

Sinucid:
Navigation item for site domain. Allows to provide navigation menu elements for application. Along with it was implemented Basic TokenResolver service to resolve some resources by string token.

Tests are not included. Storybook integration is not included.

closes to [HRZ-2403](https://spryker.atlassian.net/browse/HRZ-2403)

[HRZ-2403]: https://spryker.atlassian.net/browse/HRZ-2403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/site/navigation-item/src/navigation-item.component.ts
```diff
@@ -0,0 +1,179 @@
...
+    return html`<experience-composition
+      .uid=${this.uid}
+      close-popover
+    ></experience-composition>`;
+  }
+
+  protected get icon(): TemplateResult {
+    if (!this.componentOptions?.icon) {
+      return html``;
+    }
```
tobi-or-not-tobi:
```suggestion
  protected get icon(): TemplateResult | void {
    if (!this.componentOptions?.icon) return;
```

tobi-or-not-tobi: or

```ts
protected get icon(): TemplateResult|void {
    if (this.componentOptions?.icon) {
      return html`<oryx-icon .type=${this.componentOptions?.icon}></oryx-icon>`;
    }
  }
```

libs/domain/site/navigation-item/src/navigation-item.component.ts
```diff
@@ -0,0 +1,179 @@
...
+        if (typeof url !== 'object') {
+          return of(url);
+        }
+
+        return resolve(SemanticLinkService).get(url);
+      })
+    )
+  );
+
+  protected onTriggerClick(): void {
```
tobi-or-not-tobi, line 76: `onTrigger()`?

libs/domain/site/navigation-item/src/navigation-item.component.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  protected renderCustomButton(): TemplateResult {
+    return html`
+      <oryx-site-navigation-button
+        slot="trigger"
+        url=${ifDefined(this.url)}
+        icon=${ifDefined(this.componentOptions?.icon)}
+        text=${this.label}
+        badge=${this.badge}
```
tobi-or-not-tobi:
```suggestion
        .url=${this.url}
        .icon=${this.componentOptions?.icon}
        .text=${this.label}
        .badge=${this.badge}
```

tobi-or-not-tobi (changes requested):

libs/domain/site/navigation-item/src/navigation-item.component.ts
Sinucid: Nope, not so clear. Also `onClick` doesn't make sense what click

libs/platform/experience/src/models/component-group.model.ts
```diff
@@ -0,0 +1,12 @@
+export const enum ComponentGroup {
```
dunqan: I'm wondering should we have a central place for this, especially here? As it would mean, every package will rely on 'experience' package. What do you think @tobi-or-not-tobi ?

libs/platform/core/src/services/token-resolver/token-resolver.service.ts
```diff
@@ -0,0 +1,21 @@
+import { Observable } from 'rxjs';
+
+export interface TokenResolver {
+  resolve(resolver: string): Observable<string | null>;
+}
+
+export type ResolvedToken = Observable<string | null>;
+export type Resolver = (...args: string[]) => ResolvedToken;
+
+export interface TokenResolverService {
```
dunqan: What about just `TokenService` ?

libs/platform/core/src/services/token-resolver/default-token-resolver.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export class DefaultTokenResolverService implements TokenResolverService {
+  protected resolvers = new Map<string, TokenResolver>();
+
+  protected isToken(resolver: string): boolean {
+    return tokenRE.test(resolver);
+  }
+
+  protected getResolverKey(tokenResolver: string): string {
+    return `${TokenResourceResolver}${tokenResolver}`;
```
dunqan: Ideally we will have an ability to provide multiple resolvers per namespace/domain. But it can be done in a followup, as this one is quite big anyway.

libs/platform/auth/src/components/button/button.def.ts
```diff
@@ -0,0 +1,15 @@
+import { componentDef } from '@spryker-oryx/core';
+import { AuthButtonOptions } from './button.model';
+
+declare global {
+  interface FeatureOptions {
+    'oryx-auth-button'?: AuthButtonOptions;
+  }
+}
+
+export const authButtonComponent = componentDef({
```
dunqan: Let's rename it to `LoginButtonComponent`.

libs/platform/experience/src/models/component-group.model.ts
Sinucid: I don't see problem. Each schema importing `ContentComponentSchema` interface from experience package. `ComponentGroup` won't be used outside schemas

```
import {
  ComponentGroup,
  ContentComponentSchema,
} from '@spryker-oryx/experience';
```

libs/platform/core/src/services/token-resolver/token-resolver.service.ts
Sinucid: Here is a mess of 'resolvers'.

I would say that `TokenResolverService` is this service. And  `TokenResolver` is a domain resolver. Both interfaces are presented. But agree, will try to find more logical names

libs/platform/auth/src/components/button/button.def.ts
Sinucid: But it doing both (login|logout) depends on auth state

dunqan: yup, we know :)

tobi-or-not-tobi: I also prefer `LoginButtonComponent`, we could have an option on it to allow for also supporting the logout behaviour. I just don't like `AuthButtonComponent`, as you cannot grasp what it does (register? password related stuff? etc).

libs/platform/experience/src/models/component-group.model.ts
dunqan: but ContentComponentSchema is a generic thing, while group is per domain/package.

tobi-or-not-tobi: Will we be able to add custom groups as strings? or are we locked into the groups provided by the framework? I like other groups to be added spontaneously.

libs/platform/auth/src/components/button/button.def.ts
tobi-or-not-tobi: Should we rename to `LoginTriggerComponent` to not bake in the UI type of this component?

tobi-or-not-tobi (approved):

PR #174, opened 2023-03-20
fix: indicate cart totals delivery message is not implemented

aleespryker:
Closes [HRZ-2353](https://spryker.atlassian.net/browse/HRZ-2353)

[HRZ-2353]: https://spryker.atlassian.net/browse/HRZ-2353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/cart/totals/src/totals.component.ts
+++ b/libs/domain/cart/totals/src/totals.component.ts
@@ -151,7 +151,7 @@ export class CartTotalsComponent extends CartComponentMixin<CartTotalsComponentO
             <small class="delivery-message">
               ${ifDefined(options?.deliveryMessage)}
               <oryx-icon-button size=${Size.Sm}>
-                <a href="#" title=${i18n('cart.totals.delivery-message')}>
+                <a href="#" title=${i18n('cart.totals.not-implemented-yet')}>
                   <oryx-icon type="info"></oryx-icon>
                 </a>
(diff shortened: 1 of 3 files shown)


libs/domain/cart/totals/src/totals.component.ts
```diff
@@ -151,7 +151,7 @@ export class CartTotalsComponent extends CartComponentMixin<CartTotalsComponentO
             <small class="delivery-message">
               ${ifDefined(options?.deliveryMessage)}
               <oryx-icon-button size=${Size.Sm}>
-                <a href="#" title=${i18n('cart.totals.delivery-message')}>
+                <a href="#" title=${i18n('cart.totals.not-implemented-yet')}>
```
Sinucid: We have an issue with using `i18n` as attribute directive on ssr. Replace content of the title by just a string. It will be enough for now. And pay attention, ticket contains additional requirements: 

`!!! When it is fixed → uncomment line 180 in file apps/storefront-e2e/src/integration/cart.cy.ts and run E2E tests to check if it works again. Also remove the comment in 179`

Sinucid (changes requested):

Sinucid (approved):

PR #176, opened 2023-03-21
chore: cart refactor

tobi-or-not-tobi:
- refactor to use asyncState rather than asyncValue
- align UI with Figma's
- optional notifications for update and removals
- drop incomplete cart entry features (variants, options)

closes [HRZ-2446](https://spryker.atlassian.net/browse/HRZ-2446)

[HRZ-2446]: https://spryker.atlassian.net/browse/HRZ-2446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/action/link/src/link.component.ts
```diff
@@ -15,6 +15,7 @@ export class LinkComponent
   @property({ type: Boolean, reflect: true }) disabled?: boolean;
   @property({ reflect: true }) icon?: Icons | string;
   @property({ reflect: true }) multiLine?: boolean;
+  @property({ reflect: true }) transparent?: boolean;
```
Sinucid: I didn't get why do we need this attribute? How will it work in combination with icon? It drops flexbox layout...
We can apply this style to the oryx-link where it is necessary. Debatable I would say

libs/base/ui/overlays/modal/src/modal.component.ts
```diff
@@ -180,6 +180,7 @@ export class ModalComponent extends LitElement implements ModalProperties {
                 <oryx-button type="secondary" outline size="sm">
                   <button value="cancel">Cancel</button>
                 </oryx-button>
+                <slot name="footer-more"></slot>
```
Sinucid, line 183: slot that is slotted to the slot? What is the aim?

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,163 @@
...
-  entries: CartEntry[];
-  loading: boolean;
-  currentlyUpdated: string | null;
-  options: Partial<CartEntriesOptions>;
-}
-
+import { cartEntriesStyles } from './entries.styles';
+
+@defaultOptions({
+  silentRemove: false,
```
Sinucid: Just an idea

I'm wondering, what if we move default options set to the model, or even create additional file for this, like: `entries.options.ts` or `entries.default-options.ts` and in case if we need to override only some parts of them when extends the component or even on app level we can merge them with this imports?
```
@defaultOptions({
  ..defaultOptions,
  silentRemove: true
})
```

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,163 @@
...
     return html`
-      <section>
-        <oryx-icon type="cart"></oryx-icon>
-        <p>Your shopping cart is empty</p>
-        <oryx-button>
-          <button>Shop now</button>
-        </oryx-button>
-      </section>
+      <oryx-heading>
+        <h1>${i18n('cart.totals.<count>-items', { count: 7 })}</h1>
```
Sinucid: count: 7?

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,163 @@
...
+      </oryx-heading>
+
+      ${repeat(
+        this.entries ?? [],
+        (entry) => entry.groupKey,
+        (entry) =>
+          html`
+            <oryx-cart-entry
+              .key=${entry.groupKey}
```
Sinucid, line 46: `key` prop name not so clear. groupKey I guess will be better

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,163 @@
...
-            updating: entry.groupKey === currentlyUpdated,
-          }}
-          @submit=${(e: CustomEvent): void => this.onUpdate(e, entry)}
-          @oryx.remove=${(): void => this.onRemove(entry)}
-        ></cart-entry>
-      `
-    )}`;
+  constructor() {
+    super();
+    this.addEventListener('submit', this.onSubmit as EventListener);
```
Sinucid: Does onSubmit really work without binding the context? And I believe that LIT guideline suggest to add event listeners inside connected callback

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,163 @@
...
+  /**
+   * Handles updates on the entry. When the quantity is 0, the entry is going
+   * to be removed unless the component options require to seek for confirmation first.
+   */
+  protected onSubmit(ev: CustomEvent<CartEntryChangeEventDetail>): void {
+    const { groupKey, quantity } = ev.detail;
+
+    if (ev.detail.quantity === 0) {
+      if (this.componentOptions?.silentRemove) {
+        this.removeEntry(ev.detail.groupKey);
```
Sinucid: `this.removeEntry(groupKey);`

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,163 @@
...
-    if (options.hideItemsCount) {
-      return html``;
+  /**
+   * Handles updates on the entry. When the quantity is 0, the entry is going
+   * to be removed unless the component options require to seek for confirmation first.
+   */
+  protected onSubmit(ev: CustomEvent<CartEntryChangeEventDetail>): void {
+    const { groupKey, quantity } = ev.detail;
+
+    if (ev.detail.quantity === 0) {
```
Sinucid: `if (quantity === 0) {`

You've already extracted constants from details above

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,163 @@
...
+      if (this.componentOptions?.silentRemove) {
+        this.removeEntry(ev.detail.groupKey);
+      } else {
+        this.removeGroupKey = ev.detail.groupKey;
```
Sinucid: this.removeGroupKey = groupKey;

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,163 @@
...
+      }
+    } else {
+      this.cartService.updateEntry({ groupKey, quantity }).subscribe(() => {
+        if (this.componentOptions.notifyOnUpdate) {
```
Sinucid: A question. Should we split notification options for each case? Maybe single one will be enough? `notifyActions`

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,163 @@
...
+  }
 
-    const count = entries.reduce((acc, { quantity }) => quantity + acc, 0);
-
-    return html`
-      <oryx-chip>${count} ${count === 1 ? 'item' : 'items'}</oryx-chip>
-    `;
+  protected removeEntry(groupKey: string): void {
+    this.removeGroupKey = undefined;
+    const sku = this.entries?.find((entry) => entry.groupKey === groupKey)?.sku;
```
Sinucid: I believe that this part better to move to the complete callback of subscribe, as in `updateEntry`. We should drop removeGroupKey only if deletion is succesful

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,163 @@
...
+        <button value="remove">${i18n(`cart.entry.remove`)}</button>
+      </oryx-button>
+    </oryx-modal>`;
+  }
 
-        ${this.renderEntries(entriesData)}
-      </oryx-collapsible>
-    `;
+  protected resetConfirmation(): void {
+    this.requestUpdate();
```
Sinucid: Why do we need requestUpdate here? `removeGroupKey ` is a state, it should emit update

libs/domain/cart/entry/src/entry.component.ts
```diff
@@ -1,170 +1,203 @@
...
+  enableId: true,
+  enablePreview: true,
+} as CartEntryOptions)
+export class CartEntryComponent
+  extends CartComponentMixin(ContentMixin<CartEntryOptions>(LitElement))
+  implements CartEntryAttributes
+{
+  static styles = [cartEntryStyles];
+
+  @property() key?: string;
```
Sinucid, line 54: As is mentioned in cart entries, need to find better name

libs/domain/cart/entry/src/entry.component.ts
```diff
@@ -1,170 +1,203 @@
...
-        this.dispatchRemoveEvent();
-        return;
-      }
+  protected notificationService = resolve(NotificationService);
 
-      this.triggerConfirmationRequired$.next(true);
+  protected willUpdate(
+    changedProperties: PropertyValueMap<unknown> | Map<PropertyKey, unknown>
+  ): void {
+    this.entry = this.entries?.find((entry) => entry.groupKey === this.key);
```
Sinucid: it will make this changes on each update. Need to specify the conditions. I believe that it should perform reassigns and price formatting only if entries changes, but how can we achieve this, it is not an reactive property? Maybe better to restore previous approach where `entry` was observable that observes the `entries`?  This provides much more stable behaviour

libs/domain/cart/entry/src/styles/entry.styles.ts
```diff
@@ -0,0 +1,115 @@
+import { css } from 'lit';
+
+const pricing = css`
```
Sinucid, line 3: Why do you need to extract this to the separate const?

libs/domain/cart/src/mixins/cart.mixin.ts
```diff
@@ -1,22 +1,53 @@
...
+  superClass: T
+): Type<CartMixinInterface> & T => {
+  class CartMixinClass extends superClass {
+    @property({ reflect: true }) cartId?: string;
+    @property({ type: Object, reflect: true }) cart?: Cart;
+
+    protected cartController = new CartController(this);
 
-  return CartComponent as Type<LitElement & ContentComponentProperties<T>>;
+    @asyncState()
```
Sinucid, line 37: Debatable, but I think that we should define protected observable first and then pass it to the asyncState. This allows to have a bit more flexibility.

```
protected isEmpty$ = this.cartController.isEmpty();

@asyncState()
protected isEmpty = valueType(this.isEmpty$);

...

```

libs/domain/cart/summary/src/summary.component.ts
```diff
@@ -43,14 +42,22 @@ export class CartSummaryComponent extends CartComponentMixin<CartSummaryOptions>
...
+   * "99+".
+   */
+  protected renderMark(): TemplateResult | void {
+    const quantity =
+      this.totalQuantity &&
+      this.componentOptions?.maxVisibleQuantity &&
+      this.totalQuantity > this.componentOptions?.maxVisibleQuantity
+        ? `${this.componentOptions?.maxVisibleQuantity}+`
+        : Number(this.totalQuantity);
+    if (this.totalQuantity) {
```
Sinucid: `if (quantity) {` ?

libs/domain/content/link/src/link.component.ts
```diff
@@ -66,16 +68,17 @@ export class ContentLinkComponent extends ContentMixin<ContentLinkOptions>(
       if (options.button) {
         return html`<oryx-button
           part="wrapper"
-          icon="${ifDefined(options.icon?.trim())}"
+          icon="${ifDefined(icon?.trim())}"
```
Sinucid: trim? Why?

libs/domain/content/link/src/link.component.ts
```diff
@@ -66,16 +68,17 @@ export class ContentLinkComponent extends ContentMixin<ContentLinkOptions>(
...
         >
           ${this.renderLink(link, options)}
         </oryx-button>`;
       }
 
       return html`<oryx-link
-        icon="${ifDefined(options.icon?.trim())}"
-        ?disabled=${options?.disabled}
-        ?multiLine=${options.multiLine}
+        icon="${ifDefined(icon?.trim())}"
```
Sinucid: Same question here

Sinucid (changes requested):

libs/domain/cart/entry/src/entry.component.ts
Sinucid: This can be even a method for CartMixin: `getEntry(groupKey)`

libs/base/ui/action/link/src/link.component.ts
tobi-or-not-tobi: We need this to render links without the typical hover/active states. It's a start, and we'd ideally formalise it as a link type. I've asked UX team in slack channel for that. 

In the meantime, i've refactored it already to a `neutral` type on the link - agreed transparent was not a great one.

libs/base/ui/overlays/modal/src/modal.component.ts
tobi-or-not-tobi: The idea is to keep the cancel button while adding more buttons to the footer. It's perfectly fine to add nested slots imo, however, when the `footer` slot is provided, the `footer-more` slot is no longer available.

libs/domain/cart/entries/src/entries.component.ts
tobi-or-not-tobi: We could consider, let's bring it into the UI chat and resolve it here.

libs/domain/cart/entries/src/entries.component.ts
tobi-or-not-tobi: oops. fixed

libs/domain/cart/entries/src/entries.component.ts
tobi-or-not-tobi: I see the as 2 distinct things, especially when we allow for paralle cart entry modifications, it might be convenient to get the confirmation. but removing entries is pretty obvious.

libs/domain/cart/entry/src/styles/entry.styles.ts
tobi-or-not-tobi: just for better readability.

libs/domain/cart/src/mixins/cart.mixin.ts
tobi-or-not-tobi: I don't see the issue, you might want to challenge @dunqan outside the scope of this refactor as it's part of cart service refactor done earllier (just following a pattern here). Also, we're movig to signals which will change things again slightly.

libs/domain/cart/entry/src/entry.component.ts
tobi-or-not-tobi: For sure, I was about to ask @dunqan to support this in the service so we can remove it from the component. The mixin/controller could then be the natural inbetween layer. But not in this ticket. 

@dunqan FYI. we might want to add this going forward, perhaps when we do another run with signals.

libs/domain/cart/entries/src/entries.component.ts
tobi-or-not-tobi: I'm not sure. I think groupKey is not so clear, as it's not a common term in cart operations compared to other commerce systems. Hence, I was thinking to generalise it as just the `key`. It's similar to `id` or `code`, but `id` is reserved and `code` is typically a _logical_ id. I thought `key` is the right thing.

libs/domain/cart/entry/src/entry.component.ts
tobi-or-not-tobi: see former comment.

libs/domain/cart/entries/src/entries.component.ts
tobi-or-not-tobi: leftover

libs/domain/cart/entries/src/entries.component.ts
tobi-or-not-tobi: moved to connected callback. not sure why context should be provided, works fine.

libs/base/ui/overlays/modal/src/modal.component.ts
Sinucid: Ok, for now I don't have better ideas how to manage this other way

libs/domain/cart/entries/src/entries.component.ts
Sinucid: I don't see any changes

Sinucid (changes requested):

libs/domain/cart/src/mixins/cart.mixin.ts
```diff
@@ -1,22 +1,53 @@
...
+    protected isEmpty = valueType(this.cartController.isEmpty());
+
+    @asyncState()
```
dunqan, line 40: Wondering about all those props, I guess totals are not needed for most of components using this mixin, but every of those will subscribe and react to those (every) asyncState changes...

Signals could make it better, still overloading mixing with too many props (for specific use cases) doesn't feel like right approach

libs/domain/cart/summary/src/summary.component.ts
```diff
@@ -1,40 +1,39 @@
...
-  });
-
-  protected quantity$ = combineLatest([
-    this.cartController.getTotalQuantity(),
-    this.contentController.getOptions(),
-  ]).pipe(
-    map(([quantity, options]) => this.populateQuantity(quantity, options))
+  @asyncState()
+  protected link = valueType(
+    resolve(SemanticLinkService).get({
```
dunqan: resolving inside async state looks like too much IMO. Opinionated ofc, but I like to make it clean, which services compoennt uses

libs/domain/cart/entry/src/entry.component.ts
```diff
@@ -1,170 +1,206 @@
...
-  ): void {
-    if (e.detail.quantity === 0) {
-      e.stopPropagation();
+  @state() formattedPrice?: string;
 
-      if (isInstant) {
-        this.dispatchRemoveEvent();
-        return;
-      }
+  protected notificationService = resolve(NotificationService);
```
dunqan: left over?

libs/domain/cart/entry/src/entry.component.ts
```diff
@@ -1,170 +1,206 @@
...
 
-      this.triggerConfirmationRequired$.next(true);
+  protected willUpdate(
+    changedProperties: PropertyValueMap<unknown> | Map<PropertyKey, unknown>
+  ): void {
+    this.entry = this.entries?.find((entry) => entry.groupKey === this.key);
+    if (this.entry) {
+      this.context.provide(ProductContext.SKU, this.entry.sku);
+
+      this.pricingService
```
dunqan: we should move it to separate subscription, but let's hope signals will make it better.
Still I'd add a TODO here, to cover this case later

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,168 @@
...
 @hydratable('window:load')
-export class CartEntriesComponent extends CartComponentMixin<CartEntriesOptions>() {
-  static styles = styles;
+export class CartEntriesComponent extends CartComponentMixin(
+  ContentMixin<CartEntriesOptions>(LitElement)
+) {
+  static styles = cartEntriesStyles;
 
   protected cartService = resolve(CartService);
   protected productService = resolve(ProductService);
```
dunqan: left over?

libs/domain/cart/entry/src/entry.component.ts
```diff
@@ -1,170 +1,206 @@
...
+    return html`
+      <section class="pricing">
+        <oryx-cart-quantity-input
+          .min=${Number(!this.componentOptions?.removeByQuantity)}
+          .max=${this.available ?? Infinity}
+          .value=${this.entry?.quantity}
+          .decreaseIcon=${this.decreaseIcon}
+          submitOnChange
+          @submit=${this.onSubmit}
+          ?disabled=${this.isBusy}
```
dunqan, line 177: We are not leveraging busy per entry?

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,168 @@
...
+    if (quantity === 0) {
+      if (this.componentOptions?.silentRemove) {
+        this.removeEntry(groupKey);
+      } else {
+        this.removeGroupKey = groupKey;
+      }
+    } else {
+      this.cartService.updateEntry({ groupKey, quantity }).subscribe({
+        complete: () => {
+          if (this.componentOptions.notifyOnUpdate) {
```
dunqan: OK, to keep it. 
Just wondering if we need this logic in this component. 
We don't have global configuration for this, so basically, if you'll have more compnents that will update something in the cart, each of them should have this option. 

The same, if we would like to move removing to entry component (wihch I wonder, shouldn't be a case, then entries would just enumerate through every entry, but I guess you did consider it).

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,168 @@
...
+
+    if (quantity === 0) {
+      if (this.componentOptions?.silentRemove) {
+        this.removeEntry(groupKey);
+      } else {
+        this.removeGroupKey = groupKey;
+      }
+    } else {
+      this.cartService.updateEntry({ groupKey, quantity }).subscribe({
+        complete: () => {
```
dunqan: should probably be `next` as complete will also work for `error`

libs/domain/cart/entry/src/entry.component.ts
tobi-or-not-tobi: Mixins for the world!!

tobi-or-not-tobi: we'll leave it for mixins, I'll create a ticket.

libs/domain/cart/entry/src/entry.component.ts
tobi-or-not-tobi: yes not for now

libs/domain/cart/src/mixins/cart.mixin.ts
tobi-or-not-tobi: Moved to https://spryker.atlassian.net/browse/HRZ-2532

libs/domain/cart/entry/src/entry.component.ts
tobi-or-not-tobi: Moved to https://spryker.atlassian.net/browse/HRZ-2532

libs/domain/cart/entries/src/entries.component.ts
tobi-or-not-tobi: covered in https://spryker.atlassian.net/browse/HRZ-2532

libs/domain/cart/entries/src/entries.component.ts
dunqan: don't see a change?

libs/base/ui/structure/heading/src/styles/screen.styles.ts
```diff
@@ -134,6 +140,12 @@ const mediumScreen = css`
     --fw: var(--capw);
     --lh: var(--capl);
   }
+
+  :host([as-md='small']) {
```
supproduction, line 144: Please, use enum here

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,168 @@
...
-}
-
+import { cartEntriesStyles } from './entries.styles';
+
+@defaultOptions({
+  silentRemove: false,
+  removeByQuantity: 'showBin',
+  notifyOnUpdate: false,
+  notifyOnRemove: true,
+} as CartEntriesOptions)
```
supproduction, line 15: Do we need typecasting here? As we will fix it with ts 5?

supproduction: Or add TODO then that we will delete it after updating?

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,168 @@
...
+      </oryx-heading>
+
+      ${repeat(
+        this.entries ?? [],
+        (entry) => entry.groupKey,
+        (entry) =>
+          html`
+            <oryx-cart-entry
+              .key=${entry.groupKey}
+              ?readonly=${this.componentOptions.readonly}
```
supproduction: Please, use optional chaining for `componentOptions`. Without it we might have errors on eb side when we dynamically add this component

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,168 @@
...
+  protected onSubmit(ev: CustomEvent<CartEntryChangeEventDetail>): void {
+    const { groupKey, quantity } = ev.detail;
+
+    if (quantity === 0) {
+      if (this.componentOptions?.silentRemove) {
+        this.removeEntry(groupKey);
+      } else {
+        this.removeGroupKey = groupKey;
+      }
+    } else {
```
supproduction:
```suggestion
    if (quantity !== 0) {
     	....
     	return;
    }
    
     if (quantity === 0 && this.componentOptions?.silentRemove) {
        this.removeEntry(groupKey);
        return
      } 
      
      this.removeGroupKey = groupKey;
```

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,168 @@
...
-    const count = entries.reduce((acc, { quantity }) => quantity + acc, 0);
-
-    return html`
-      <oryx-chip>${count} ${count === 1 ? 'item' : 'items'}</oryx-chip>
-    `;
+  protected removeEntry(groupKey: string): void {
+    this.removeGroupKey = undefined;
+    const sku = this.entries?.find((entry) => entry.groupKey === groupKey)?.sku;
+    this.cartService.deleteEntry({ groupKey }).subscribe({
+      complete: () => {
```
supproduction: Same as @dunqan has mentioned, please, use `next` instead of `complete`

supproduction (changes requested):

libs/base/ui/structure/heading/src/styles/screen.styles.ts
tobi-or-not-tobi: we're goign to rip it apart again, so let's not worry for now

supproduction (dismissed):

Sinucid (dismissed):

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,67 @@
...
-import { ProductService } from '@spryker-oryx/product';
-import { asyncValue, hydratable } from '@spryker-oryx/utilities';
-import { html, TemplateResult } from 'lit';
+import { CartComponentMixin } from '@spryker-oryx/cart';
+import { ContentMixin, defaultOptions } from '@spryker-oryx/experience';
+import { hydratable, i18n } from '@spryker-oryx/utilities';
+import { html, LitElement, TemplateResult } from 'lit';
 import { repeat } from 'lit/directives/repeat.js';
-import { BehaviorSubject, combineLatest, map, of, switchMap, tap } from 'rxjs';
+import { RemoveByQuantity } from '../../entry/src';
```
supproduction: If it's cross package you can use alises `@spryker-oryx/...`

libs/domain/cart/entries/src/entries.model.ts
```diff
@@ -1,29 +1,3 @@
-import { CartEntry } from '@spryker-oryx/cart';
+import { CartEntryOptions } from '../../entry/src';
```
supproduction: Please, use alias

libs/domain/cart/entries/src/entries.schema.ts
```diff
@@ -0,0 +1,12 @@
+import { ContentComponentSchema } from '@spryker-oryx/experience';
+import { cartEntryComponentSchema } from '../../entry/src/entry.schema';
```
supproduction, line 2: Can we use alias here?

libs/domain/cart/entries/src/stories/entries-demo.stories.ts
```diff
@@ -5,24 +5,15 @@ import { CartEntriesOptions } from '../entries.model';
 
 export default {
   title: `${storybookPrefix}/Entries`,
-  args: {
-    readonly: false,
-    collapsible: false,
-    expanded: false,
-    hideItemsCount: false,
-  },
+  // parameters: {
```
supproduction: Is it lefrover?

libs/domain/cart/entry/src/entry.schema.ts
```diff
@@ -0,0 +1,36 @@
+import { ContentComponentSchema } from '@spryker-oryx/experience';
+import { FormFieldType } from '@spryker-oryx/form';
+import { CartEntryComponent } from './entry.component';
+
+export const cartEntryComponentSchema: ContentComponentSchema<CartEntryComponent> =
+  {
+    type: 'oryx-cart-entry',
```
supproduction: We don't need type here

libs/domain/cart/entries/src/entries.schema.ts
```diff
@@ -0,0 +1,12 @@
...
+import { CartEntriesComponent } from './entries.component';
+
+export const cartEntriesComponentSchema: ContentComponentSchema<CartEntriesComponent> =
+  {
+    type: 'oryx-cart-entries',
```
supproduction: We don't need type here

libs/domain/cart/summary/src/summary.component.ts
```diff
@@ -43,14 +42,22 @@ export class CartSummaryComponent extends CartComponentMixin<CartSummaryOptions>
...
+   *
+   * When the quantity is larger that the allowed max visible,
+   * the quantity is _truncated_ to the `maxVisibleQuantity`, e.g.
+   * "99+".
+   */
+  protected renderMark(): TemplateResult | void {
+    const quantity =
+      this.totalQuantity &&
+      this.componentOptions?.maxVisibleQuantity &&
+      this.totalQuantity > this.componentOptions?.maxVisibleQuantity
```
supproduction, line 56: This might lead to issues with ts 5
```
Number(this.totalQuantity) > this.componentOptions?.maxVisibleQuantity
```

libs/domain/cart/summary/src/summary.component.ts
supproduction:
```suggestion
    if (Number(this.totalQuantity)) {
```

libs/domain/cart/summary/src/summary.schema.ts
```diff
@@ -0,0 +1,14 @@
+import { ContentComponentSchema } from '@spryker-oryx/experience';
+import { FormFieldType } from '@spryker-oryx/form';
+import { CartSummaryComponent } from './summary.component';
+
+export const cartSummaryComponentSchema: ContentComponentSchema<CartSummaryComponent> =
+  {
+    type: 'oryx-cart-summary',
```
supproduction: Type

libs/domain/cart/entries/src/entries.component.ts
```diff
@@ -1,169 +1,67 @@
...
-}
-
+import { cartEntriesStyles } from './entries.styles';
+
+@defaultOptions({
+  removeByQuantity: RemoveByQuantity.ShowBin,
+  enableItemId: true,
+  enableItemImage: true,
+  enableItemPrice: true,
+} as CartEntriesOptions)
```
supproduction, line 15: We don't need typecasting here

libs/domain/cart/entry/src/entry.component.ts
```diff
@@ -1,170 +1,267 @@
...
+
+/**
+ * Supports updating the quantity as well as removing the entry entirely.
+ */
+@defaultOptions({
+  removeByQuantity: RemoveByQuantity.ShowBin,
+  enableItemImage: true,
+  enableItemId: true,
+  enableItemPrice: true,
+} as CartEntryOptions)
```
supproduction, line 43: We don't need typecasting here

libs/domain/cart/totals/src/totals.component.ts
```diff
@@ -1,88 +1,85 @@
...
 
+@defaultOptions({
+  enableDiscounts: true,
+  enableExpense: true,
+  enableSubtotal: true,
+  enableTaxAmount: true,
+  enableDelivery: true,
+  enableTaxMessage: true,
+  deliveryMessage: 'not yet implemented',
+} as CartTotalsComponentOptions)
```
supproduction, line 25: Typecasting is not needed

libs/domain/cart/totals/src/totals.schema.ts
```diff
@@ -0,0 +1,28 @@
+import { ContentComponentSchema } from '@spryker-oryx/experience';
+import { FormFieldType } from '@spryker-oryx/form';
+import { CartTotalsComponent } from './totals.component';
+
+export const cartTotalsComponentSchema: ContentComponentSchema<CartTotalsComponent> =
+  {
+    name: 'oryx-cart-totals',
```
supproduction:
```suggestion
    name: 'Cart Totals',
```

libs/domain/order/entries/src/entries.component.ts
```diff
@@ -17,7 +17,10 @@ import { styles } from './entries.styles';
 @defaultOptions({
   limit: 5,
   threshold: 3,
-})
+  enableItemImage: true,
+  enableItemId: true,
+  enableItemPrice: false,
+} as OrderEntriesOptions)
```
supproduction, line 23: Typecasting

supproduction (changes requested):

libs/domain/cart/totals/src/totals.component.ts
tobi-or-not-tobi: but very useful as long as we refactor; keeping it until we have proper typing in place.

libs/domain/order/entries/src/entries.component.ts
tobi-or-not-tobi: but very useful as long as we refactor; keeping it until we have proper typing in place.

libs/domain/cart/entry/src/entry.component.ts
tobi-or-not-tobi: but very useful as long as we refactor; keeping it until we have proper typing in place.

libs/domain/cart/entries/src/entries.component.ts
tobi-or-not-tobi: but very useful as long as we refactor; keeping it until we have proper typing in place.

libs/domain/cart/entries/src/entries.schema.ts
tobi-or-not-tobi: are you sure?

libs/domain/cart/entries/src/entries.schema.ts
tobi-or-not-tobi: how?

libs/domain/cart/entries/src/entries.component.ts
tobi-or-not-tobi: but it's same package

libs/domain/cart/entries/src/entries.model.ts
tobi-or-not-tobi: how?

libs/domain/cart/entries/src/entries.component.ts
supproduction: Assume you need `@spryker-oryx/cart/entry` while you in `@spryker-oryx/cart/entries`

libs/domain/cart/entries/src/entries.model.ts
supproduction: `@spryker-oryx/cart/entry`

libs/domain/cart/entries/src/entries.schema.ts
supproduction: `@spryker-oryx/cart/entry`

supproduction: ah or we don't expose schema as public API, right?

apps/storefront-e2e/src/support/page_fragments/cart-entry.fragment.ts
```diff
@@ -14,20 +14,10 @@ export class CartEntryFragment {
...
-      .find('.col cart-entry-price')
-      .eq(1)
-      .find('[part="price"]');
-  getTotalPrice = () =>
-    this.getWrapper().find('cart-entry-totals').find('[part="price"]');
+  getPrice = () => this.getWrapper().find('oryx-product-price');
+  getSubtotal = () => this.getWrapper().find('.entry-price');
   getRemoveBtn = () => this.getWrapper().find('[aria-label="remove"]');
-  getConfirmRemoveBtn = () => this.getWrapper().contains('Delete product(s)');
+  // getConfirmRemoveBtn = () => this.getWrapper().contains('Delete product(s)');
```
supproduction: Leftover?

libs/domain/cart/summary/src/summary.component.ts
```diff
@@ -43,14 +42,22 @@ export class CartSummaryComponent extends CartComponentMixin<CartSummaryOptions>
...
+   * "99+".
+   */
+  protected renderMark(): TemplateResult | void {
+    const quantity =
+      this.totalQuantity &&
+      this.componentOptions?.maxVisibleQuantity &&
+      +this.totalQuantity > this.componentOptions?.maxVisibleQuantity
+        ? `${this.componentOptions?.maxVisibleQuantity}+`
+        : Number(this.totalQuantity);
+    if (Number(this.totalQuantity)) {
```
supproduction: Why in one cases `+` in other `Number`?

supproduction (changes requested):

libs/domain/cart/summary/src/summary.component.ts
tobi-or-not-tobi: That's the design, where there are more than the configured number, we show "99+" in the cart summary.

supproduction (approved):

libs/domain/cart/src/controllers/cart.controller.ts
```diff
@@ -21,22 +23,52 @@ export class CartController {
     this.observe = new ObserveController(host);
   }
 
+  protected get cartQualifier(): Observable<CartQualifier | undefined> {
+    return this.observe
+      .get('cartId')
+      .pipe(switchMap((cartId) => of(cartId ? { cartId } : undefined)));
```
supproduction, line 29:
```suggestion
      .pipe(switchMap((cartId) => of(cartId ? { cartId } : undefined)));
```
Why do we need this part?

libs/domain/cart/summary/src/summary.component.ts
supproduction: I am talking about these parts
image

supproduction (changes requested):

supproduction (dismissed):

supproduction (approved):

PR #179, opened 2023-03-21
feat: modal full width footer button

aleespryker:
Closes [HRZ-2521](https://spryker.atlassian.net/browse/HRZ-2521)

[HRZ-2521]: https://spryker.atlassian.net/browse/HRZ-2521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/overlays/modal/src/modal.styles.ts
```diff
@@ -60,4 +60,9 @@ export const styles = css`
   oryx-icon-button:first-child button {
     padding: 0;
   }
+
+  ::slotted(oryx-button[full-width]),
+  oryx-button[full-width] {
+    width: 100%;
+  }
```
tarashaliuk: I guess we can use just CSS to make the button full width without using extra full-width atrrbute
```
:host([footerButtonFullWidth]) footer oryx-button,
:host([footerButtonFullWidth])::slotted(oryx-button) {
  width: 100%;
}

Comment (tarashaliuk) [reply libs/base/ui/overlays/modal/src/modal.styles.ts]: Also, we can set the gap between buttons if there will be more than one button

**Review by tarashaliuk — CHANGES_REQUESTED** [review]

File: libs/base/ui/overlays/modal/src/stories/modal-demo.stories.ts [hunk 2]
```diff
@@ -26,6 +26,9 @@ interface Props {
 
   firstModalPreventCloseByEscape?: boolean;
   secondModalPreventCloseByEscape?: boolean;
+
+  firstFooterButtonFullWidth?: boolean;
+  secondFooterButtonFullWidth?: boolean;
```
tarashaliuk, line 31: Can you add static stories with full width button?

tarashaliuk (changes requested):

libs/base/ui/overlays/modal/src/modal.component.spec.ts
```diff
@@ -254,6 +254,20 @@ describe('Modal', () => {
...
+    beforeEach(async () => {
+      element = await fixture(
+        html`<oryx-modal enableFooter footerButtonFullWidth></oryx-modal>`
+      );
+    });
+
+    it('should add attribute to oryx-button', () => {
+      expect(
+        element.renderRoot.querySelector('oryx-button[full-width]')
+      ).not.toBeNull();
```
gund: Better to use `.toBeTruthy()` instead of inversions:
```suggestion
      ).toBeTruthy();
```

gund (commented):
Looks good, small comment for unit tests.

libs/base/ui/overlays/modal/src/modal.styles.ts
gund: I like this idea too, seems like an easier way to achieve the same visuals.

aleespryker: Seems the footer already supplies a gap.

libs/base/ui/overlays/modal/src/modal.component.spec.ts
aleespryker: Now that we don't have that attribute on the oryx-button, we don't need this test anymore.

libs/base/ui/overlays/modal/src/stories/modal-demo.stories.ts
tarashaliuk: still can't see static stories

aleespryker: Forgot to add the file. It's there now.

libs/base/ui/overlays/modal/src/stories/static/footer-button-full-width.stories.ts
```diff
@@ -0,0 +1,22 @@
...
+} as Meta;
+
+const Template: Story = (): TemplateResult => {
+  return html`
+    <oryx-modal open enableFooter footerButtonFullWidth>
+      <p>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+        tempor incididunt ut labore et dolore magna aliqua
+      </p>
+    </oryx-modal>
```
tarashaliuk, line 18: I would like to see different cases:
1) default button
2) one slotted button
3) two slotted buttons

tarashaliuk (changes requested):

tarashaliuk (approved):

gund (approved):

PR #180, opened 2023-03-21
feat: implement product card and picking logic

nsilva-spryker:
-Add new product card to pick items
-Adapte logic on list to handle each event of edit, submit and change items

Closes [HRZ-2411](https://spryker.atlassian.net/browse/HRZ-2411)

[HRZ-2411]: https://spryker.atlassian.net/browse/HRZ-2411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,262 @@
...
+import { BehaviorSubject } from 'rxjs';
+import {
+  ItemsFilters,
+  PickingListItem,
+  ProductItemPickedEvent,
+  SummaryInfo,
+} from '../../models';
+import { styles } from './picking-product-card.styles';
+
+export const EVENT_CHANGE_NUMBER_OF_PICKED = 'oryx.change-number-of-picked';
```
Sinucid: better to keep aliases for events inside model

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,262 @@
...
+export const EVENT_SUBMIT = 'oryx.submit';
+export const EVENT_EDIT = 'oryx.edit';
+
+export class PickingProductCardComponent extends LitElement {
+  static styles = styles;
+
+  @property() productItem?: PickingListItem;
```
Sinucid, line 20: I believe that we should implement an interface for this public properties

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,262 @@
...
+        () => html`
+          <oryx-card>
+            <div slot="heading" class="title">
+              ${this.productItem?.orderItem?.name}
+            </div>
+            <div class="subtitle">
+              <span>${this.productItem?.orderItem.sku}</span>
+            </div>
+
+            <img
```
Sinucid: you can use oryx-image. It also supports fallback image

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,262 @@
...
+
+export class PickingProductCardComponent extends LitElement {
+  static styles = styles;
+
+  @property() productItem?: PickingListItem;
+  @property() status: string = ItemsFilters.Picked;
+
+  @state() isCorrectNumberOfPickedProvided?: boolean = true;
+
+  protected currentNumberOfPicked$ = new BehaviorSubject(0);
```
Sinucid: Why are the observables? Looks like just lit states is enough here

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,262 @@
...
+      this.dispatchPickingEvents(EVENT_SUBMIT, {
+        productId: this.productItem?.product.id,
+      } as ProductItemPickedEvent);
+    } else {
+      this.isConfirmPickingDialogOpen$.next(true);
+    }
+  }
+
+  protected confirmPartialPicking(): void {
+    this.dispatchPickingEvents(EVENT_CHANGE_NUMBER_OF_PICKED, {
```
Sinucid: I believe that 2 separate events dispatching are overhead. Can submit also provides a number of picked in details?

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,262 @@
...
+              () => this.renderEditStatus()
+            )}
+          </oryx-card>
+        `
+      )}
+    `;
+  }
+
+  protected renderEditStatus(): TemplateResult {
+    this.summaryInfo = this.getSummaryInfo();
```
Sinucid: Why do you need to store it to the property? This doesn't have sense

Sinucid:
```
protected renderEditStatus(): TemplateResult {
    let text = '';
    let label = '';
    let subtext = '';
    
    if (this.status === ItemsFilters.Picked) {
        text = `${this.productItem.numberOfPicked}/${this.productItem.quantity}`;

        if (this.productItem.numberOfPicked < this.productItem.quantity) {
            label = i18n('picking.product-card.items-picked');
        } else {
            subtext = i18n('picking.product-card.all-items-picked')
        }
    } else if (this.status === ItemsFilters.NotFound) {
        text = `${this.productItem.numberOfNotPicked}/${this.productItem.quantity}`;

        if (this.productItem.numberOfNotPicked < this.productItem.quantity) {
            label = i18n('picking.product-card.items-not-found');
        } else {
            subtext = i18n('picking.product-card.no-items-found')
        }
    }

    return html` <div class="summary-info">
        <p>${text} ${label}</p>
        ${when(
            subtext,
            () => <p>${subtext}</p>
        )}
      </div>
      <oryx-button>
        <button @click=${this.editProductPicking}>
          <oryx-icon type="edit"></oryx-icon>
          ${i18n('picking.product-card.edit-items')}
        </button>
      </oryx-button>`;
  }
```

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,262 @@
...
+    return html`${this.renderPickingProduct()} ${this.renderConfirmationModal()}`;
+  }
+
+  protected renderPickingProduct(): TemplateResult {
+    return html`
+      ${when(
+        this.productItem,
+        () => html`
+          <oryx-card>
+            <div slot="heading" class="title">
```
Sinucid: can we use oryx-heading here?

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,262 @@
...
+              ${this.productItem?.orderItem?.name}
+            </div>
+            <div class="subtitle">
```
Sinucid: Why do you need to put span inside div?

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,262 @@
...
+                    ref="numberOfPickedInput"
+                    min="0"
+                    .max="${this.productItem?.quantity}"
+                    .value="${this.productItem?.numberOfPicked}"
+                    submitOnChange
+                    @update=${this.onChangeQuantity}
+                    @submit=${this.onChangeQuantity}
+                  ></oryx-cart-quantity-input>
+
+                  <div class="edit-quantity-info">
```
Sinucid: The same, span inside div

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,262 @@
...
+                    <span
+                      >${i18n('picking.product-card.of')}
+                      ${this.productItem?.quantity}
+                      ${i18n('picking.product-card.items')}</span
+                    >
+                  </div>
+
+                  <oryx-button>
+                    <button
+                      type="submit"
```
Sinucid: not needed, button without type works as submit

libs/domain/picking/src/components/picking-product-card/picking-product-card.styles.ts
```diff
@@ -0,0 +1,84 @@
...
+    color: var(--oryx-color-ink);
+  }
+
+  .image {
+    height: 106px;
+    margin-block-start: 5px;
+    object-fit: cover;
+    max-width: 182px;
+
+    &__fade {
```
Sinucid: left over?

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -3,26 +3,41 @@ import { asyncState, i18n, observe, valueType } from '@spryker-oryx/utilities';
...
   );
 
-  protected tabs$ = this.pickingList$.pipe(
-    map((list) => [
+  @asyncState()
+  protected pickingList = valueType(this.pickingList$);
+
+  protected pickingUpdate$ = new BehaviorSubject<string | null>(null);
+
+  protected tabs$ = combineLatest([
```
Sinucid: Why do you need to combine them?

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -53,6 +68,46 @@ export class PickingComponent extends LitElement {
   @asyncState()
   protected tabs = valueType(this.tabs$);
 
+  protected changeNumberOfPicked(event: Event): void {
+    const detail = (event as CustomEvent<ProductItemPickedEvent>).detail;
+
+    this.pickingList?.items.map((item) => {
+      if (
```
Sinucid: this is working not properly, this changing elements of reactive array doesn't trigger update. You need to ensure whole approach

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -53,6 +68,46 @@ export class PickingComponent extends LitElement {
...
+        item.product.id === detail.productId &&
```
Sinucid: item.product.id === detail.productId && detail.numberOfPicked && detail.numberOfPicked <= item.quantity

Sinucid (changes requested):

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
nsilva-spryker: What are the benefits of having an interface for just this two properties?

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
nsilva-spryker: It make sense to separated into getSummaryInfo() method, by doing that the code are more modular and easier to understand

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
Sinucid: This is public interface for the element. This allows us:
1) To get available properties and their types when query the element
2) Use this interface for storybook as a type for storybook properties

But you might be right. If there are no stories for this component, this can be useless. But if it requires stories - better to create interface and reuse it. Up to you

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  protected summaryInfo: SummaryInfo | undefined;
+
+  protected onSubmit(e: SubmitEvent): void {
+    e.preventDefault();
+
+    this.currentNumberOfPicked =
+      this.currentNumberOfPicked ?? this.productItem?.numberOfPicked;
+
+    if (this.productItem?.product.id) {
```
Sinucid:
```
this.dispatchPickingEvents(EVENT_SUBMIT, {
        productId: this.productItem.product.id!,
        numberOfPicked: this.currentNumberOfPicked,
});
```

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,184 @@
...
+      this.dispatchPickingEvents(EVENT_SUBMIT, {
+        productId: this.productItem.product.id,
+        numberOfPicked: this.currentNumberOfPicked,
+      });
+    }
+  }
+
+  protected editProductPicking(): void {
+    this.dispatchPickingEvents(EVENT_EDIT, {
+      productId: this.productItem?.product.id,
```
Sinucid: `productId: this.productItem.product.id!,`

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
Sinucid: I don't think so. This has sense if you are reusing this part somewhere or its to long. My version is contains ~30 rows that is not so long and fits one page in editor. Not a case I would say

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -1,53 +1,142 @@
...
+  static styles = styles;
+
+  protected isConfirmPickingDialogOpen$ = new BehaviorSubject(false);
+  @asyncState()
+  protected isConfirmPickingDialogOpen = valueType(
+    this.isConfirmPickingDialogOpen$
+  );
+
+  @state()
+  protected partialPicking: PartialPicking | undefined;
```
Sinucid: `protected partialPicking: PartialPicking | null = null;`

You can use this one as a trigger for confirmation modal. It will contain partial picking data or null. When it has data -> show modal. And you can clean up `isConfirmPickingDialogOpen`

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -1,53 +1,142 @@
...
+  PickingTab,
+  ProductItemPickedEvent,
+} from '../../models';
+import { styles } from './picking.styles';
 export class PickingComponent extends PickingListMixin(LitElement) {
-  protected tabs$ = this.pickingList$.pipe(
-    map((list) => [
+  static styles = styles;
+
+  protected isConfirmPickingDialogOpen$ = new BehaviorSubject(false);
```
Sinucid: Why do you need observable here? What the point? 

Why not just a lit state?

```
@state()
protected isConfirmPickingDialogOpen = false;
```

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -56,29 +145,117 @@ export class PickingComponent extends PickingListMixin(LitElement) {
...
-                    </oryx-picking-product-card>`
-                )
-              : this.renderFallback()}
-          </div>`
-        )}`
-      : html``;
+  protected renderTabContents(tabs: PickingTab[]): TemplateResult {
+    return html`
+      ${when(
+        tabs.length,
```
Sinucid: It will always have length, this check is useless

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,184 @@
...
+          <oryx-card>
+            <oryx-heading slot="heading" class="title">
+              ${this.productItem?.orderItem?.name}
+            </oryx-heading>
+            <div class="subtitle">${this.productItem?.orderItem.sku}</div>
+
+            <oryx-image
+              .src="${this.productItem?.product?.image}"
+              alt="${ifDefined(this.productItem?.orderItem?.name)}"
+              class="${this.status === ItemsFilters.NotFound
```
Sinucid: use `classMap` lit directive please

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -56,29 +145,117 @@ export class PickingComponent extends PickingListMixin(LitElement) {
...
+                `,
+                () => this.renderFallback()
+              )}
+            </div>`
+          )}`
+      )}
+    `;
   }
 
   protected renderFallback(): TemplateResult {
```
Sinucid, line 161:
```
protected renderFallback(): TemplateResult | void {
    if (!this.pickingList?.items.every(
      (item) => item.status === ItemsFilters.Picked
    )) {
      return html`
        <div class="picking-complete">
          <div class="img-wrap">
            <oryx-image resource="picking-items-processed"></oryx-image>
          </div>
          <h3 class="title-empty">${i18n(`picking.great-job`)}!</h3>
          <p>${i18n(`picking.all-items-are-processed`)}!</p>
        </div>`
    }
  }
```

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -56,29 +145,117 @@ export class PickingComponent extends PickingListMixin(LitElement) {
...
+        ?open=${this.isConfirmPickingDialogOpen}
+        enableFooter
+        preventclosebybackdrop
+      >
+        <div slot="heading">
+          ${i18n('picking.product-card.confirm-picking')}
+        </div>
+
+        <span>
+          You only picked
```
Sinucid: Add this ToDo please:
`//ToDo: should be refactored when i18n is supporting elements as a parts of templates`

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -56,29 +145,117 @@ export class PickingComponent extends PickingListMixin(LitElement) {
...
+        </div>
+      </oryx-modal>
+    `;
+  }
+
+  private updatePickingItem(
+    productIndex: number,
+    numberOfPicked?: number
+  ): void {
+    this.items[productIndex].numberOfPicked = numberOfPicked ?? 0;
```
Sinucid, line 225: Do items contain some unique identifier? Using index is not very safe

Sinucid (changes requested):

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,184 @@
...
+  EVENT_EDIT,
+  EVENT_SUBMIT,
+  ItemsFilters,
+  PickingListItem,
+  ProductItemPickedEvent,
+  SummaryInfo,
+} from '../../models';
+import { styles } from './picking-product-card.styles';
+
+export class PickingProductCardComponent extends LitElement {
```
tarashaliuk, line 17: Add `index.ts` which exports component and styles. The same suggestion for `picking`

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,184 @@
...
+  static styles = styles;
+
+  @property() productItem?: PickingListItem;
+  @property() status?: string;
+
+  @state() isCorrectNumberOfPickedProvided = true;
+  @state() currentNumberOfPicked?: number;
+  @state() pickedDataEvent?: ProductItemPickedEvent;
```
tarashaliuk: Looks that this state variable is not used anywhere

libs/domain/picking/src/components/picking/picking.component.ts
nsilva-spryker: We want the opposite if that, to know that every item from the picking list was picked already

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,184 @@
...
+  protected override render(): TemplateResult {
+    return html`${this.renderPickingProduct()}`;
+  }
+
+  protected renderPickingProduct(): TemplateResult {
+    if (this.productItem && this.currentNumberOfPicked) {
+      this.isCorrectNumberOfPickedProvided =
+        0 <= this.currentNumberOfPicked &&
+        this.currentNumberOfPicked <= this.productItem.quantity;
+    }
```
tarashaliuk: Can we move this logic to `onChangeQuantity` method?

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
nsilva-spryker: Optional chain expressions can return undefined by design - using a non-null assertion is unsafe and wrong

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,184 @@
...
+                  <div class="edit-quantity-info">
+                    ${i18n('picking.product-card.of')}
+                    ${this.productItem?.quantity}
+                    ${i18n('picking.product-card.items')}
+                  </div>
+
+                  <oryx-button>
+                    <button
+                      ?disabled="${!this.isCorrectNumberOfPickedProvided}"
+                    >
```
tarashaliuk: `type="submit"`

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,184 @@
...
+        this.currentNumberOfPicked <= this.productItem.quantity;
+    }
+
+    return html`
+      ${when(
+        this.productItem,
+        () => html`
+          <oryx-card>
+            <oryx-heading slot="heading" class="title">
+              ${this.productItem?.orderItem?.name}
```
tarashaliuk: let's use `if (!this.productItem) return` before returning the template to avoid conditional chains

libs/domain/picking/src/components/picking-product-card/picking-product-card.def.ts
```diff
@@ -0,0 +1,9 @@
+import { componentDef } from '@spryker-oryx/core';
+
+export const pickingProductCardComponent = componentDef({
+  name: 'oryx-picking-product-card',
+  impl: () =>
+    import('./picking-product-card.component').then(
+      (m) => m.PickingProductCardComponent
+    ),
+});
```
tarashaliuk: Add schema, please

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
tarashaliuk, line 17: Let's have stories for new components

libs/domain/picking/src/components/picking-product-card/picking-product-card.styles.ts
```diff
@@ -0,0 +1,70 @@
+import { css } from 'lit';
+
+export const styles = css`
+  :host {
+    text-align: center;
+    margin: 0 20px;
+  }
+
+  .title {
```
tarashaliuk: can we use `[slot='heading']` to avoid using extra classes?

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,184 @@
...
+            </oryx-heading>
+            <div class="subtitle">${this.productItem?.orderItem.sku}</div>
```
tarashaliuk: we can use `oryx-heading` here and avoid using extra styles

libs/domain/picking/src/components/picking-product-card/picking-product-card.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  .subtitle {
+    font-weight: 600;
+    font-size: 12px;
+    line-height: 16px;
+    letter-spacing: 0.01em;
+    color: var(--oryx-color-neutral-400);
+  }
+
+  .edit-quantity {
```
tarashaliuk:
```suggestion
  form {
```

libs/domain/picking/src/components/picking-product-card/picking-product-card.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+    margin-block-start: 7px;
+  }
+
+  .edit-quantity-info {
```
tarashaliuk:
```suggestion
  form div {
```

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -56,29 +145,117 @@ export class PickingComponent extends PickingListMixin(LitElement) {
...
+        enableFooter
+        preventclosebybackdrop
+      >
+        <div slot="heading">
+          ${i18n('picking.product-card.confirm-picking')}
+        </div>
+
+        <span>
+          You only picked
+          <span class="bold-text">
```
tarashaliuk:
```suggestion
          <span class="picked-items-info">
```

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -56,29 +145,117 @@ export class PickingComponent extends PickingListMixin(LitElement) {
...
+              ${i18n('picking.product-card.cancel')}
+            </button>
+          </oryx-button>
+
+          <oryx-button type="primary">
+            <button @click=${this.confirmPartialPicking}>
+              ${i18n('picking.product-card.confirm')}
+            </button>
+          </oryx-button>
+        </div>
```
tarashaliuk:
```suggestion
          <oryx-button slot="footer" outline type="secondary">
            <button @click=${this.onModalClose}>
              ${i18n('picking.product-card.cancel')}
            </button>
          </oryx-button>

          <oryx-button slot="footer" type="primary">
            <button @click=${this.confirmPartialPicking}>
              ${i18n('picking.product-card.confirm')}
            </button>
          </oryx-button>
```

libs/domain/picking/src/components/picking/picking.styles.ts
```diff
@@ -0,0 +1,42 @@
...
+  oryx-modal [slot='footer'] {
+    width: 100%;
+    display: flex;
+    gap: 10px;
+  }
+
+  oryx-modal oryx-button {
+    width: 100%;
+  }
+`;
```
tarashaliuk: let's use `footerButtonFullWidth` prop for `oryx-modal` when [Alvina's PR](https://github.com/spryker/oryx/pull/179) will be merged

tarashaliuk (changes requested):

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
Sinucid: Yes, but in render function you are rendering form with submit handling only when `this.productItem` is truthy, that means it can't dispatch submit till `this.productItem` is falsy.

libs/domain/picking/src/components/picking/picking.component.ts
Sinucid: Ok, just change the value of the logical expression to the opposite:
`if (this.pickingList?.items.every(`

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
Sinucid: button without type eq `type=submit`

Sinucid (changes requested):

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
nsilva-spryker: Let's use 'when' directive instead

libs/domain/picking/src/components/picking-product-card/picking-product-card.def.ts
nsilva-spryker: what's the benefit of using schema ?

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
nsilva-spryker: If stories is necessary for this case let's add a sub-task or an independent task fot that, let's not deal with that on this pr

libs/domain/picking/src/components/picking-product-card/picking-product-card.def.ts
tarashaliuk: To be able to use this component in EB

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
nsilva-spryker: unfortunately not, we need a different style on this one

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
nsilva-spryker: Let's change/adapt this if we need stories for that in another sub-task or independent task

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
nsilva-spryker: Your point is also right, but from my point of view, extracting this logic to another method and keep 'renderEditStatus' more readable and more focus on the html would be more clean.
Nevertheless I would adapt the code to your version

libs/domain/picking/src/components/picking/picking.component.ts
nsilva-spryker: let's just use the `${i18n('picking.product-card.you-only-picked')}` for now

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
nsilva-spryker: I think for this case using non-assertion could be, like this:
`productId: this.productItem!.product.id`

The productItem should never be null

libs/domain/picking/src/components/picking/picking.component.ts
nsilva-spryker: Since the items are not being push or pop, i don't think there are high risk on this. Also we are providing type annotations for the callback function, that ensures typeScript infers the correct type and avoid potential type or runtime errors

libs/domain/picking/src/components/picking/picking.styles.ts
aleespryker: It's merged, you can use it now.

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,213 @@
...
+    this.currentNumberOfPicked = quantity;
+
+    if (this.productItem && this.currentNumberOfPicked) {
+      this.isCorrectNumberOfPickedProvided =
+        0 <= this.currentNumberOfPicked &&
+        this.currentNumberOfPicked <= this.productItem.quantity;
+    }
+  }
+
+  protected getSummaryInfo(): SummaryInfo {
```
Sinucid: Left over I guess?

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
tarashaliuk: There is `h4` tag with the same styles.
You can find typography styles here:
image

tarashaliuk: Looks that you should update your branch with development

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
```diff
@@ -0,0 +1,213 @@
...
+                  <oryx-heading slot="heading"
+                    >${this.productItem?.orderItem.name}
+                  </oryx-heading>
+                  <div class="subtitle">${this.productItem?.orderItem.sku}</div>
+                `
+            )}
+
+            <oryx-image
+              .src="${this.productItem?.product?.image}"
+              alt="${ifDefined(this.productItem?.orderItem?.name)}"
```
tarashaliuk: `product` and `orderItem` are not optional

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
tarashaliuk: `when` creates one more nested level and requires conditional chains.

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -1,53 +1,134 @@
...
+  ProductItemPickedEvent,
+} from '../../models';
+import { styles } from './picking.styles';
 export class PickingComponent extends PickingListMixin(LitElement) {
-  protected tabs$ = this.pickingList$.pipe(
-    map((list) => [
+  static styles = styles;
+
+  @state()
+  protected isConfirmPickingDialogOpen = false;
```
Sinucid: Left over? As I suggested before you can just use `partialPicking` as a trigger for confirmation modal

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -1,53 +1,134 @@
...
+      };
+
+      this.isConfirmPickingDialogOpen = true;
+      return;
+    }
+
+    this.updatePickingItem(productIndex, detail.numberOfPicked);
+  }
+
+  protected editPickingItem(event: Event): void {
```
Sinucid:
```
  protected savePickingItem(event: CustomEvent<ProductItemPickedEvent>): void {
    const { productId } = event.detail;
```

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -1,53 +1,134 @@
...
+            item.status === ItemsFilters.Picked &&
+            (item.numberOfNotPicked || item.numberOfPicked < item.quantity)
         ),
       },
-    ])
-  );
+    ];
+  }
+
+  protected savePickingItem(event: Event): void {
```
Sinucid:
```
  protected savePickingItem(event: CustomEvent<ProductItemPickedEvent>): void {
    const { productId, numberOfPicked } = event.detail;
```

Sinucid (changes requested):

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
nsilva-spryker: yup, It was updated with this changes from 7 days ago, I will use this instead

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
nsilva-spryker: Sorry not getting this one, the if is also a conditional, using 'when' I would say it's more lit native since this is essentially a wrapper around a ternary expression. 
We don't see many if statements inside our html templates

libs/domain/picking/src/components/picking/picking.component.ts
Sinucid: I don't see this changes

libs/domain/picking/src/components/picking/picking.component.ts
Sinucid: Either

Sinucid (dismissed):

libs/domain/picking/src/components/picking-product-card/picking-product-card.component.ts
tarashaliuk: You can use `if` statement with return because you haven’t `else` state. Benefit of using is better readability of code because you shouldn’t scroll down to see if it is `else` statement and there are less nesting lines. 
Also when you will use `if` statement then you can avoid conditional chains (`this.productItem?.`)

tarashaliuk (approved):

PR #181, opened 2023-03-21
fix: move locale into i18n package

gund:
- Add refresh token to oauth feature
- Reset auth state if token is invalid and refresh failed
- Refactor locale stuff from site package into i18n package
- Refactor LocaleService to use adapter which does not depend on StoreService and introduce SapiLocaleAdapter which uses StoreService to resolve locales
- Fix small issue with default i18n interpolation

BREAKING CHANGES: `LocaleService` and `LocaleChanged` were moved from `site` package into `i18n` package.

Fixes [HRZ-2416](https://spryker.atlassian.net/browse/HRZ-2416)

[HRZ-2416]: https://spryker.atlassian.net/browse/HRZ-2416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/site/src/services/locale/accept-language.interceptor.ts
```diff
@@ -36,24 +36,22 @@ export class AcceptLanguageInterceptor implements HttpInterceptor {
...
     options: RequestOptions
   ): RequestOptions {
-    return {
-      ...options,
-      headers: {
-        ...options.headers,
-        [this.headerName]: locale,
-      } as HeadersInit,
-    };
+    const headers = new Headers(options.headers);
```
supproduction, line 39: By the way. We used strings approach because of node-fetch issues. Now we can use the same approach as in angular that you create new Request. Could you add TODO please?

libs/platform/i18n/src/lib/feature.ts
```diff
@@ -15,6 +14,14 @@ export class I18nFeature implements AppFeature {
     this.providers = [
       ...i18nProviders,
       options && { provide: DefaultI18nLoaderConfig, useValue: options },
+      options?.locale && {
+        provide: DefaultLocaleAdapterConfig,
+        useValue: options?.locale,
```
supproduction:
```suggestion
        useValue: options.locale,
```

libs/platform/i18n/src/lib/locale/adapter/default-locale.adapter.ts
```diff
@@ -0,0 +1,39 @@
...
+  readonly defaultLocale?: string;
+}
+
+export const DefaultLocaleAdapterConfig = 'oryx.DefaultLocaleAdapterConfig';
+
+declare global {
+  interface InjectionTokensContractMap {
+    [DefaultLocaleAdapterConfig]: DefaultLocaleAdapterConfig;
+  }
+}
```
supproduction, line 39: Can we move it to the `locale.adapter`?

libs/domain/site/src/services/locale/accept-language.interceptor.ts
gund: Discussed this and it's not related to this PR.

libs/platform/i18n/src/lib/locale/adapter/default-locale.adapter.ts
gund: Discussed this and decided to keep it together with the default adapter as it's only related to that one.

supproduction (dismissed):

libs/platform/i18n/src/lib/locale/adapter/default-locale.adapter.ts
```diff
@@ -0,0 +1,39 @@
...
+  getDefault(): Observable<string> {
+    return of(
+      this.config?.defaultLocale ??
+        this.config?.locales?.[0].code ??
+        this.getFallbackLang()
+    );
+  }
+
+  protected getFallbackLang(): string {
+    return navigator.language;
```
dunqan: won't work on SSR?

libs/platform/i18n/src/lib/locale/default-locale.service.ts
```diff
@@ -0,0 +1,74 @@
...
+      new DefaultLocaleAdapter(),
+    protected queryService = inject(QueryService)
+  ) {
+    this.active$ = concat(
+      this.adapter.getDefault().pipe(takeUntil(this.setActive$)),
+      this.setActive$
+    ).pipe(
+      distinctUntilChanged(),
+      tap((locale) =>
+        this.queryService.emit({ type: LocaleChanged, data: locale })
```
dunqan: previously this was emitting not on frist locale. but after a change

dunqan (changes requested):

libs/platform/i18n/src/lib/locale/adapter/default-locale.adapter.ts
gund: No it won't, but if you don't want it to get called - you should manually provide "default" locale instead.

libs/platform/i18n/src/lib/locale/default-locale.service.ts
gund: Yep, I thought it was a bug actually because previously there was no way of knowing the initial locale via this event.
Shouldn't it be fixed like this?

libs/platform/i18n/src/lib/locale/adapter/default-locale.adapter.ts
gund: We will hardcode "en" as a fallback.

libs/domain/site/src/services/store/store.interceptor.ts
```diff
@@ -41,13 +41,11 @@ export class StoreInterceptor implements HttpInterceptor {
...
-    return {
-      ...options,
-      headers: {
-        ...options.headers,
-        'X-Store': store,
-      } as HeadersInit,
-    };
+    const headers = new Headers(options.headers);
+
+    headers.set('X-Store', store);
```
tobi-or-not-tobi, line 46: Can we change this to `Store` , according to AA on multi-store, this should already be the new standard. 

Also, we might need to make this optional, as it's only a thing in case of multi-store. Something for a follow up task I'd say.

tobi-or-not-tobi: The whole store interceptor could maybe be renamed to `MultiStoreInterceptor`

libs/platform/i18n/src/lib/locale/adapter/default-locale.adapter.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+}
+
+export interface DefaultLocaleAdapterConfig {
+  readonly locales?: Locale[];
+  readonly defaultLocale?: string;
+}
+
+export const DefaultLocaleAdapterConfig = 'oryx.DefaultLocaleAdapterConfig';
```
tobi-or-not-tobi, line 33: Should it not be `LocaleAdapterConfig`?

libs/template/presets/src/features/fulfillment/app.ts
```diff
@@ -11,32 +12,34 @@ import {
...
+}
+
 export const fulfillmentTheme = { ...theme };
+
+export const fulfillmentResources: Resources = {
+  graphics: { ...resourceGraphics },
+};
+
+export const defaultFulfillmentI18nConfig: I18nFeatureOptions = {
+  locale: { defaultLocale: 'en' },
```
tobi-or-not-tobi: what about moving `en` as the default in to the `LocaleService`?

tobi-or-not-tobi (changes requested):

libs/platform/i18n/src/lib/locale/default-locale.service.ts
gund: Will fix to skip first emition.

libs/domain/site/src/services/store/store.interceptor.ts
gund: This task is not changing how store interceptor works - I only updated the way we work with headers as previously it was very fragile.
I would suggest creating another ticket for the interceptor changes.

libs/platform/i18n/src/lib/locale/adapter/default-locale.adapter.ts
gund: We discussed with Stan and will bring this topic to catchup as there are some things to consider.
For now we will leave it as is.

libs/template/presets/src/features/fulfillment/app.ts
gund: Actually now that we've made 'en' a fallback locale I will remove this config altogether.

tobi-or-not-tobi (dismissed):

tobi-or-not-tobi (approved):

dunqan (approved):

PR #184, opened 2023-03-23
chore: release labs workflow

tolerants:
Introduce release labs workflow

closes: https://spryker.atlassian.net/browse/HRZ-2452


.github/workflows/release-labs.yml
```diff
@@ -0,0 +1,49 @@
+name: Release labs Workflow
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run && github.event.workflow_run.conclusion == 'success' || true }}
```
gund: We don't need this check as this workflow does not depend on any other.

gund (commented):
Looks good, just a small comment.

.github/workflows/release-labs.yml
tolerants: Good catch, fixed.

tobi-or-not-tobi (approved):

PR #187, opened 2023-03-23
chore: new workflow for release testing

tolerants:
- Added new workflow for release testing
- Added separate application based on storefront for testing
- Removed labs feature from storefront app
- Fixed e2e tests

closes https://spryker.atlassian.net/browse/HRZ-2538, https://spryker.atlassian.net/browse/HRZ-2593


.github/workflows/release-test.yml
```diff
@@ -3,154 +3,80 @@ name: Release test Workflow
...
+          restore-keys: |
+            npm-node-
+      - name: Auth Git
+        run: |
+          git config user.email "${{ env.GIT_EMAIL }}"
+          git config user.name "${{ env.GIT_USER }}"
+      - name: Install Deps
+        run: npm ci
+      - name: Build packages
+        run: npx nx run-many --target=build --all --parallel=2 --exclude testfront, storybook, fulfillment, storefront
```
AndreyYevtukhov, line 42: The list of excluded packages is different here - https://github.com/spryker/oryx/pull/251/files#diff-fab5d050d343dd5b95a7a1e24ae6bea6c4dfbd5a714e432d0588caa6fbbfae46R20 (should we adjust these after merges?)

AndreyYevtukhov: Maybe create a small follow up ticket to not forget about it)

.github/workflows/release-test.yml
```diff
@@ -3,154 +3,80 @@ name: Release test Workflow
...
+      - name: Release minor
+        run: npm run release-minor
+      - name: Install dependencies for testfront
+        run: cd apps/testfront && npm i && npm list --depth=0 && cd ../..
+      - name: Run testfront
+        run: npm run testfront
         env:
-          SCOS_BASE_URL: ${{ secrets.SCOS_BASE_URL }}
-          CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+          FES_CONTENT_BACKEND_URL: https://test-experience-builder.netlify.app/.netlify/functions/server
```
AndreyYevtukhov: hardcoded URLs - should we remove them to ENV vars?

apps/testfront/.env.template
```diff
@@ -0,0 +1 @@
+SCOS_BASE_URL=https://glue.de.faas-suite-prod.cloud.spryker.toys
```
AndreyYevtukhov: URL that probably should not be public

apps/testfront/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
AndreyYevtukhov, line 1: Comment regarding all `testfront` files

---

What about creating a small script (node/bash) to check that `storefront` and `testfront` are syncronized in common files? 

Because it might happen that they are a bit different (e.g. someone changed the file in SF but forget/didn't know about TF) and it will cause some issues (e2e tests might work, but UI can be different, etc.)

Or another option is to just copy common files from `SF` to `TF` before each build (with some kind of plugin or something?)

AndreyYevtukhov (commented):
Left small comments

.github/workflows/release-test.yml
tolerants: Yep, I'll adjust.

apps/testfront/package.json
tolerants: Yep, I'll create a follow up ticket for it.

tolerants: https://spryker.atlassian.net/browse/HRZ-2693

.github/workflows/release-test.yml
tolerants: Thanks, fixed.

AndreyYevtukhov (approved):

PR #194, opened 2023-03-24
feat: add app initializer

supproduction:
adds useExisting provider
adds app initializer service
adds initializers token

closes [HRZ-2541](https://spryker.atlassian.net/browse/HRZ-2541)

[HRZ-2541]: https://spryker.atlassian.net/browse/HRZ-2541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/core/src/services/app-initializer/app-initializer.ts
+++ b/libs/platform/core/src/services/app-initializer/app-initializer.ts
@@ -1,15 +1,14 @@
-import { OnDestroy } from '@spryker-oryx/di';
 import { Observable } from 'rxjs';
 
 export const AppInitializerService = 'oryx.AppInitializerService';
 export const AppInitializer = 'oryx.AppInitializer*';
 
-export interface AppInitializerService extends OnDestroy {
-  initialize(): void;
+export interface AppInitializerService {
+  initialize(): Promise<void>;
 }
 
 export interface AppInitializer {
-  initialize(): void | Observable<unknown>;
+  initialize(): void | Observable<void> | Promise<void>;
 }
 
 declare global {
--- a/libs/platform/core/src/services/app-initializer/app-initializer.ts
+++ b/libs/platform/core/src/services/app-initializer/app-initializer.ts
@@ -1,0 +1,20 @@
+import { OnDestroy } from '@spryker-oryx/di';
+import { Observable } from 'rxjs';
+
+export const AppInitializerService = 'oryx.AppInitializerService';
+export const AppInitializer = 'oryx.AppInitializer*';
+
+export interface AppInitializerService extends OnDestroy {
+  initialize(): void;
+}
+
+export interface AppInitializer {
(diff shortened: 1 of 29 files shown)


libs/platform/core/src/services/app-initializer/app-initializer.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const AppInitializerService = 'oryx.AppInitializerService';
+export const AppInitializer = 'oryx.AppInitializer*';
+
+export interface AppInitializerService extends OnDestroy {
+  initialize(): void;
+}
+
+export interface AppInitializer {
+  initialize(): void | Observable<unknown>;
```
gund: Return type should be consistent, also maybe we can allow for sync methods as well:
```suggestion
  initialize(): void | Promise<void> | Observable<void>;
```

libs/platform/core/src/services/app-initializer/app-initializer.ts
```diff
@@ -0,0 +1,20 @@
+import { OnDestroy } from '@spryker-oryx/di';
+import { Observable } from 'rxjs';
+
+export const AppInitializerService = 'oryx.AppInitializerService';
+export const AppInitializer = 'oryx.AppInitializer*';
+
+export interface AppInitializerService extends OnDestroy {
```
gund: Why do you need to enforce that `AppInitializerService` should also implement `OnDestroy`?
Shouldn't this be implementation details?

gund (commented):
Looks good just a few small suggestions.

gund (approved):

libs/platform/core/src/services/app-initializer/app-initializer.ts
```diff
@@ -0,0 +1,19 @@
+import { Observable } from 'rxjs';
+
+export const AppInitializerService = 'oryx.AppInitializerService';
+export const AppInitializer = 'oryx.AppInitializer*';
+
+export interface AppInitializerService {
+  initialize(): Promise<void>;
+}
+
+export interface AppInitializer {
```
dunqan, line 10: would be nice to be able to use just promise or observable, not full blown class

libs/platform/core/src/services/app-initializer/default-app-initializer.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  async initialize(): Promise<void> {
+    const initializers = this.injector.inject(AppInitializer, []);
+
+    for (const initializer of initializers) {
+      const result = initializer.initialize();
+
+      if (isPromise(result)) {
+        await result;
+      }
```
dunqan, line 17: shouldn't we use `else` here?

PR #196, opened 2023-03-24
[WIP] chore: split cart load

dunqan:


dunqan:
not doing it yet, because of;
- it's not fully improving performance (getting carts without entries sill has slow response time)
- we would like to make it configurable, to optimize single cart usages

PR #199, opened 2023-03-27
fix: auto close notification

tobi-or-not-tobi:
The notifications in the notification center no longer closed after the auto close time passed. This has been fixed as well as some refactors to improve the code and UI:
- improved animations – a stacked notification will first appear above the stack list before it will animate down into the stack list
- improve stories
- simplify registry (small controller that controls the list of active notifications)

fixes [HRZ-2578](https://spryker.atlassian.net/browse/HRZ-2578)

[HRZ-2578]: https://spryker.atlassian.net/browse/HRZ-2578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tolerants:
Is it really a `chore` or it's a `fix` actually? We are changing ui/ux here as I can see.

libs/base/ui/overlays/notification-center/src/registry.controller.ts
```diff
@@ -1,142 +1,110 @@
...
+    setTimeout(() => {
+      registryItem.visible = true;
+      this.host.requestUpdate();
+    }, 0);
   }
 
-  handleNotificationClose(key: string): void {
-    this.removeFromRegistry(key, true);
+  protected scheduleAutoClosing(item: NotificationRegistry): void {
+    // const key = item.key as string;
```
tolerants: Leftover?

libs/base/ui/overlays/notification-center/src/registry.controller.ts
```diff
@@ -1,142 +1,110 @@
...
 
-  protected notificationIsVisible(targetKey: string): boolean {
-    return Boolean(
-      (
-        this.registry.find(
-          ({ key }) => key === targetKey
-        ) as NotificationRegistry
-      )?.visible
+  hostConnected?(): void {
+    this.host.addEventListener('oryx.close', (ev: any) =>
```
tolerants: Why `ev`? Suggest to use `event` or `e` which is kind most popular abbreviation for event.

libs/base/ui/overlays/notification-center/src/registry.controller.ts
```diff
@@ -1,142 +1,110 @@
...
-        return item;
-      }
-    );
+import { AlertType } from '@spryker-oryx/ui';
+import { Notification } from '../../notification';
 
-    this._registry = newRegistryWithParams;
-    this.host.requestUpdate();
-  }
+const DESTROY_DELAY_TIME = 600;
```
Sinucid: maybe this should be a property with default value?

libs/base/ui/overlays/notification-center/src/registry.controller.ts
Sinucid: left over?

libs/base/ui/overlays/notification-center/src/registry.controller.ts
```diff
@@ -1,142 +1,110 @@
...
+      strategy.autoClose =
+        strategy.type === AlertType.Info || AlertType.Success === strategy.type;
     }
-  }
 
-  protected removeFromRegistry(targetKey: string, force = false): void {
-    if (force) {
-      //should clear the delayed callback
-      this.preventAutoClose(targetKey);
+    if (!strategy.closable && !strategy.autoClose) {
```
Sinucid: Why do we need this resolving? If I'm not mistaken our notifications are supporting custom content where can be custom close button. But this logic prevents having non autoclosable notification without general close button.

I believe that will be enough just this:
`const strategy = { floating: true, closable: true, autoClose: item.type === AlertType.Info || item.type === AlertType.Success, ...item };`

libs/base/ui/overlays/notification-center/src/registry.controller.ts
```diff
@@ -1,142 +1,110 @@
...
-      }, 0);
-    }
+  protected onClose(key: number): void {
+    this.destroy(key);
   }
 
-  constructor(public host: LitElement) {
-    this.host.addController(this);
+  protected onMouseEnter(): void {
+    Object.keys(this.autoCloseQueue).forEach((key) => {
```
Sinucid, line 78: Should it prevent auto close of all notifications or thats one that was mousevered?

libs/domain/site/notification-center/src/notification-center.component.ts
```diff
@@ -1,29 +1,26 @@
...
 import { SiteNotificationCenterOptions } from './notification-center.model';
 
 @defaultOptions({
-  marginBlock: '40px',
-  marginInline: '30px',
-  maxWidth: '470px',
-  enableStacking: true,
   position: NotificationPosition.BottomStart,
+  enableStacking: true,
+  defaultAutoCloseTime: 8,
```
Sinucid, line 17: I believe that milliseconds is a standard

libs/domain/site/notification-center/src/notification-center.component.ts
```diff
@@ -34,31 +31,24 @@ export class SiteNotificationCenterComponent extends ContentMixin<SiteNotificati
       if (!(this.centerRef.value && 'open' in this.centerRef.value)) {
         await customElements.whenDefined('oryx-notification-center');
       }
-      this.centerRef.value?.open?.(notification);
+      if (
```
Sinucid: `notification.autoCloseTime ??= this.componentOptions.defaultAutoCloseTime`

Sinucid (changes requested):

libs/base/ui/overlays/notification-center/src/registry.controller.ts
tobi-or-not-tobi: What about changing it to 1s to allow for enough time. I'd rather keep the implementation simple here, more time doesn't harm.

libs/base/ui/overlays/notification-center/src/registry.controller.ts
tobi-or-not-tobi: agreed, doesn't need the complexity

libs/base/ui/overlays/notification-center/src/registry.controller.ts
tobi-or-not-tobi: all

libs/domain/site/notification-center/src/notification-center.component.ts
tobi-or-not-tobi: And seconds are easier for business users.

tolerants (dismissed):

libs/domain/site/notification-center/src/notification-center.component.ts
```diff
@@ -31,13 +32,9 @@ export class SiteNotificationCenterComponent extends ContentMixin<SiteNotificati
...
       }
-      if (
-        !notification.autoCloseTime &&
-        this.componentOptions?.defaultAutoCloseTime
-      ) {
-        notification.autoCloseTime =
-          this.componentOptions.defaultAutoCloseTime * 1000;
-      }
+      notification.autoCloseTime ??=
+        (this.componentOptions.defaultAutoCloseTime ??
```
Sinucid: Why do we need to use `options.defaultAutoCloseTime` as fallback value? If user doesn't provide defaultAutoCloseTime in custom config - it will use `defaultAutoCloseTime` from default options by default.

Sinucid (changes requested):

Sinucid (dismissed):

Sinucid (dismissed):

Sinucid (approved):

tolerants (approved):

PR #206, opened 2023-03-28
feat: replace hardcoded time formatting method

tarashaliuk:
closes: [HRZ-2520](https://spryker.atlassian.net/browse/HRZ-2520)

[HRZ-2520]: https://spryker.atlassian.net/browse/HRZ-2520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/i18n/src/lib/locale/default-locale.service.ts
```diff
@@ -57,21 +57,34 @@ export class DefaultLocaleService implements LocaleService {
     this.setActive$.next(value);
   }
 
-  formatDate(stamp: string | number, showTime = false): Observable<string> {
+  formatDate(
+    stamp: string | number | Date,
+    showTime = false,
+    showDate = false
+  ): Observable<string> {
```
gund: Can we split this method into multiple once to remove those boolean flags as it's really unclear what that means in other places when it's being used.
We can have at least 2 separate methods `formatDate` and `formatTime`.
If needed you may create the third one `formatDateTime` which will render both date and time but if it's not used anywhere no need for it.

gund (changes requested):

libs/platform/i18n/src/lib/locale/default-locale.service.ts
```diff
@@ -57,34 +57,35 @@ export class DefaultLocaleService implements LocaleService {
...
-    let template = {};
-
-    if (showDate) {
-      template = {
-        day: 'numeric',
-        month: 'short',
-        year: 'numeric',
-      };
-    }
+  formatDate(stamp: string | number): Observable<string> {
```
gund:
```suggestion
  formatDate(stamp: string | number | Date): Observable<string> {
```

libs/platform/i18n/src/lib/locale/default-locale.service.ts
```diff
@@ -57,34 +57,35 @@ export class DefaultLocaleService implements LocaleService {
...
+    return this.get().pipe(
+      map((locale) =>
+        Intl.DateTimeFormat(locale.replace('_', '-')).format(new Date(stamp))
```
gund: I think now it's possible to cache each of the Intl instance for specific method and reuse it between the calls to optimize formatting performance.
You can create 3 observables which will compute it's own specific Intl instance with shareReplay() operators so that it's not recomputed for every call.

gund (changes requested):

libs/platform/i18n/src/lib/locale/default-locale.service.ts
```diff
@@ -18,6 +18,9 @@ export class DefaultLocaleService implements LocaleService {
   protected setActive$ = new ReplaySubject<string>(1);
   protected active$;
   protected all$;
+  private dateFormat$: Observable<Intl.DateTimeFormat>;
+  private timeFormat$: Observable<Intl.DateTimeFormat>;
+  private dateTimeFormat$: Observable<Intl.DateTimeFormat>;
```
gund: We always use `protected` access modifier for extendibility purpose:
```suggestion
  protected dateFormat$: Observable<Intl.DateTimeFormat>;
  protected timeFormat$: Observable<Intl.DateTimeFormat>;
  protected dateTimeFormat$: Observable<Intl.DateTimeFormat>;
```

libs/platform/i18n/src/lib/locale/default-locale.service.ts
```diff
@@ -57,35 +88,24 @@ export class DefaultLocaleService implements LocaleService {
...
   }
 
-  formatDate(stamp: string | number): Observable<string> {
-    return this.get().pipe(
-      map((locale) =>
-        Intl.DateTimeFormat(locale.replace('_', '-')).format(new Date(stamp))
-      )
+  formatDate(stamp: string | number | Date): Observable<string> {
+    return this.dateFormat$.pipe(
+      map((dateFormat) => dateFormat.format(new Date(stamp)))
```
gund: Seems like missing check to avoid extra date conversion:
```suggestion
      map((dateFormat) => stamp instanceof Date ? stamp : new Date(stamp))
```

gund (changes requested):

gund (dismissed):

gund (approved):

PR #208, opened 2023-03-29
chore: locale and currency selector unit test coverage

tobi-or-not-tobi:
Unit tests added for context selectors. 

closes [HRZ-2490](https://spryker.atlassian.net/browse/HRZ-2490), [HRZ-2494](https://spryker.atlassian.net/browse/HRZ-2494)

[HRZ-2490]: https://spryker.atlassian.net/browse/HRZ-2490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-2494]: https://spryker.atlassian.net/browse/HRZ-2494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/site/currency-selector/src/currency-selector.component.spec.ts
+++ b/libs/domain/site/currency-selector/src/currency-selector.component.spec.ts
@@ -1,0 +1,193 @@
+import { fixture } from '@open-wc/testing-helpers';
+import { useComponent } from '@spryker-oryx/core/utilities';
+import { createInjector, destroyInjector } from '@spryker-oryx/di';
+import { LocaleService } from '@spryker-oryx/i18n';
+import {
+  Currency,
+  CurrencyService,
+  siteCurrencySelectorComponent,
+} from '@spryker-oryx/site';
+import { html } from 'lit';
+import { of } from 'rxjs';
+import { SiteCurrencySelectorComponent } from './currency-selector.component';
+
+class MockCurrencyService implements Partial<CurrencyService> {
+  get = vi.fn().mockReturnValue('EUR');
+  getAll = vi.fn().mockReturnValue([]);
+  set = vi.fn();
+}
+
+class MockLocaleService implements Partial<LocaleService> {
+  get = vi.fn().mockReturnValue(of('en'));
+}
+
+describe('SiteCurrencySelectorComponent', () => {
+  let element: SiteCurrencySelectorComponent;
+  let currencyService: MockCurrencyService;
+  let localeService: MockLocaleService;
+
+  beforeAll(async () => {
+    await useComponent([siteCurrencySelectorComponent]);
+  });
+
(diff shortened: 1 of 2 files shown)


supproduction (dismissed):

libs/domain/site/currency-selector/src/currency-selector.component.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+      });
+
+      it('should render the GBP option', () => {
+        const eur = element.shadowRoot?.querySelector('oryx-option[value=GBP]');
+        expect(eur).toBeDefined();
+        expect(eur?.textContent).toContain('libra esterlina');
+      });
+    });
+
+    describe('and the language is German', () => {
```
tarashaliuk: what is the benefit of testing two languages? Looks like it is enough to test one language

libs/domain/site/locale-selector/src/locale-selector.component.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+        html`<oryx-site-locale-selector></oryx-site-locale-selector>`
+      );
+    });
+
+    it('should not show the locale selector', () => {
+      expect(element).not.toContainElement('button');
+    });
+  });
+
+  describe('when there multiple only one locales', () => {
```
tarashaliuk:
```suggestion
  describe('when there is only one locale', () => {
```

libs/domain/site/locale-selector/src/locale-selector.component.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+        html`<oryx-site-locale-selector></oryx-site-locale-selector>`
+      );
+    });
+
+    it('should not show the locale selector', () => {
+      expect(element).not.toContainElement('button');
+    });
+  });
+
+  describe('when there are more only one locales', () => {
```
tarashaliuk:
```suggestion
  describe('when there are more than one locale', () => {
```

tarashaliuk (changes requested):

libs/domain/site/currency-selector/src/currency-selector.component.spec.ts
tobi-or-not-tobi: Just to ensure we do react to the current locale, rather than falling back to a language. But, that being said, agreed, we can only test Spanish, as this is not likely to be our default. Will change it.

tarashaliuk (approved):

PR #209, opened 2023-03-29
feat: translate local in its own language

tobi-or-not-tobi:
Each locale option in the locale selector is now rendered in it’s own locale, so that the user who ended up in the wrong local is able to switch. Previously, labels where rendered in the active locale, which causes difficulties for users to switch.

closes [HRZ-2581](https://spryker.atlassian.net/browse/HRZ-2581)

[HRZ-2581]: https://spryker.atlassian.net/browse/HRZ-2581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/site/locale-selector/src/locale-selector.component.spec.ts
+++ b/libs/domain/site/locale-selector/src/locale-selector.component.spec.ts
@@ -102,10 +102,19 @@ describe('SiteLocaleSelectorComponent', () => {
       expect(element).toContainElement('button');
     });
 
-    it('should render the options', () => {
-      expect(element).toContainElement('oryx-option[value=en');
-      expect(element).toContainElement('oryx-option[value=de');
-      expect(element).toContainElement('oryx-option[value=es');
+    it('should render the english option in its native locale', () => {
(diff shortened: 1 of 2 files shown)


libs/domain/site/locale-selector/src/locale-selector.component.spec.ts
```diff
@@ -102,10 +102,19 @@ describe('SiteLocaleSelectorComponent', () => {
       expect(element).toContainElement('button');
     });
 
-    it('should render the options', () => {
-      expect(element).toContainElement('oryx-option[value=en');
-      expect(element).toContainElement('oryx-option[value=de');
-      expect(element).toContainElement('oryx-option[value=es');
+    it('should render the english option in its native locale', () => {
+      const en = element.shadowRoot?.querySelector('oryx-option[value=en');
```
Sinucid: Closes the square brackets in all selectors, please, to make them consistent

Sinucid (changes requested):

Sinucid (approved):

PR #213, opened 2023-03-30
feat: implement fulfillment login page

aleespryker:
Closes [HRZ-2548](https://spryker.atlassian.net/browse/HRZ-2548), [HRZ-2549](https://spryker.atlassian.net/browse/HRZ-2549)

[HRZ-2548]: https://spryker.atlassian.net/browse/HRZ-2548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-2549]: https://spryker.atlassian.net/browse/HRZ-2549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/offline/data-plugin.ts
```diff
@@ -0,0 +1,107 @@
...
+    return onlineAdapter.get({}).pipe(
+      map((pl) => {
+        const productIds = new Set<string>();
+
+        return {
+          pickingLists: pl,
+          products: pl
+            .map((pickingList) => pickingList.items.map((item) => item.product))
+            .flat()
+            .filter((product) => {
```
Sinucid, line 79:
```
.filter((product, i, arr) => {
  return arr.findIndex(p => p.id === product.id) === i;
}),
```

libs/platform/auth/src/components/login/login.component.ts
```diff
@@ -55,7 +55,8 @@ export class AuthLoginComponent extends ContentMixin<LoginOptions>(LitElement) {
     }),
     switchMap((request) =>
       this.authLoginStrategy.login(request).pipe(
-        catchError(() => {
+        catchError((e) => {
+          console.error(e);
```
Sinucid, line 59: Do we need it?

libs/template/application/src/auth/bapi/feature.ts
```diff
@@ -34,10 +35,10 @@ export class BapiAuthFeature extends OauthFeature implements AppFeature {
           clientId: 'frontend',
           grantType: 'authorization_code',
           authUrl: new URL('/login', globalThis.location.origin).toString(),
-          tokenUrl: new URL(
-            '/token',
-            injectEnv('ORYX_FULFILLMENT_BACKEND_URL')
-          ).toString(),
+          tokenUrl: urlJoin(
```
Sinucid, line 44: What is the benefit of using third-party package here?

Sinucid (changes requested):

libs/platform/auth/src/components/login/login.component.ts
aleespryker: @gund Recommended it to have more useful error messages in case the login fails.

libs/template/application/src/auth/bapi/feature.ts
aleespryker: Because it doesn't get the right url. If you have a `ORYX_FULFILLMENT_BACKEND_URL` like this: `https://fulfillment-app.netlify.app/functions/server`, you end up with `https://fulfillment-app.netlify.app/token` using `new URL` instead of `https://fulfillment-app.neltify.app/functions/server/token`, which is what we want.

libs/platform/auth/src/components/login/login.component.ts
Sinucid: Is standard interception by notification center not enough? I'm not against this, just asking

libs/domain/picking/offline/data-plugin.ts
aleespryker: This doesn't work. And it's missing `productIds.add(product.id)`.

libs/platform/auth/src/components/login/login.component.ts
aleespryker: Had a closer look. We don't currently have notification center in fulfillment app. Also, we catch the error, so even if notification center was set up, it will only display uncaught errors. If we really wanted to tie it to a notification center, then we might use the ErrorService directly for that.

libs/domain/picking/offline/data-plugin.ts
Sinucid: Why? This is doing the same thing, filter duplications, but in one string, without additional Set's and etc. `productIds` is not needed anymore

libs/domain/picking/offline/data-plugin.ts
```diff
@@ -0,0 +1,104 @@
...
+    );
+  }
+
+  protected populateDb(injector: Injector): Observable<void> {
+    const onlineAdapter = injector.inject(PickingListOnlineAdapter);
+    const dexieIdbService = injector.inject(DexieIndexedDbService);
+
+    return onlineAdapter.get({}).pipe(
+      map((pl) => {
+        const productIds = new Set<string>();
```
Sinucid, line 73: Leftover

Sinucid (changes requested):

libs/domain/picking/offline/data-plugin.ts
gund: This is going to be a much slower performance wise as you are doing loop in a loop to find duplicates while Set is O(1) which is much faster.
So I would suggest to revert to a previous implementation assuming that we are downloading the whole set of picking lists in one request.

libs/template/presets/src/features/fulfillment/components/fulfillment-root/fulfillment-root.component.ts
```diff
@@ -11,6 +16,35 @@ export class FulfillmentRootComponent extends LitElement {
...
+        this.authService.login();
+      }
+      if (
+        authenticated &&
+        (currentRoute.startsWith('/login') || currentRoute.startsWith('/oauth'))
+      ) {
+        this.routerService.navigate('/');
+      }
+    })
+  );
```
gund: I think it's better to move this to a separate plugin in the `application` package instead of embedding it into a component.

libs/template/presets/src/features/fulfillment/components/fulfillment-root/fulfillment-root.component.ts
```diff
@@ -11,6 +16,35 @@ export class FulfillmentRootComponent extends LitElement {
...
+  @subscribe()
+  protected guard$ = combineLatest([
+    this.authService.isAuthenticated(),
+    this.routerService.currentRoute(),
+  ]).pipe(
+    tap(([authenticated, currentRoute]) => {
+      if (
+        !authenticated &&
+        currentRoute !== '/login' &&
+        !currentRoute.startsWith('/oauth')
```
gund: Can you inject oauth config to get login url and oauth handler url instead of the hardcoded ones?

libs/platform/auth/src/oauth/providers/code-grant-login.strategy.ts
```diff
@@ -16,12 +16,15 @@ export class CodeGrantAuthLoginStrategy implements AuthLoginStrategy {
...
+      .request<CodeGrantAuthLoginResponse[]>(this.config.loginUrl, {
         method: this.config.loginMethod ?? 'POST',
         credentials: 'omit',
         body,
+        headers: {
+          'Content-Type': 'application/x-www-form-urlencoded',
+        },
       })
-      .pipe(switchMap((response) => this.redirectWith(response)));
+      .pipe(switchMap((response) => this.redirectWith(response[0])));
```
gund, line 27: Do you need to update mock server to make it's response match with this change?

gund (changes requested):

libs/platform/auth/src/oauth/providers/code-grant-login.strategy.ts
aleespryker: We needed it for the token but not authorize. If we want authorize to have the same format so that it's not wrapped in an array, then I will update the mock server for it as well.

gund: Ok let's keep it, seems like BAPI already using this format anyways.

libs/domain/picking/offline/data-plugin.ts
Sinucid: So, do you think that just reading data already stored in memory is less performance then writing new data to the memory during updating the Set and searching in it? I think the difference is not significant.

gund: Absolutely, you will only write once for every item with a Set, otherwise you will read every time the same array for every item.
The difference is literally exponential.

libs/template/application/src/auth/bapi/plugin.ts
```diff
@@ -0,0 +1,47 @@
...
+    ])
+      .pipe(
+        tap(([authenticated, currentRoute]) => {
+          const oauthConfig = resolve(OauthServiceConfig);
+          const redirectUrl = new URL(
+            oauthConfig.providers[0].redirectUrl as string
+          );
+          if (
+            !authenticated &&
+            currentRoute !== '/login' &&
```
gund: What about login route, could you inject this one as well from the config?

package.json
```diff
@@ -159,6 +159,7 @@
     "rxjs": "7.5.6",
     "serverless-http": "^3.2.0",
     "tslib": "^2.3.0",
+    "url-join": "^5.0.0",
```
gund, line 168: Btw you should also add this dependency to the `application` package.

gund (changes requested):

gund (dismissed):

Sinucid (dismissed):

gund (approved):

PR #214, opened 2023-03-30
chore: add mock server to Fulfillment app

tarashaliuk:
closes: [HRZ-2551](https://spryker.atlassian.net/browse/HRZ-2551)

[HRZ-2551]: https://spryker.atlassian.net/browse/HRZ-2551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/fulfillment/mock/index.js
```diff
@@ -0,0 +1,8 @@
+require('dotenv').config();
+const { createMockServer } = require('./server');
+
+const port = process.env.MOCK_PORT_PA || 3001;
```
gund: It's not a PA (PickingApp) but Fulfillment so let's rename it:
```suggestion
const port = process.env.ORYX_FULFILLMENT_MOCK_PORT || 3001;
```

apps/fulfillment/mock/routes.json
```diff
@@ -0,0 +1,4 @@
+{
+  "/picking-orders?filter[order.ids]=:id": "/orders/:id",
+  "/picking-orders/:id": "/orders/:id"
```
gund: There are only used in picking app and are not used in fulfillment app so please remove them.

apps/fulfillment/mock/server.js
```diff
@@ -0,0 +1,152 @@
...
+    router.use(createProxyRouter(PROXY_URL, proxyRoutes, BASE_PATH));
+  }
+
+  router.use(jsonServer.defaults({ noCors: true }));
+
+  router.post('/authorize', mapRequestToGet);
+  router.post('/token', mapRequestToGet);
+  router.post('/push-token', mapRequestToGet);
+
+  router.get('/picking-orders', (req, res) => {
```
gund: Please remove this endpoint as it's not used in fulfillment app.

libs/domain/picking/offline/feature.ts
```diff
@@ -31,7 +31,7 @@ export class OfflinePickingFeature implements AppFeature {
         PickingSyncActionHandlerService
       ),
       { provide: PickingListService, useClass: PickingListDefaultService },
-      { provide: PickingListAdapter, useClass: PickingListOfflineAdapter },
+      { provide: PickingListAdapter, useClass: PickingListDefaultAdapter },
```
gund: Why did you change this? In offline mode an offline adapter should be used instead of the default one.

apps/fulfillment/mock/db.json
```diff
@@ -0,0 +1,18811 @@
...
+      "token_type": "Bearer",
+      "expires_in": 28800,
+      "refresh_token": "def5020029b7becd9092d801c6a68fc64e3b145dc518e0a06caa930a848a0b9efd63ebf72af6738c459e1668023b47a25998f1dddb46a395bd491c10b9d0505160f506e7f4d704c25e34bcd8887c79a0a29808546bfb1ce1f7b062e8186827e6eade6c5ee90cbeac540475cfc5ebdc3f9c1cc6e2fc91657d11ea63421ea1632f5e3fe504d12443697edb6e54fb8508a366238d637892df5b9a8b8778a2dc19de28a72476cdb7328aed976f7d08ecfcdf3bc35a1a7ecfefa3fd9560d45177ce2be9be23a70b3b53f0ddf92014b1cf16e76fab102bde3327585ce2d5d75568fed534db4a43fc9b380f3a98a63e67ae440a955bec572043f09ab21550d2ab21524bbc650d88bdb11a670b39a15361266ca3545752fa95921849bf306815d57e4e50e33b49f6a6062b8c6fdee8a563e59618a217f6672286365866b60addfdc22e626dd4f5bd306e8d91d7d2d0e122d83701773714c751f0eac0787f7b7e71f1f1ea5110523aec4645d5b3eea43de5cbf13bdf0f89c2a0dcab4d09d4470c18ef08a4944ddb11eca69b8fe1dd443b74067b8102383cb3fa6fffda2cdf8d46c6acae36ef6e673fb850583f0724e9c1fe9a04060dbd5c000b4e8160ff92cd7f256591ce17daa175b63070c31ae3cd0b7201"
+    }
+  ],
+  "device-token": {
```
gund: Please remove old endpoints that are not used in fulfillment app right now such as `/device-token`, `/push-token`, `/storage-zones`, `/orders`, `/products`.

gund (changes requested):

libs/domain/picking/offline/feature.ts
gund: We should also change `DemoDataPlugin` to load picking lists from the online adapter and store it in the DB instead of the hardcoded data.
It will eliminate the need for this change.

tolerants:
I believe it's `chore`, not `feat`.

apps/fulfillment/mock/db.json
```diff
@@ -0,0 +1,888 @@
...
+      "token_type": "Bearer",
+      "expires_in": 28800,
+      "refresh_token": "def5020029b7becd9092d801c6a68fc64e3b145dc518e0a06caa930a848a0b9efd63ebf72af6738c459e1668023b47a25998f1dddb46a395bd491c10b9d0505160f506e7f4d704c25e34bcd8887c79a0a29808546bfb1ce1f7b062e8186827e6eade6c5ee90cbeac540475cfc5ebdc3f9c1cc6e2fc91657d11ea63421ea1632f5e3fe504d12443697edb6e54fb8508a366238d637892df5b9a8b8778a2dc19de28a72476cdb7328aed976f7d08ecfcdf3bc35a1a7ecfefa3fd9560d45177ce2be9be23a70b3b53f0ddf92014b1cf16e76fab102bde3327585ce2d5d75568fed534db4a43fc9b380f3a98a63e67ae440a955bec572043f09ab21550d2ab21524bbc650d88bdb11a670b39a15361266ca3545752fa95921849bf306815d57e4e50e33b49f6a6062b8c6fdee8a563e59618a217f6672286365866b60addfdc22e626dd4f5bd306e8d91d7d2d0e122d83701773714c751f0eac0787f7b7e71f1f1ea5110523aec4645d5b3eea43de5cbf13bdf0f89c2a0dcab4d09d4470c18ef08a4944ddb11eca69b8fe1dd443b74067b8102383cb3fa6fffda2cdf8d46c6acae36ef6e673fb850583f0724e9c1fe9a04060dbd5c000b4e8160ff92cd7f256591ce17daa175b63070c31ae3cd0b7201"
+    }
+  ],
```
gund: Btw `/token` response should be changed to just object without any arrays as it's part of the Oauth spec.

gund (changes requested):
What about adjusting the `DemoDataPlugin` to use mock server for loading data and storing it in the DB (see [comment](https://github.com/spryker/oryx/pull/214#discussion_r1153232201))?
Because right now the mock server is not being used by the app at all.

gund (approved):

apps/fulfillment/mock/db.json
gund: Btw this is not updated.

gund (changes requested):
One comment is not addressed yet.

gund (approved):

PR #215, opened 2023-03-30
fix: experience-composition hydration

supproduction:
adds reflect true to uid and route

closes [HRZ-2608](https://spryker.atlassian.net/browse/HRZ-2608)

[HRZ-2608]: https://spryker.atlassian.net/browse/HRZ-2608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/composition/src/experience-composition.component.ts
```diff
@@ -27,13 +35,13 @@ export class ExperienceCompositionComponent extends ContentMixin<CompositionProp
 ) {
   static styles = [compositionStyles];
 
-  @property()
+  @property({ reflect: true })
```
tobi-or-not-tobi, line 38: What about making both uid and route optional (undefined) by default?

tobi-or-not-tobi (changes requested):

libs/platform/experience/composition/src/experience-composition.component.ts
supproduction: I can. But it will be the same behavior. But instead of empty string we will emit undefined

tobi-or-not-tobi (approved):

PR #216, opened 2023-03-30
chore: signal usage

dunqan:
Relative PR with signal usage

### DIFF
--- a/libs/domain/product/src/mixins/product.mixin.ts
+++ b/libs/domain/product/src/mixins/product.mixin.ts
@@ -15,7 +15,7 @@ export declare class ProductMixinInterface
 {
   sku?: string;
   product?: Product | null;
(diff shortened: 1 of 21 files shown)


libs/domain/product/src/mixins/product.mixin.ts
```diff
@@ -15,6 +15,7 @@ export declare class ProductMixinInterface
 {
   sku?: string;
   product?: Product | null;
+  productS: Signal<Product | null>;
```
tolerants: Don't think everyone understand that `S` is a `Signal`. So suggest to use full word.
Upd: but probably it's temporary and just for example :)

dunqan: yes, also changed it aleady :)

dunqan:
will be included in a follow up

PR #218, opened 2023-03-30
chore: e2e test on search integration

nsilva-spryker:


apps/storefront-e2e/src/integration/search.cy.ts
```diff
@@ -24,7 +24,9 @@ describe('Search suite', () => {
       .shadow()
       .should('contain.text', 'Sony NEX-VG20EH');
 
-    search.getViewAllBtn().scrollIntoView().should('be.visible');
+    search.getSearchResultsWrapper().find('div').scrollTo('bottom');
```
tolerants: But it's really easy to break this test. At the same time it's impossible to understand why we are looking for this div without diving into html/css structure.
Can we probably at least create a separate method like `getSearchResultsContainer` inside `search.fragment.ts`?

tolerants (changes requested):

tolerants:
Can you please change PR title to `chore`.

apps/storefront-e2e/src/integration/search.cy.ts
nsilva-spryker: Yes, I believe having this method `getSearchResultsContainer` gives a better perception that we are trying to find the container to scroll down

tolerants (approved):

PR #220, opened 2023-03-30
feat: signals with usage

dunqan:
Signals implementation with initial usage examples.

Closes [HRZ-2361](https://spryker.atlassian.net/browse/HRZ-2361)

[HRZ-2361]: https://spryker.atlassian.net/browse/HRZ-2361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/utilities/src/decorators/hydratable/hydratable.ts
```diff
@@ -63,24 +65,47 @@ function hydratableClass<T extends Type<HTMLElement>>(
...
         this[DEFER_HYDRATION] = true;
       }
     }
 
+    willUpdate(_changedProperties: PropertyValues): void {
+      super.willUpdate(_changedProperties);
+      if (isServer) {
+        this[SIGNAL_META].effect = effect(() => {
+          this[SIGNAL_META].effectRuns++;
+          this.render();
```
supproduction, line 104: Is it mean that we have two renders on ssr, when we call it in `willUpdate` and when lit call it?

libs/base/utilities/src/signals/signal.controller.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+      requestUpdate: vi.fn(),
+    } as Partial<ReactiveControllerHost> as ReactiveControllerHost;
+  });
+
+  test('constructor should add controller and set up consumer', () => {
+    const signalController = new SignalController(host);
+    expect(host.addController).toHaveBeenCalledWith(signalController);
+  });
+
+  test('constructor should not add the same controller twice', () => {
```
supproduction:
```suggestion
  it('constructor should add controller and set up consumer', () => {
    const signalController = new SignalController(host);
    expect(host.addController).toHaveBeenCalledWith(signalController);
  });

  it('constructor should not add the same controller twice', () => {
```

libs/domain/cart/src/mixins/cart.mixin.ts
```diff
@@ -48,6 +54,15 @@ export const CartComponentMixin = <
 
     @asyncState()
     protected totalQuantity = valueType(this.cartController.getTotalQuantity());
+
+    protected $isEmpty = signal(this.cartController.isEmpty(), false);
```
supproduction, line 58: Is it convention start signals name with `$`?

libs/base/utilities/src/decorators/hydratable/hydratable.ts
dunqan: we have only one render on SSR, but we can run render method multiple times on a server, if signal used in a render will trigger SSR awaiter (and change of any signal used in last render run)

libs/base/utilities/src/signals/signal.controller.spec.ts
dunqan: 😊

libs/domain/cart/src/mixins/cart.mixin.ts
dunqan: Not yet, it's to make it possible to add them to mixins alongside asyncValue versions (which we will remove in the future, if all goes well), but we will discuss it, as it has some benefits

tobi-or-not-tobi (approved):

supproduction (approved):

PR #221, opened 2023-03-31
fix: cart layout is too small

tobi-or-not-tobi:
The two column layout has changed:
- 8 columns on large screens (was 7 before)
- full size (8 columns) on medium screens (was 5 before)
- full size (4 columns) on small screens (no change)

Aditionally, we added:
- fix icon button background
- added font
- pluralisation for discount heading

fixes [HRZ-2564](https://spryker.atlassian.net/browse/HRZ-2564)

[HRZ-2564]: https://spryker.atlassian.net/browse/HRZ-2564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/labs/src/i18n/translations/en.ts
```diff
@@ -1,6 +1,8 @@
 export default {
   'cart.totals.<count>-items':
     'My cart ({count, plural, one {one item} other {{count} items}})',
+  'cart.totals.<count>-discounts':
+    '{count, plural, one { Discount} other {Discounts}}',
```
AndreyYevtukhov: There is a space in the `{ Discount}` block - does it influence something?

libs/domain/cart/totals/components/discount/discount.component.spec.ts
```diff
@@ -108,8 +107,8 @@ describe('CartTotalsDiscountComponent', () => {
 
     it('should render label and subtotal', () => {
       const spans = element.shadowRoot?.querySelectorAll('span');
-      expect(spans?.[0].textContent).toBe('Discount');
-      expect(spans?.[1].textContent).toBe('price');
+      expect(spans?.[0].textContent).toContain('1 discounts');
```
AndreyYevtukhov, line 110: Should it be `.toContain('1 discount');` (singular)

AndreyYevtukhov: And what about adding a test for 2+ discounts?

AndreyYevtukhov (commented):
Left 2 comments

supproduction (dismissed):

libs/domain/cart/totals/components/discount/discount.component.spec.ts
tobi-or-not-tobi: we do not have any i18n labels integrated into unit tests, hence there's no plurarlisation to be tested in unit tests

AndreyYevtukhov (approved):

PR #223, opened 2023-03-31
fix: missed loading states on "Start picking" and "Proceed to picking"

tarashaliuk:
closes: [HRZ-2606](https://spryker.atlassian.net/browse/HRZ-2606)

[HRZ-2606]: https://spryker.atlassian.net/browse/HRZ-2606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sinucid (commented):
I believe we should implement loading state on picking service level and not to have bunch of local ones. This should be global thing, for example to prevent start picking of a few picking lists at once
cc @gund What do you think?

libs/domain/picking/src/services/picking-list-default.service.ts
```diff
@@ -14,8 +22,14 @@ export class PickingListDefaultService implements PickingListService {
     shareReplay({ bufferSize: 1, refCount: true })
   );
 
+  protected isStartPickingLoading$ = new BehaviorSubject<string | null>(null);
```
Sinucid:
```
protected  pickingInProgressId = new BehaviorSubject<string | null>(null);

getPickingInProgressId(): Observable<string | null> {
    return this.pickingInProgressId$;
}
```

libs/domain/picking/src/services/picking-list-default.service.ts
```diff
@@ -14,8 +22,14 @@ export class PickingListDefaultService implements PickingListService {
...
+
   constructor(protected adapter = inject(PickingListAdapter)) {}
 
+  protected setStartPickingLoading(pickingListId: string | null): void {
```
Sinucid: not needed

libs/domain/picking/src/services/picking-list-default.service.ts
```diff
@@ -37,8 +51,19 @@ export class PickingListDefaultService implements PickingListService {
     );
   }
 
-  startPicking(pickingList: PickingList): Observable<PickingList> {
-    return this.adapter.startPicking(pickingList);
+  startPicking(pickingList: PickingList): Observable<PickingList | null> {
```
Sinucid, line 51: I believe that we should keep this logic as simple as possible:

```
 startPicking(pickingList: PickingList): Observable<PickingList> {
      this.isStartPickingLoading$.next(pickingList.id);
      return this.adapter.startPicking(pickingList).pipe(
         catchError(() => this.isStartPickingLoading$.next(null)),
         tap(() => this.isStartPickingLoading$.next(null)),
      )
}

```

libs/domain/picking/src/services/picking-list-default.service.ts
```diff
@@ -37,8 +51,19 @@ export class PickingListDefaultService implements PickingListService {
...
+    return this.isStartPickingLoading(pickingList.id).pipe(
+      filter((isLoading) => !isLoading),
+      tap(() => this.setStartPickingLoading(pickingList.id)),
+      switchMap(() => this.adapter.startPicking(pickingList)),
+      tap(() => this.setStartPickingLoading(null))
+    );
+  }
+
+  isStartPickingLoading(pickingListId: string): Observable<boolean> {
```
Sinucid: not needed

libs/domain/picking/src/components/customer-note/src/customer-note.component.ts
```diff
@@ -1,16 +1,27 @@
...
-import { tap } from 'rxjs';
+import { of, switchMap, tap } from 'rxjs';
 import { styles } from './customer-note.styles';
 
 export class CustomerNoteComponent extends PickingListMixin(LitElement) {
   static styles = styles;
 
   protected routerService = resolve(RouterService);
 
+  protected isStartPickingLoading$ = this.pickingList$.pipe(
```
Sinucid: And we can update `PickingListMixin` and provide extraction of currently picking list id.

Add this to the mixin:
```
@asyncState()
protected pickingInProgressId = valueType(this.pickingListService.getPickingInProgressId());
```

and then you can manage buttons state based on the value:

```
<oryx-button ?loading=${this.pickingInProgressId === this.pickingListId or what is the correct name}>
        <button ?disabled=${!!this.pickingInProgressId}>
```

Sinucid: Don't forget to update interfaces

Sinucid (changes requested):

libs/domain/picking/src/components/customer-note/src/customer-note.component.ts
```diff
@@ -41,8 +41,14 @@ export class CustomerNoteComponent extends PickingListMixin(LitElement) {
 
       <p>${this.pickingList?.cartNote}</p>
 
-      <oryx-button>
-        <button @click=${this.onProceed}>
+      <oryx-button
+        ?loading=${this.upcomingPickingListId === this.pickingList.id}
```
Sinucid: For this case is enough just to check whether `upcomingPickingListId` truthy  or not. There are no another picking lists, and this changes can be initiated by this component itself only

`?loading=${!!this.upcomingPickingListId}`

libs/domain/picking/src/components/customer-note/src/customer-note.component.ts
```diff
@@ -41,8 +41,14 @@ export class CustomerNoteComponent extends PickingListMixin(LitElement) {
...
+      >
+        <button
```
Sinucid: `?disabled=${!!this.upcomingPickingListId}`

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
```diff
@@ -100,8 +92,17 @@ export class PickingListItemComponent
...
-        <oryx-button slot="footer" type=${ButtonType.Primary} size=${Size.Lg}>
-          <button ?disabled=${this.isDisabled} @click=${this.startPicking}>
+        <oryx-button
+          slot="footer"
+          type=${ButtonType.Primary}
+          size=${Size.Lg}
+          ?loading=${this.upcomingPickingListId === this.pickingList.id}
+        >
+          <button
+            ?disabled=${this.upcomingPickingListId &&
```
Sinucid, line 97: Should we disable only other picking lists but not that one that proceed picking? Maybe just:

`?disabled=${!!this.upcomingPickingListId}`

Sinucid (changes requested):

libs/domain/picking/src/components/customer-note/src/customer-note.component.ts
tarashaliuk: loading state makes the button inactive, so the user can't interact with it.

libs/domain/picking/src/components/picking-list-item/picking-list-item.component.ts
tarashaliuk: As I said before, the loading state makes the button inactive, so the user can't interact with it.

Sinucid (approved):

PR #224, opened 2023-03-31
feat(cli): create CLI package

gund:
- [x] Implement create app command
- [x] Implement `create-oryx` package for `npm init` support
- [x] Add unit tests

### DIFF
--- a/libs/platform/core/src/services/feature-options/default-feature-options.service.ts
+++ b/libs/platform/core/src/services/feature-options/default-feature-options.service.ts
@@ -5,14 +5,16 @@ import {
 } from './feature-options.service';
 
 export class DefaultFeatureOptionsService implements FeatureOptionsService {
-  protected options?: FeatureOptions;
+  protected options?: Record<string, object | undefined>;
 
   constructor(protected optionsArr = inject(FeatureOptions, null)) {
     this.options =
       optionsArr?.reduce((acc, options) => ({ ...acc, ...options }), {}) ?? {};
   }
 
-  getFeatureOptions(name: string): FeatureOptions {
+  getFeatureOptions<N extends keyof FeatureOptions>(name: N): FeatureOptions[N];
+  getFeatureOptions(name: string): object;
+  getFeatureOptions(name: string): object {
     return this.options?.[name.toLowerCase()] ?? {};
   }
 }
--- a/libs/template/cli/bin/cli.spec.ts
+++ b/libs/template/cli/bin/cli.spec.ts
@@ -1,0 +1,27 @@
+import { vi } from 'vitest';
+
+const mockCliApp = {
+  withOptions: vi.fn().mockReturnThis(),
+  create: vi.fn(),
+};
+
+vi.mock('@spryker-oryx/cli', () => ({ cliApp: mockCliApp }));
+
+describe('cli.ts', () => {
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it('should create cli app with arguments skipping first 2', async () => {
+    vi.stubGlobal('process', {
+      argv: ['skip1', 'skip2', 'cmd', '--arg1', '--arg2'],
+    });
+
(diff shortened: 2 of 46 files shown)


libs/platform/core/src/services/feature-options/default-feature-options.service.ts
```diff
@@ -5,14 +5,16 @@ import {
...
 
   constructor(protected optionsArr = inject(FeatureOptions, null)) {
     this.options =
       optionsArr?.reduce((acc, options) => ({ ...acc, ...options }), {}) ?? {};
   }
 
-  getFeatureOptions(name: string): FeatureOptions {
+  getFeatureOptions<N extends keyof FeatureOptions>(name: N): FeatureOptions[N];
+  getFeatureOptions(name: string): object;
+  getFeatureOptions(name: string): object {
```
supproduction, line 17:
```suggestion
  getFeatureOptions(name: string): object {
```

libs/template/cli/bin/cli.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { vi } from 'vitest';
```
supproduction, line 1: We don't need this import

libs/template/cli/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "dependencies": {
+    "@clack/prompts": "^0.6.3",
+    "picocolors": "^1.0.0",
+    "request": "^2.88.2",
+    "rxjs": "^7.5.0",
+    "unzipper": "^0.10.11"
+  },
+  "peerDependencies": {
+    "@spryker-oryx/core": "^0.4.0",
+    "@spryker-oryx/di": "^0.4.0"
```
supproduction: Is it ok that in other packages we use `latest` ?

libs/template/cli/src/commands/create.spec.ts
```diff
@@ -0,0 +1,467 @@
+import * as prompts from '@clack/prompts';
+import fs, { PathLike } from 'fs';
+import path from 'path';
+import { expect, it, vi } from 'vitest';
```
supproduction: We use global in tscofig. So exports are not necessary. I think same for other files

libs/template/cli/src/commands/create.spec.ts
```diff
@@ -0,0 +1,467 @@
...
+    get = vi.fn();
+  }
+
+  class MockNodeUtilService {
+    downloadFile = vi.fn();
+    extractZip = vi.fn();
+    copyFolder = vi.fn();
+  }
+
+  function setup() {
```
supproduction, line 21: So we agreed using setup everywhere?

supproduction (changes requested):

libs/platform/core/src/services/feature-options/default-feature-options.service.ts
gund: We can't remove that overload version as it's not going to be compatible with the generic overload and implementation.
image

libs/template/cli/bin/cli.spec.ts
gund: `vi` is not global, if I remove it I get `Cannot find name 'vi'.` so it should be imported.
Also global typings are from jest which are not quite correct so I also needed to import `expect` to make sure it's properly typed.

libs/template/cli/src/commands/create.spec.ts
gund: I need at least `vi` and `expect` imports for the typings to be correct, `it` can be removed.

libs/template/cli/src/commands/create.spec.ts
gund: It's been using in the i18n and auth so I think we are ok with using this style and beforeEach?

supproduction (dismissed):

supproduction (approved):

PR #226, opened 2023-04-02
feat: child hydration before hydration of main element

supproduction:
fixes hydration flickering 

closes [HRZ-2614](https://spryker.atlassian.net/browse/HRZ-2614)

[HRZ-2614]: https://spryker.atlassian.net/browse/HRZ-2614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/utilities/src/decorators/hydratable/hydratable.ts
```diff
@@ -159,18 +161,18 @@ function hydratableClass<T extends Type<HTMLElement>>(
       const states = this[asyncStates];
 
       setTimeout(() => {
```
dunqan: all this logic should only happen on clientside, yes?

libs/base/ui/structure/text/src/text.component.ts
```diff
@@ -39,14 +64,10 @@ export class TextComponent extends LitElement implements TextProperties {
...
 
-  protected get shouldBeTruncated(): boolean {
-    return !!Number(getComputedStyle(this).getPropertyValue('--line-clamp'));
-  }
-
   protected override render(): TemplateResult {
     return html`
       <div ${ref(this.containerRef)}>
-        <slot @slotchange=${this.setup.bind(this)}></slot>
+        <slot></slot>
```
dunqan, line 60: We dropped reacting on `slotchange` event here. Are we safe? maybe @tobi-or-not-tobi can you take a look at this one?

libs/base/ui/structure/text/src/text.component.ts
```diff
@@ -10,27 +10,52 @@ import { truncateFix } from './prehydrate';
...
+    }
+
+    super.willUpdate(changedProperties);
+  }
+
+  protected updated(changedProperties: PropertyValues): void {
+    if (
+      changedProperties.has('truncateAfter') &&
+      this.truncateAfter &&
+      !this.resizeObserver
```
tobi-or-not-tobi: let's leave this to the setup method maybe?

libs/base/ui/structure/text/src/text.component.ts
tobi-or-not-tobi: yes, we have an alternative implementation using the lifecycle method

libs/base/utilities/src/decorators/hydratable/hydratable.ts
supproduction: we don't trigger this on server side. But agree

libs/base/ui/structure/text/src/text.component.ts
tobi-or-not-tobi:
```
if (!this.truncateAfter || !this.resizeObserver) this.truncation = false;
```

libs/base/ui/structure/text/src/text.component.ts
```diff
@@ -10,27 +10,52 @@ import { truncateFix } from './prehydrate';
...
+  protected resizeObserver?: ResizeObserver;
+
+  protected willUpdate(changedProperties: PropertyValues): void {
+    if (changedProperties.has('truncateAfter')) {
+      if (this.truncateAfter) {
+        this.style.setProperty('--line-clamp', String(this.truncateAfter));
+      } else {
+        this.style.removeProperty('--line-clamp');
+      }
+    }
```
tobi-or-not-tobi: nitpick: we could simplify as `this.style.removeProperty('--line-clamp', null);` should remove the property afaik

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #232, opened 2023-04-04
fix(fa): fix start picking response handling

gund:
- [x] Deploy mock server as lambda function to Netlify
- [x] Add Netlify specific build command
- [x] Use dynamic fallback env to allow preview deploys of Netlify
- [x] Fix start picking response handling
- [x] Fix storing of picking lists without embedded products

Fixes [HRZ-2643](https://spryker.atlassian.net/browse/HRZ-2643)

[HRZ-2643]: https://spryker.atlassian.net/browse/HRZ-2643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/fulfillment/.env.netlify-test
```diff
@@ -0,0 +1,2 @@
+# ORYX_FULFILLMENT_MOCK_PROXY_URL=https://backend-api.de.demo-picking-app.cloud.spryker.toys/
+# ORYX_FULFILLMENT_MOCK_BASE_PATH=/.netlify/functions/server
```
AndreyYevtukhov: Do we need this variable? It seems like we are not using it

gund: For now we are not proxying to real bapi because login is not implemented yet but we will enable it once that is done.

gund: Actually one is required now so I've uncommented it but the other for now is not needed.

tarashaliuk:
I believe we should have separate build script for netlify as it was in PA.
```
"build-netlify": {
      "executor": "nx:run-commands",
      "outputs": ["dist/apps/fulfillment"],
      "options": {
        "commands": [
          "npx nx run fulfillment:init-env:netlify-test",
          "npx nx build fulfillment"
        ],
        "parallel": false
      }
    },
    "init-env": {
      "executor": "nx:run-commands",
      "options": {
        "command": "cp .env.template .env",
        "cwd": "apps/fulfillment"
      },
      "configurations": {
        "mock": {
          "command": "cp .env.mock .env",
          "cwd": "apps/fulfillment"
        },
        "netlify-test": {
          "command": "cp .env.netlify-test .env",
          "cwd": "apps/fulfillment"
        }
      }
    },

Sinucid (approved):

PR #233, opened 2023-04-04
chore: product refactoring

tobi-or-not-tobi:
- use signals
- use schema's

closes [HRZ-2647](https://spryker.atlassian.net/browse/HRZ-2647)

[HRZ-2647]: https://spryker.atlassian.net/browse/HRZ-2647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/product/media/src/media.component.ts
```diff
@@ -67,21 +48,17 @@ export class ProductMediaComponent extends ProductMixin(
...
 
-  protected renderVideo(src: string): TemplateResult | void {
-    return html`<oryx-video url=${src}></oryx-video>`;
-  }
-
   /**
    * Resolves the media set. When there's no mediaSet index provided, the first
    * media set is returned.
    */
   protected getMediaSet(): ProductMediaSet | undefined {
```
dunqan: could be changed to computed

libs/domain/product/src/services/product-list.service.ts
```diff
@@ -8,7 +8,7 @@ export interface ProductListService {
     qualifier: ProductListQualifier
   ): Observable<QueryState<ProductList | undefined>>;
 
-  getSearchParams(qualifier: ProductListQualifier): Record<string, string>;
+  // getSearchParams(qualifier: ProductListQualifier): Record<string, string>;
```
dunqan: left over

libs/platform/core/src/services/context/context.controller.ts
```diff
@@ -37,6 +38,10 @@ export class ContextController implements ReactiveController {
       ),
     ]).pipe(
       skip(1),
+      tap(([overrideContext, context]) => {
```
dunqan, line 41: we should have unit tests for this behavior, also maybe @supproduction can take a look, if we don't break some edge cases (when we already have context provided on the host + override for example)

libs/domain/product/card/src/card.component.ts
```diff
@@ -38,52 +29,43 @@ export class ProductCardComponent extends ProductMixin(
 
   protected context = new ContextController(this);
```
supproduction: Do we need it?

libs/domain/product/src/services/product-list.service.ts
tobi-or-not-tobi: done

libs/domain/product/media/src/media.component.ts
tobi-or-not-tobi: done

supproduction (dismissed):

dunqan (approved):

PR #234, opened 2023-04-04
feat: fetch picking list data on login

aleespryker:
Closes [HRZ-2550](https://spryker.atlassian.net/browse/HRZ-2550)

[HRZ-2550]: https://spryker.atlassian.net/browse/HRZ-2550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/offline/data-plugin.ts
```diff
@@ -0,0 +1,106 @@
...
+      this.subscription = authService
+        .isAuthenticated()
+        .pipe(
+          switchMap((authenticated) => {
+            if (authenticated) {
+              return this.clearDb(injector).pipe(
+                switchMap(() => this.populateDb(injector))
+              );
+            }
+            return this.clearDb(injector);
```
gund: We should not clear DB after a logout as in the future we will keep updating the DB state in the background using push notifications.

gund (changes requested):

gund (approved):

PR #242, opened 2023-04-04
fix: simplify picking service

Sinucid:
Simplification of picking list service that is decreasing complexity related to getting picking lists.

Also folder structures of `navigate back` and `customer note` components were aligned with the others.

closes to [HRZ-2667](https://spryker.atlassian.net/browse/HRZ-2667)

[HRZ-2667]: https://spryker.atlassian.net/browse/HRZ-2667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/services/picking-list.service.ts
```diff
@@ -2,10 +2,7 @@ import { Observable } from 'rxjs';
 import { PickingList, PickingListQualifier } from '../models';
 
 export interface PickingListService {
-  get(): Observable<PickingList[]>;
-  setQualifier(
-    qualifier: PickingListQualifier
-  ): Observable<PickingListQualifier>;
+  get(qualifier: PickingListQualifier): Observable<PickingList[]>;
   getById(id: string): Observable<PickingList | null>;
```
gund: Let's also remove getById method and use get({ id }).

gund (changes requested):

libs/domain/picking/src/services/picking-list.service.ts
gund: Let's keep it for convenience.

gund (dismissed):

gund (approved):

PR #243, opened 2023-04-04
fix: adjust selected tab chip color

aleespryker:
Closes [HRZ-2615](https://spryker.atlassian.net/browse/HRZ-2615)

[HRZ-2615]: https://spryker.atlassian.net/browse/HRZ-2615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/components/picking/picking.styles.ts
```diff
@@ -62,4 +62,8 @@ export const styles = css`
     font-size: 18px;
     font-weight: 600;
   }
+
+  oryx-tab[selected] oryx-chip {
+    background-color: var(--oryx-color-success-100);
```
gund: Assuming our latest discussion on design approach, I would assume that we should not do such style tweaks on the component level anymore?
cc @tobi-or-not-tobi

tobi-or-not-tobi: We should stick to the system here, see https://test-storybooks.netlify.app/?path=/story/ui-graphical-chip-static--states. Any reason to not use the success type here?

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -305,6 +310,25 @@ export class PickingComponent extends PickingListMixin(LitElement) {
...
+    this.onTabChange();
+  }
+
+  protected onTabChange(): void {
+    for (let i = 0; i < this.tabRefs.length; i++) {
+      const tab = this.tabRefs[i].value;
+      if (!tab) {
+        continue;
+      }
+      const chip = tab.querySelector('oryx-chip');
```
nsilva-spryker: Should we use ref again to not query directly by name('oryx-chip'), in case somehow in the future this could change ?

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -305,6 +310,25 @@ export class PickingComponent extends PickingListMixin(LitElement) {
...
+      if (tab.selected) {
+        chip?.setAttribute('appearance', 'success');
```
nsilva-spryker, line 339: Let's update unit tests, to guarantee that when the tab is selected the oryx-chip has the right appearance

nsilva-spryker (changes requested):
Looks good!
Few suggestion

nsilva-spryker (approved):

gund (approved):

PR #248, opened 2023-04-06
fix: product card edit on picking page

nsilva-spryker:
After editing a picking product, dispatch tab click event to navigate to not picked tab and scroll to respective element, refactor some code.

Closes [HRZ-2616](https://spryker.atlassian.net/browse/HRZ-2616)

[HRZ-2616]: https://spryker.atlassian.net/browse/HRZ-2616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -96,6 +101,32 @@ export class PickingComponent extends PickingListMixin(LitElement) {
     this.items[productIndex].status = ItemsFilters.NotPicked;
 
     this.items = [...this.items];
+
+    const tabs = this.buildTabs();
+    this.renderRoot
+      .querySelectorAll('oryx-tab')
```
aleespryker: Can we not use a createRef for this?

aleespryker (dismissed):

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -137,52 +165,64 @@ export class PickingComponent extends PickingListMixin(LitElement) {
   protected override render(): TemplateResult {
     const tabs = this.buildTabs();
 
-    return html`<oryx-tabs appearance="secondary" sticky shadow>
+    return html`
+      <oryx-tabs appearance="secondary" sticky shadow>
```
tarashaliuk: Use enum for `appearance`

tarashaliuk (changes requested):

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -96,6 +103,27 @@ export class PickingComponent extends PickingListMixin(LitElement) {
     this.items[productIndex].status = ItemsFilters.NotPicked;
 
     this.items = [...this.items];
+
+    this.tabRef.value?.dispatchEvent(
+      new CustomEvent('click', {
+        composed: true,
+        bubbles: true,
+      })
+    );
```
tarashaliuk: Why tab should dispatch click event?

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -137,52 +165,64 @@ export class PickingComponent extends PickingListMixin(LitElement) {
...
-                    `
-                )}
-              </div>
-              ${this.renderFinishButton(true)}
-            `,
-            () => this.renderPlaceholderComplete()
-          )}
-        </div>`
+        (tab) => html`
+          <div slot="panels" id="tab-${tab.id}" class="tab-panels">
```
tarashaliuk, line 199: Why do we need id here?

tarashaliuk (changes requested):

libs/domain/picking/src/components/picking/picking.component.ts
nsilva-spryker: This id will be connected to tabs slot 'for' attribute, to show or hide depending on it. It's the current behaviour we have for tabs component

libs/domain/picking/src/components/picking/picking.component.ts
nsilva-spryker: In this case to navigate to 'Not picked' tab, when user wants to edit a product already 'Picked' or 'Not found'

tarashaliuk: As I remember, we can change the tab using `activeTabIndex`.

nsilva-spryker: The `activeTabIndex` is being used mainly for the first rendering of tabs, if user change tabs manually the activeIndex won't update. Also, the tab component should dispatch an event to indicate to the component's owner that the selected property changed

apps/fulfillment/mock/db.json
```diff
@@ -130,21 +130,9 @@
...
                 "id": "80839794-a5dc-58e7-b95f-55667543258f"
               },
-              {
-                "type": "picking-list-items",
-                "id": "c3d7aa66-72bb-5d29-8a35-8d440910f4f6"
-              },
-              {
-                "type": "picking-list-items",
-                "id": "20415014-987d-528d-b22e-35aeba72f03a"
-              },
```
gund, line 140: Why did you change mock data?

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -37,6 +42,9 @@ export class PickingComponent extends PickingListMixin(LitElement) {
     })
   );
 
+  protected productCardRef: Ref<PickingProductCardComponent> = createRef();
+  protected tabRef: Ref<TabComponent> = createRef();
```
gund: Let's rename it to reflect that it's specific tab:
```suggestion
  protected notPickedTabRef: Ref<TabComponent> = createRef();
```

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -96,6 +104,27 @@ export class PickingComponent extends PickingListMixin(LitElement) {
...
+        top: this.productCardRef.value?.offsetTop,
+        behavior: 'smooth',
+      });
+
+      (
+        this.productCardRef.value?.renderRoot.querySelector(
+          'oryx-cart-quantity-input'
+        ) as HTMLElement
+      )?.focus();
+    }, 100);
```
gund: I wonder if there is a better way of scheduling scrolling without setTimeout with 100ms delay as it seems like an unreliable way of waiting for a re-render.
What about trying to use [`this.updateComplete`](https://lit.dev/docs/components/lifecycle/#updatecomplete) promise instead? In theory this should wait until the tabs are rendered and then you could scroll to it.

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -96,6 +104,27 @@ export class PickingComponent extends PickingListMixin(LitElement) {
...
+        left: 0,
+        top: this.productCardRef.value?.offsetTop,
+        behavior: 'smooth',
+      });
+
+      (
+        this.productCardRef.value?.renderRoot.querySelector(
+          'oryx-cart-quantity-input'
+        ) as HTMLElement
+      )?.focus();
```
gund: Can we delegate focusing to the product card component instead of reaching into it's shadowRoot?
Maybe the card can do the focus automatically when it's switched to edit mode?

gund (commented):
Looks fine, added a few suggestions.

apps/fulfillment/mock/db.json
nsilva-spryker: This picking list, had repeated products
Screenshot 2023-04-06 at 16 41 13

libs/domain/picking/src/components/picking/picking.component.ts
nsilva-spryker: Don't really know if we can inject that directly into the component :(

libs/domain/picking/src/components/picking/picking.component.ts
nsilva-spryker: We could have multiple product card component in edit mode, but only one should be focus when getting back from 'picking' or 'not found' tabs and it's the parent component(picking component) that is controlling it.

libs/domain/picking/src/components/picking/picking.component.ts
tarashaliuk: it can be ok for now, but I guess we should refactor the Tab component to make it able change tabs using prop, but not click event

tarashaliuk (dismissed):

libs/domain/picking/src/components/picking/picking.component.ts
gund: Sure that's fine, what I mean is to add facade method to the product card component that will focus the input internally and here you would just call that method instead of reaching into the component internals.

gund (dismissed):

gund (approved):

PR #257, opened 2023-04-09
feat: meta service

supproduction:
adds meta service

closes [HRZ-2635](https://spryker.atlassian.net/browse/HRZ-2635)

[HRZ-2637]: https://spryker.atlassian.net/browse/HRZ-2637

[HRZ-2635]: https://spryker.atlassian.net/browse/HRZ-2635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/core/src/services/meta/default-meta.service.ts
```diff
@@ -0,0 +1,44 @@
+import { MetaService, TagAttributes, TagDefinition } from './meta.service';
+
+export class DefaultMetaService implements MetaService {
+  addTags(tags: TagDefinition[]): void {
```
tolerants: We are adding Nodes/Elements, not tags.

libs/platform/core/src/services/meta/meta.service.ts
```diff
@@ -0,0 +1,44 @@
+export const MetaService = 'oryx.MetaService';
```
tolerants: I'm not really sure about the name of the service as long as we can add whatever we want here, there are no restrictions. And at the same time we can only add it to the `head`. So for now it more looks like a `HeadDOMService` or something.

libs/platform/core/src/services/head-dom/head-dom.service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export interface ElementDefinition {
+  name: string;
+  attrs: ElementAttributes;
+}
+
+export interface HeadDOMService {
+  addElements(definitions: ElementDefinition[]): void;
+  addElement(definition: ElementDefinition): void;
+  updateElement(definition: ElementDefinition): void;
```
tobi-or-not-tobi:
```suggestion
  add(definitions: ElementDefinition | ElementDefinition[]): void;
  update(definition: ElementDefinition): void;
```

libs/platform/core/src/services/head-dom/head-dom.service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+interface MetaAttributes {
+  charset?: string;
+  content?: string;
+  'http-equiv'?: string;
+  id?: string;
+  itemprop?: string;
+  url?: string;
+  name?: string;
+}
```
tobi-or-not-tobi: move into `head-dommodel.ts`?

tobi-or-not-tobi (dismissed):

libs/template/application/src/ssr/render.ts
```diff
@@ -45,6 +63,19 @@ export const renderApp = async (
...
-  return stream;
+
+  const component = element.strings[0];
+  const componentName = component.substring(
+    component.indexOf('<') + 1,
+    component.indexOf('>')
+  );
+  const htmlAttrs = headDefinition?.find((def) => def.name === 'html')?.attrs;
+  const html = template
+    .replace('<html', `<html ${headDom.getElementAttributes(htmlAttrs ?? {})}`)
```
dunqan: this logic ideally will be covered inside the service, as service knows what to do

libs/platform/core/src/services/page-head/default-page-head.service.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export class DefaultPageHeadService implements PageHeadService {
+  addElements(definitions: ElementDefinition | ElementDefinition[]): void {
+    if (!Array.isArray(definitions)) {
+      definitions = [definitions];
+    }
+
+    for (const definition of definitions) {
+      if (definition.name === 'html') {
+        this.updateHtmlElement(definition.attrs);
```
dunqan: I think we concluded, we don't want to handle 'html' element as a special case, but as a separate method? 🤔 
and we already have separate method I see `updateHtmlElement`

libs/platform/core/src/services/page-head/page-head.service.ts
```diff
@@ -0,0 +1,16 @@
+import { ElementAttributes, ElementDefinition } from './page-head.model';
+
+export const PageHeadService = 'oryx.PageHeadService';
+
+export interface PageHeadService {
+  addElements(definitions: ElementDefinition | ElementDefinition[]): void;
+  updateElement(definition: ElementDefinition): void;
+  updateHtmlElement(attrs: ElementAttributes): void;
```
dunqan: do we need both `updateHtmlElement` and `setAttributes`?

we just want to be able to crud meta tags with the service (where "read" is not what we need now, but could be added for completness, the same for "delete") so api should be fairly concise  (+ special case for managing html attributes)

libs/platform/core/src/services/page-head/page-head.service.ts
```diff
@@ -0,0 +1,16 @@
+import { ElementAttributes, ElementDefinition } from './page-head.model';
+
+export const PageHeadService = 'oryx.PageHeadService';
+
+export interface PageHeadService {
+  addElements(definitions: ElementDefinition | ElementDefinition[]): void;
```
dunqan: addTag still works better for me here (for link, script, title or meta, all of those are tags) than 'element', but happy to hear your opinion

dunqan: or maybe just `add` and that's it?

dunqan: looks like title, link, etc. are also considered meta-related elements (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta), then 
```
export interface PageMetaService {
  add(definition: MetaElement | MetaElement[])
  // ...
```
looks quite concise?

supproduction: But we decided them PageMeta incorrect name. Cause we can add also title\script which is not meta things

libs/platform/core/src/services/page-head/page-head.service.ts
supproduction: IMO we don't . We decidet to have separate method for `html`

libs/platform/core/src/services/page-head/page-head.service.ts
supproduction: ok, I'll fix it

libs/platform/core/src/services/page-head/page-head.service.ts
supproduction: dropped

libs/platform/core/src/services/page-meta/page-meta.service.ts
```diff
@@ -0,0 +1,15 @@
+import { ElementAttributes, ElementDefinition } from './page-meta.model';
+
+export const PageMetaService = 'oryx.PageMetaService';
+
+export interface PageMetaService {
+  add(definitions: ElementDefinition | ElementDefinition[]): void;
+  update(definition: ElementDefinition): void;
+  setAttributes(attrs: ElementAttributes, element: HTMLElement): void;
```
dunqan: To avoid confusion and overusing this one, I'd chage it either to
```suggestion
  setHtmlAttributes(attrs: ElementAttributes): void;
```
or limiting elements
```suggestion
  setAttributes(attrs: ElementAttributes, element: 'html' | 'body'): void;
```

libs/template/application/src/ssr/render.ts
```diff
@@ -22,12 +32,20 @@ export const renderApp = async (
...
+  globalThis.location = route;
+  /* eslint-disable @typescript-eslint/no-non-null-assertion */
+  const injector = orchestrator.findPlugin(InjectionPlugin)!.getInjector();
+  const headDefinition = orchestrator
+    .findPlugin(ThemePlugin)!
+    .getHeadDefinition();
+  /* eslint-enable @typescript-eslint/no-non-null-assertion */
+  const routerService = injector.inject(RouterService);
+  const awaiter = injector.inject(SSRAwaiterService);
+  const headDom = injector.inject(PageMetaService) as ServerPageMetaService;
```
dunqan: let's rename this const to make it consistent

libs/platform/core/server/services/page-meta/server-page-meta.service.ts
```diff
@@ -0,0 +1,58 @@
...
+      }
+
+      stream += `\n<${name}${this.setAttributes(attrs)} />`;
+    }
+
+    this.template = this.template.replace('</head>', `${stream}\n</head>`);
+  }
+
+  setHtmlAttributes(attrs: ElementAttributes): void {
+    this.template = this.template.replace(
```
dunqan, line 38: not sure if this one will work on the server app, if used inside some component or service?

dunqan (approved):

PR #259, opened 2023-04-10
test: SF addresses E2E tests

AndreyYevtukhov:
closes https://spryker.atlassian.net/browse/HRZ-2373


apps/storefront-e2e/src/test-data/default-address.ts
supproduction, line 1: Suggest, change folder name to `mocks` as we use everywhere. 
And for each file name add prefix `mock-` and for each mocked const/class... `Mock...\mock...`

supproduction (changes requested):

apps/storefront-e2e/src/test-data/default-address.ts
AndreyYevtukhov: these are not mocks 

> A mock - is a simulated or artificial version of a software component that is used for testing purposes. The purpose of using a mock is to isolate the component being tested from its dependencies, such as databases, networks, or other services.

but real data used for testing (test-data)

> The test data is used to verify that the software under test is functioning correctly and producing the expected outputs. This data can include different types of inputs such as text, numbers, dates, files, and other parameters. The use of pre-defined test data ensures that the same tests are executed consistently, and it allows for repeatable test results.

supproduction: Should we separate mocks in libs by test-data and mocks by these definition or unify this folder\files\names in e2e?

supproduction: cc @dunqan @tobi-or-not-tobi

dunqan: If it's a part of e2e app only, then 'test-data' may work, as long as we won't mix them togeter (i.e. having both mocked data and test data in e2e, as then they are basically the same, or rather test data is a subset of mock data). 

In out libs, mocks are a bit more than just data (we have mocked services, etc.).

AndreyYevtukhov: When we will have real mocks for E2E (for example mocked BE responses) - we will have that data in the `mock` folder. 

But for now - we only need `test data` to use real users/orders/etc. and verify that our FE<->BE integration works as it should (without any mocks)

supproduction (approved):

PR #264, opened 2023-04-12
feat: navigation items tests

Sinucid:
closes to [HRZ-2395](https://spryker.atlassian.net/browse/HRZ-2395)

[HRZ-2395]: https://spryker.atlassian.net/browse/HRZ-2395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    `);
+  });
+
+  it('passes the a11y audit', async () => {
+    await expect(element).shadowDom.to.be.accessible();
+  });
+
+  it('should render oryx-heading', () => {
+    expect(element).toContainElement('oryx-heading');
+  });
```
tarashaliuk: We can also check a text inside oryx-heading

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    beforeEach(async () => {
+      element = await fixture(html`
+        <oryx-site-navigation-button icon="test"></oryx-site-navigation-button>
+      `);
+    });
+
+    it('should render oryx-icon', () => {
+      expect(element).toContainElement('oryx-icon');
+    });
+  });
```
tarashaliuk: Also, you can check the `type` attribute to be sure that correct icon will be rendered

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+  describe('when badge is provided', () => {
+    beforeEach(async () => {
+      element = await fixture(html`
+        <oryx-site-navigation-button badge="test"></oryx-site-navigation-button>
+      `);
+    });
+
+    it('should render mark element', () => {
+      expect(element).toContainElement('mark');
+    });
```
tarashaliuk: check mark content

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    beforeEach(async () => {
+      element = await fixture(html`
+        <oryx-site-navigation-button></oryx-site-navigation-button>
+      `);
+    });
+
+    it('should not render oryx-heading', () => {
+      expect(element).not.toContainElement('oryx-heading');
+    });
+  });
```
tarashaliuk: you can also check that `oryx-icon`, `mark` and `a` shouldn't render

tarashaliuk (changes requested):

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
Sinucid: Not sure that this is necessary. There is no any other conditions inside, and if it text passes `when` directive and does not render then -> we have much bigger problem =)

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
Sinucid: The same stuff, `when` directive is doing all necessary work, it's too much to check it twice

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
Sinucid: The same

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
Sinucid: It is double work, not sure that this is necessary. We have a conditional rendering in the code and it's enough just to test the condition when it should render something. Otherwise means that it should not

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
tarashaliuk: From a testing perspective, we are creating the unit tests to be sure about component functionality and to warn the developer if he breaks something. In the future, somebody can accidentally remove the text rendering or modify the text - tests will pass it without any warning or error. Ideally, every line of code should be covered by tests.
@AndreyYevtukhov what do you think about it?

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
tarashaliuk: We should be sure that it works correctly. Somebody can change the conditions and it will render this stuff when we are not expecting it. Unit tests should catch it before somebody will mention it on CR or even on production :D

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
Sinucid: We should not test the code itself, but we should test the independent behavior related to current piece of functionality. We should not make some predictions, we can't cover all future cases. 

Here is a simple conditional rendering: if we have a text -> we render a block with it. We should not care about how `when` directive works, or how `oryx-heading` renders the text. These are behaviors of that pieces of functionality, that are tested separately in their testing files. And if in the future `oryx-heading` will require an attribute to show the text this is not a case for this test, this is a case of `oryx-heading` and its behavior. 

Thanks god we have a combination of testing approaches (unit  + e2e + regression) that in theory should cover this case that you described

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
Sinucid: but again, if someone makes some changes then he|she should align unit-tests. We should not cover future cases. And the same, conditional rendering here is area of responsibility  of `when` directive. And for falsy condition we even do not have separate render function. And if we test this falsy result -> we will test behavior of `when` that is out of our current testing behavior

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
AndreyYevtukhov: So, we are talking about unit tests. 

And the main ideas behind unit tests are:
- **Isolation**: Unit tests are designed to test _individual units of code in isolation_. This means that any dependencies that the unit has should be mocked or stubbed, so that the unit can be tested in a controlled environment. By isolating the unit from its dependencies, unit tests can help identify and isolate defects more easily.
- **Simplicity**: Unit tests should be simple and easy to understand. They should test _a single behavior or feature of the code_, and be focused on verifying that the code behaves correctly under different scenarios.

In our case - the component under the test is the `navigation button component` meaning that unit tests should cover only ITS functionality (and nothing else)

---

Ok, now let's go further and check this line:

https://github.com/spryker/oryx/blob/b116279c7d8fdfe5d74d117c2d0d1bbbf3b264aa/libs/domain/site/navigation-button/src/navigation-button.component.ts#L21

The functionality can be described as a set of requirements:
- **If the user adds an icon (while configuring the component) - that exact icon should be rendered inside the button** 
- **If the user does not provide an icon -> icon should not be rendered**

---

How to test it and prove that these requirements are implemented in the code? 

If we have a test like: 
```js
describe('when icon is provided', () => {
  beforeEach(async () => {
    element = await fixture(html`
      <oryx-site-navigation-button icon="test"></oryx-site-navigation-button>
    `);
  });

  it('should render oryx-icon', () => {
    expect(element).toContainElement('oryx-icon');
  });
});
```

What it proves? It just checks that `oryx-icon` is rendered, but it does NOT PROVE that the icon rendered was selected by the user.

If I change the code to:

```js
${when(this.icon, () => html`<oryx-icon type="whatever_is_here_I_am_a_bug"></oryx-icon>`)}
```

the test will be green (because the icon is still there) - but the `this.icon` param is completely ignored now, meaning we have a bug in the code which is not caught by unit tests (because the requirement to render THE EXACT ICON that the user selected DOES NOT WORK now as intended). 

---

To cover that requirement we have to change the test to:

```js
describe('when icon is provided', () => {
  beforeEach(async () => {
    element = await fixture(html`
      <oryx-site-navigation-button icon="test"></oryx-site-navigation-button>
    `);
  });

  // changed the name of the test to be more accurate
  it('should render oryx-icon with provided type', () => {
    expect(element).toContainElement('oryx-icon');
   
    const icon = element.getIcon(); // pseudocode
    // this assertion PROVES that the requirement to render the correct icon is covered
    // and it will catch the bug described above (if someone will hardcode the type, ignoring user input)
    expect(icon).toHaveType('test'); // pseudocode 
  });
});
```

Is it enough? Actually... no :)

---

There is another requirement: "If the user does not provide an icon -> it is not rendered"

Do we have a test (and proof) that this works? Not actually.
We know what happens when the icon is provided, but we have no idea what happens if it is not (or `this.icon=false`)

(I don't know how the `when` function works, but I'll assume that it is a simple `if/else`)

If we won't add other tests and someone will accidentally change the code to:

```js
${when(this.icon, () => html`<oryx-icon type="this.icon"></oryx-icon>`, html`<oryx-text>I might be a bug</oryx-text>`)}
```
we will have a bug in the case of `this.icon=false`, because we had a requirement that **nothing should be rendered**, but now some text is rendered. 

The test to cover it might look simple:
```js
describe('when icon is NOT provided', () => {
  beforeEach(async () => {
    element = await fixture(html`
      <oryx-site-navigation-button></oryx-site-navigation-button>
    `);
  });

  it('oryx-icon is not rendered', () => {
    expect(element).toNOTContainElement('oryx-icon'); // pseudocode
  });
});
```
but it also won't catch that `oryx-text` rendering (because there is no icon, so everything is OK?)

A good one should look like this:
```js
describe('when no configurations are provided', () => {
  beforeEach(async () => {
    element = await fixture(html`
      <oryx-site-navigation-button></oryx-site-navigation-button>
    `);
  });

  it('empty component should be rendered', () => {
    // default component checks, aka
    expect(element).toNOTContainElement('oryx-icon'); // pseudocode
    expect(element).toNOTContainElement('oryx-link'); // pseudocode
    expect(element).toNOTContainElement('oryx-heading'); // pseudocode

    //This number of elements check will catch that issue with oryx-text because we expect NOTHING to be rendered
    // but some element appeared
    expect(element.children.length).toEq(0); // pseudocode, meaning that NOTHING is rendered inside
  });
});
```

---

So, to make unit tests better, in this case, we have to:
- add a check for icon `type` to prove that the requirement **If the user adds an icon (while configuring the component) - that exact icon should be rendered inside the button** is implemented and works as expected 
- add tests to cover a case without an icon, to prove that the requirement **If the user does not provide an icon -> icon should not be rendered** is implemented and works as expected

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
tarashaliuk: I'm not talking that we should have those checks in `when text is not provided` scope. But I think that it is good to test the component without any provided attribute. 

- when no attribute provided:
  - should not render `oryx-heading` component
  - should not render `oryx-icon` component
  - should not render `mark` tag 
  - ...

I think that we should test it to be sure that our conditions work correctly even if the false condition renders nothing, but we should check it.

libs/domain/site/navigation-button/src/navigation-button.component.spec.ts
dunqan: I was asked for an opinion :) 

IMO it boils down to clear requiements. I'd try to find a balance, for sure test all requirements, but use common sense with the rest, and not cover all trivial cases, go with practicity in mind (rather than too theoretical one), to keep the investment return from added effort (writing test unit test) on a good level. 

Like: 
- "If the user does not provide an icon -> icon should not be rendered" - was this an original requirement? If yes, should be covered. If no, it's more about covering invalid input case, than expected behavior.
- "empty component should be rendered" -> this is even more specific than "icon should not be rendered", is it covering the requirement, or going into implementation detail? (often could be a pain, as some simple change may make it failing, and require rewritting). 

Also, we should expect some thoroughness from devs. It's hard to imagine, that someone will add second condition "html`<oryx-text>I might be a bug</oryx-text>`" "by accident", so either new requirement was added (was not specified clearly earlier, so should be covered afterwards) or requirements changed (failing unit test won't save us from anything here, and we will rewrite it anyway). 

So, common sense from my POV ;)

AndreyYevtukhov: > IMO it boils down to clear requirements.

Totally agree :) 

---

> thoroughness from devs and "by accident"

Also true. But sometimes there are cases when code is actually changed by accident (e.g. replace in all files during the refactoring), and it is nice to have it covered))) 

But in any case - the developer should not create "imaginary" cases which do not provide a value (=code coverage). 

All tests should go from the requirements and common sense. 

Also, it is a good practice to have guidance from coverage reports - they can highlight which lines/statements/branches are not covered and help you to improve the tests. But now I wonder how lines with `with()` are being analysed by coverage tools :)

Sinucid: Wow, holy war is starting. Maybe will be easier just to answer for the question: How far must we go when write unit tests? From my perspective we should not test implementation details of other separate parts. 

For example `oryx-icon`. Here and now we don't know what types of icons it will supports in the future, and if that svg icon that we checking is dropped or renamed in the future, we will get a broken test without changing inner functionality. This is definitely implementation details of `oryx-icon`, we should not care about it here

Another example, `when` directive. I believe that testing of both cases has sense only if we have render funcs for both of them:

```javascript
${when(
  this.icon, 
  () => html`<oryx-icon type="this.icon"></oryx-icon>`, 
  () => html`<oryx-text>I might be a bug</oryx-text>`
)}
``` 

Or if we have default option that turning on this feature by default. In this case we should test falsy case manually. 
Otherwise we are testing implementation details of `when`. Don't we trust its implementation?

tarashaliuk: Firstly we need to understand for what reason we are writing the unit tests. I think that developers create unit tests to convert business requirements into checks that will control themselves from unwanted behavior of components now and in the future.
From my POV we should test all the stuff that we can control in component. Regarding the `oryx-icon`, we can control provided attribute `type` - let's test it. Also, we are slotting the text into `oryx-header` - let's test this text. In the case of having `when` directive without false, we should also test that we have NOTHING, because it is a business requirement.

tarashaliuk (approved):

PR #267, opened 2023-04-13
feat: support observables in computed

dunqan:
Add built-in support for observables in computed. 

When result of computation will return an observable, it will be transparently converted to ConnectableSignal. 

Closes [HRZ-2678](https://spryker.atlassian.net/browse/HRZ-2678)

[HRZ-2678]: https://spryker.atlassian.net/browse/HRZ-2678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/base/utilities/src/signals/signal-from.spec.ts
+++ b/libs/base/utilities/src/signals/signal-from.spec.ts
@@ -1,5 +1,6 @@
-import { SignalConsumer } from '@spryker-oryx/utilities';
-import { BehaviorSubject, of } from 'rxjs';
+import { wait } from '@spryker-oryx/utilities';
+import { BehaviorSubject, delay, of } from 'rxjs';
+import { SignalConsumer } from './core/signals';
 import { signalFrom, SignalObservable } from './signal-from';
 
 describe('signalFrom', () => {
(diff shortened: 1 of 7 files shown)


tobi-or-not-tobi (dismissed):

libs/base/utilities/src/signals/signal-from.spec.ts
```diff
@@ -11,16 +12,26 @@ describe('signalFrom', () => {
...
 
-  it('should connect and update the value when the observable emits', () => {
+  it('should connect and update the value when the async observable emits', async () => {
     const observable = new BehaviorSubject(42);
-    const signal = signalFrom(observable);
-    expect(signal()).toBeUndefined();
+    const signal = signalFrom(observable.pipe(delay(0)));
+    expect(signal()).toBe(undefined);
     signal.connect();
+    await wait(0);
```
supproduction: can we use `await nextFrame()` instead?

supproduction: or use fake timer?

libs/base/utilities/src/signals/signal.spec.ts
```diff
@@ -15,3 +16,60 @@ describe('signal', () => {
...
+    const c = computed(() => interval(period()));
+
+    const values: any[] = [];
+    const maxValues = 3;
+
+    const ef = effect(() => {
+      values.push(c());
+    });
+
+    await wait(100 * maxValues + 50);
```
supproduction: Can we use fake timer here?

supproduction (changes requested):

supproduction (approved):

PR #273, opened 2023-04-14
fix: fa e2e issues

AndreyYevtukhov:
closes https://spryker.atlassian.net/browse/HRZ-2729


waitOnConfig.json
gund, line 1: Let's use kebab-case naming `wait-on-config.json`.

gund (changes requested):

gund (approved):

PR #277, opened 2023-04-18
feat: meta improvements

supproduction:
improves meta service 
adds metas resolver
fixes ssr

closes [HRZ-2743](https://spryker.atlassian.net/browse/HRZ-2743)

[HRZ-2743]: https://spryker.atlassian.net/browse/HRZ-2743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/core/src/services/page-meta/page-meta-resolver/page-meta-resolver.service.ts
```diff
@@ -0,0 +1,31 @@
+import { OnDestroy } from '@spryker-oryx/di';
+import { Observable } from 'rxjs';
+import { AppInitializer } from '../../app-initializer';
+import { ElementDefinition } from '../page-meta.model';
+
+export const PageMetaResolverService = 'oryx.PageMetaResolverService';
+export const PageMetaResolver = 'oryx.PageMetaResolver*';
+
+export interface ElementResolver extends Partial<ElementDefinition> {
```
dunqan: Can we use type union type, to improve typesafety?

libs/platform/core/src/services/page-meta/page-meta-resolver/default-page-meta-resolver.service.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  constructor(
+    protected meta = inject(PageMetaService),
+    protected resolvers = inject(PageMetaResolver)
+  ) {}
+
+  initialize(): void {
+    this.subscription.add(
+      this.data$.subscribe((data) => {
+        console.log(data);
```
dunqan: 👀

libs/platform/core/src/services/page-meta/page-meta.service.ts
```diff
@@ -3,7 +3,10 @@ import { ElementAttributes, ElementDefinition } from './page-meta.model';
 export const PageMetaService = 'oryx.PageMetaService';
 
 export interface PageMetaService {
-  add(definitions: ElementDefinition | ElementDefinition[]): void;
+  add(
+    definitions: ElementDefinition | ElementDefinition[],
+    forceCreation?: boolean
```
dunqan: Can you explain, why we need it?

libs/platform/core/server/services/page-meta/server-page-meta.service.ts
```diff
@@ -41,16 +19,44 @@ export class ServerPageMetaService extends DefaultPageMetaService {
     );
   }
 
+  getTemplateHtml(template: string): string {
```
dunqan, line 22: this one still won't work in ssr for all cases?

libs/platform/core/server/services/page-meta/server-page-meta.service.ts
```diff
@@ -6,32 +6,10 @@ import {
...
-      }
-
-      stream += `\n<${name}${this.setAttributes(attrs)} />`;
-    }
-
-    this.template = this.template.replace('</head>', `${stream}\n</head>`);
+    this.metas = Array.isArray(definitions) ? definitions : [definitions];
   }
 
   setHtmlAttributes(attrs: ElementAttributes): void {
```
dunqan, line 25: this one will won't work with ssr?

libs/platform/core/src/services/page-meta/page-meta.service.ts
supproduction: Right now if we have the same element in the head we do nothing. Force creation forcibly removes this element and recreate new one (e.g load page - add title => go to the product - recreate title attribute)

libs/platform/core/server/services/page-meta/server-page-meta.service.ts
supproduction: It works

libs/platform/core/src/services/page-meta/page-meta-resolver/default-page-meta-resolver.service.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+      await nextFrame();
+
+      expect(mockMetaService.add).toHaveBeenCalledWith(
+        [
+          { name: 'title', attrs: { text: 'b' } },
+          { name: 'og:meta', attrs: { description: 'b' } },
+          { name: 'link', attrs: { href: 'a' } },
+          {
+            name: 'link',
+            id: 'link:a',
```
dunqan: what's the role of `id` here?

libs/platform/core/src/services/page-meta/page-meta-resolver/page-meta-resolver.service.ts
```diff
@@ -0,0 +1,31 @@
...
+  script?: string;
+  title?: string;
+  id?: string;
+}
+
+export interface PageMetaResolverService extends AppInitializer, OnDestroy {
+  getTitle(): Observable<string | undefined>;
+}
+
+export interface PageMetaResolver {
```
dunqan, line 13: let's move this type out, to separate file

libs/platform/core/src/services/page-meta/page-meta-resolver/page-meta-resolver.service.ts
```diff
@@ -0,0 +1,31 @@
...
+  getScore(): Observable<number>;
```
dunqan: can we make it optional? (assume it's `0` by default? or better 1, so normal?)

Also, what about some enum with -1 : not used, 0 : fallback, 1: default?

libs/platform/core/src/services/page-meta/page-meta.service.ts
dunqan: but we just should update the text, we don't have to recreate an attribute?

apps/storefront/src/index.html
```diff
@@ -1,7 +1,9 @@
...
 <html>
-  <head></head>
-  <body style="margin: 0">
+  <head>
+    <meta charset="UTF-8" />
+  </head>
+  <body>
     <root-app></root-app>
   </body>
   <script type="module" src="/app.ts"></script>
```
tobi-or-not-tobi, line 9:
```suggestion
  <script type="module" src="./app.ts"></script>
```

libs/platform/core/src/services/page-meta/page-meta-resolver/page-meta-resolver.model.ts
```diff
@@ -0,0 +1,11 @@
...
+  title?: string;
+  [key: string]: string | undefined;
+}
+
+export const enum ResolverScore {
+  NotUsed = -1,
+  Fallback = 0,
+  Default = 1,
+  OverrideDefault = 2,
+}
```
tobi-or-not-tobi: I'm not sure about these. Agreed that having a few standard numbers is great, but we'd need to be able to override the override.

libs/domain/search/src/services/resolvers/category-page-title-meta.resolver.ts
```diff
@@ -0,0 +1,63 @@
...
+    return this.router
+      .currentQuery()
+      .pipe(
+        map((query) =>
+          query?.category
+            ? ResolverScore.OverrideDefault
+            : ResolverScore.NotUsed
+        )
+      );
+  }
```
tobi-or-not-tobi: what about we have a base score (say 0), and then we get additional points. In this example it would add 1 point since it's a category. 

Say we have a special resolver for Canon categories, we'd get another point. 
Say we have a special resolver Canon in week 41, we'd get another point.

tobi-or-not-tobi (changes requested):

apps/storefront/src/index.html
supproduction: This approach might break hydration with relative path

libs/domain/search/src/services/resolvers/category-page-title-meta.resolver.ts
supproduction: Unfortunately we can't use here reliable calculation between resolvers as they are separate

tobi-or-not-tobi: Correct. If we have other teams inside Spryker creating resolvers (say PDP for marketplace), as well as partners add resolvers as well as customers write theirs. We cannot avoid conflicts.

tobi-or-not-tobi (approved):

PR #279, opened 2023-04-18
chore: refactored layout

tobi-or-not-tobi:
Changes:
- Responsive layouts are back
- Layout specific styles are only applied when needed (e.g. grid vs carousel vs column, etc) and do not pollute the DOM unnecessarily
- Screen size driven styles are added only when used; each screen size can contribute to the base styles regardless.
- The grid system has been refactored to better calculate the grid size; the column width is calculated by the root container size to ensure a consistent width, but will at the same time not leak out of the available space. As a result, the layout of PLP is much more stable.
- Page bleeding is introduced (formerly known as max-size) to align better with industry naming. Also, carousel now allows to bleed out to have carousel items running outside the container size.
- Layout style rules are now created by the layout logic (controller, service) and can be conveniently applied by the new LayoutMixin; other components (like the product list) only need to leverage the LayoutMixin to leverage layout. 
- The Layout component is no longer used inside compositions and the product list component, instead new layout mixin is used. This avoid additional DOM nodes for every composition. 
- PropertySignals are used to observe property updates directly inside signal logic.
- Layout types and layout properties (sticky, bleed, overlap) are provided by options or properties and can be refined per screen size. xs, xm, md, lg and xl are standard supported screen sizes. 
- Tabular styles are dropped for the moment as they do not fit with the new design; they will come back (similar to accordions whenever we roll out the concept of layout plugins).
- New overlap layout feature introduce to have multiple layouts on top of each other.

closes [HRZ-2719](https://spryker.atlassian.net/browse/HRZ-2719), [HRZ-2708](https://spryker.atlassian.net/browse/HRZ-2708)
fixes [HRZ-2737](https://spryker.atlassian.net/browse/HRZ-2737)

[HRZ-2719]: https://spryker.atlassian.net/browse/HRZ-2719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HRZ-2708]: https://spryker.atlassian.net/browse/HRZ-2708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-2737]: https://spryker.atlassian.net/browse/HRZ-2737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/src/mixins/layout.mixin.ts
```diff
@@ -0,0 +1,78 @@
...
+        rules,
+        this.uid
+      );
+      const graph = this.layoutController.getLayoutInfos(props, rules);
+
+      return this.layoutService
+        .getStyles(graph)
+        .pipe(map((layoutStyles) => `${layoutStyles}\n${componentStyles}`));
+    });
+  }
```
tobi-or-not-tobi: what about having only the controller as a deps here? So we would make `getLayoutInfos` a protected method in the contoller, and only have `return this.layoutController.getStyles()`;

libs/platform/experience/layout/src/stories/demo.stories.ts
```diff
@@ -1,67 +1,87 @@
...
-      control: { type: 'number' },
-      table: { category: 'DesignTokens' },
+    // layoutCols: {
+    //   control: { type: 'number' },
+    //   table: { category: 'DesignTokens' },
+    // },
+    // layoutFactor: {
+    //   control: { type: 'number' },
+    //   table: { category: 'DesignTokens' },
+    // },
```
supproduction: Do we need it?

supproduction: We check lots of commented code here

libs/platform/experience/src/services/layout/default-layout.builder.ts
```diff
@@ -135,41 +125,67 @@ export class DefaultLayoutBuilder implements LayoutBuilder {
 
     add(
       {
-        '--cols': data.columnCount,
-        '--grid-column': data.gridColumn,
-        '--grid-row': data.gridRow,
-        '--span': data.span,
-        '--z-index': data.zIndex,
-        '--rotate': data.rotate,
+        // '--cols': data.columnCount,
```
supproduction: Do we need it?

supproduction (changes requested):

libs/platform/experience/src/services/layout/default-layout.builder.ts
tobi-or-not-tobi: no, we replaced it with --col-pos and --row-pos

supproduction (dismissed):

supproduction (approved):

PR #280, opened 2023-04-18
feat: e2e test for start picking with customer note

tarashaliuk:
closes: [HRZ-2714](https://spryker.atlassian.net/browse/HRZ-2714)

[HRZ-2714]: https://spryker.atlassian.net/browse/HRZ-2714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/fulfillment-e2e/src/integration/start-picking-picklist-with-customer-note.cy.ts
```diff
@@ -0,0 +1,126 @@
...
+    customerNoteFragment
+      .getIllustration()
+      .invoke('outerHeight')
+      .should('be.eq', 250);
+  });
+
+  it('should display back button', () => {
+    customerNoteFragment.getNavigateBackButton().should('be.visible');
+
+    customerNoteFragment.getNavigateBackButton().click();
```
Sinucid, line 64: click?

Sinucid, line 52: click? For what reason?

Sinucid (changes requested):

apps/fulfillment-e2e/src/integration/start-picking-picklist-with-customer-note.cy.ts
tarashaliuk: leftover. removed it

Sinucid (dismissed):

apps/fulfillment-e2e/src/integration/start-picking-picklist-with-customer-note.cy.ts
```diff
@@ -0,0 +1,122 @@
...
+      .within(() => {
+        pickingListsFragment
+          .getIdentifier(true)
+          .invoke('text')
+          .then((text) => {
+            identifier = text.trim();
+
+            pickingListsFragment.getStartPickingButton(true).click();
+          });
+      });
```
AndreyYevtukhov: It would be good to extract this into a method in `pickingListsFragment`, e.g. `pickingListsFragment.getIdentifier()`

Also, instead of using `let identifier;` we should use aliases - see https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Variables for details

apps/fulfillment-e2e/src/integration/start-picking-picklist-with-customer-note.cy.ts
```diff
@@ -0,0 +1,122 @@
...
+  it('should navigate to correct url', () => {
+    cy.url().should('include', `/customer-note-info/${identifier}`);
+  });
+
+  it('should display customer note', () => {
+    customerNoteFragment
+      .getNote()
+      .should('be.visible')
+      .and('contain', 'Some cart note');
+  });
```
AndreyYevtukhov: Atomic tests are very cool and I like that)
but, we use Cypress Dashboard and pay for EACH test (basically for each `it` call). 

That's why we should combine checks it 1-2 tests

apps/fulfillment-e2e/src/integration/start-picking-picklist-with-customer-note.cy.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    customerNoteFragment
+      .getIllustration()
+      .invoke('outerWidth')
+      .should('be.eq', 250);
+
+    customerNoteFragment
+      .getIllustration()
+      .invoke('outerHeight')
+      .should('be.eq', 250);
```
AndreyYevtukhov: I'm not sure that CSS checks are needed in e2e (these will be covered by UI tests which we will add later)

apps/fulfillment-e2e/src/support/page_fragments/picking-lists.fragment.ts
```diff
@@ -1,5 +1,14 @@
 export class PickingListsFragment {
+  private getElement = (selector: string, withoutWrapper?: boolean) => {
+    return withoutWrapper ? cy.get(selector) : this.getWrapper().find(selector);
+  };
```
AndreyYevtukhov: This function looks strange. Do we really need it? Or maybe it can be refactored? (or maybe UI can be refactored so that we don't have to do this)

AndreyYevtukhov: The goal is to have the simplest page object (and tests) possible

Each additional method or function adds complexity and increases the costs of maintaining tests (which we do not want, right?)

AndreyYevtukhov (changes requested):
Overall - it is good, page object and fragment (as well as selectors) are fine.

The only major issue I see is the number of tests (these are not unit tests + we are paying for each `it` block :)) and the E2E checks.

---

E2E tests should be focused on functionality of the app, not the small UI checks.

That's why: routing, navigation, clicks on elements, blocks/text visibility checks, BE data displaying checks (e.g. Order ID is displayed in the right place) etc. are OK, but CSS classes, attributes, element sizes, colors, etc. should not be here as they should be covered on unit level (if you want) or in UI regression tests

AndreyYevtukhov (approved):
Nice job!

PR #281, opened 2023-04-19
feat: picking all products e2e test

nsilva-spryker:
Add new e2e integration to test full picking.
Add new data test id attribute to handle some elements, without relying on their CSS classes 

Closes [HRZ-2715](https://spryker.atlassian.net/browse/HRZ-2715)

[HRZ-2715]: https://spryker.atlassian.net/browse/HRZ-2715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/fulfillment-e2e/src/support/commands.ts
```diff
@@ -4,6 +4,7 @@ declare global {
   namespace Cypress {
     interface Chainable {
       clearIndexedDB(): void;
+      findByTestId(selector: string): Chainable<JQuery<HTMLElement>>;
```
AndreyYevtukhov: Do we need the `getByTestId` command as well? (or we can add it later?)

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -248,7 +248,7 @@ export class PickingComponent extends PickingListMixin(LitElement) {
     }
 
     return html`
-      <div class="picking-complete">
+      <div class="picking-complete" data-testid="picking-complete-placeholder">
```
AndreyYevtukhov: I'm not sure about these, because @tobi-or-not-tobi don't want them to appear in production code (and I agree that we should avoid it) :) 

Let's remove them now (together with that cypress `findById` action) and have a discussion about selectors with everyone to come up with the final solution.

libs/domain/picking/src/components/picking/picking.component.ts
```diff
@@ -269,6 +269,7 @@ export class PickingComponent extends PickingListMixin(LitElement) {
         class="submit-wrapper${classMap({
           'scroll-shadow': hasShadow ?? false,
         })}"
+        data-testid="picking-complete-finish-button"
```
AndreyYevtukhov: Also to be removed

apps/fulfillment-e2e/src/integration/fully-picking-a-picklist.cy.ts
```diff
@@ -0,0 +1,45 @@
+import { PickingFragment } from '../support/page_fragments/picking.fragment';
+
+describe('Fully pick a picking list', () => {
+  const pickingListId = '37cb241e-f18a-5768-985c-a2d7aff4875e';
```
AndreyYevtukhov, line 4: @tarashaliuk added a nice method to get an ID and not hardcode it in the tests - here https://github.com/spryker/oryx/pull/280 - can we merge that PR in dev (it is approved, so everyone can do that) and use it here?

apps/fulfillment-e2e/src/integration/fully-picking-a-picklist.cy.ts
```diff
@@ -0,0 +1,45 @@
...
+        const oryxCartQuantity = product.find('oryx-cart-quantity-input');
+        const increaseButton = oryxCartQuantity.find('[part="increase"]');
+
+        increaseButton.click({ multiple: true });
+      });
+
+      [pickingFragment.getProducts()].forEach((product) => {
+        const doneButton = product
+          .findByTestId('product-card-done')
+          .should('exist');
```
AndreyYevtukhov: Shouldn't these be in the page object/page fragment? 

we should avoid adding selectors in test files, because in the future when some selectors or update HTML - it might cause huge pain :)

apps/fulfillment-e2e/src/integration/fully-picking-a-picklist.cy.ts
```diff
@@ -0,0 +1,45 @@
...
+
+        doneButton.click({ multiple: true });
+      });
+    });
+
+    it('must have all products on picked tab after done all of them', () => {
```
AndreyYevtukhov, line 30: I'm missing `Verify that the “Great job” illustration is shown and the “Finish picking” button appears` check here (we have it in ticket description)

AndreyYevtukhov: or is it `getAllItemsDonePlaceholder`?

AndreyYevtukhov (changes requested):
Left some small comments, but it looks nice! 

And that split to 2 tests (instead of 1) makes sense, nice catch!

apps/fulfillment-e2e/src/integration/fully-picking-a-picklist.cy.ts
nsilva-spryker: Currently we can't take any id on the picking page(like theres is on picking lists items), because we still don't have the header information. 
As soon we have that, we use the id information on the header and not been needed to mock it on the e2e test

libs/domain/picking/src/components/picking/picking.component.ts
nsilva-spryker: Yup, let's remove it for now and discuss this later, I see the point of avoiding this.

apps/fulfillment-e2e/src/support/commands.ts
nsilva-spryker: Let's remove the `findByTestId` for now

apps/fulfillment-e2e/src/integration/fully-picking-a-picklist.cy.ts
```diff
@@ -18,9 +18,7 @@ describe('Fully pick a picking list', () => {
       });
 
       [pickingFragment.getProducts()].forEach((product) => {
-        const doneButton = product
-          .findByTestId('product-card-done')
-          .should('exist');
+        const doneButton = product.find('oryx-button').should('exist');
```
AndreyYevtukhov: Do we need the `.should('exist')` check here?

It seems like this can be replaced with `product.find('oryx-button').click({ multiple: true });` 

---

Also, `.find('oryx-button')` should be moved in page object / page fragment :)

apps/fulfillment-e2e/src/integration/fully-picking-a-picklist.cy.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  beforeEach(() => {
+    cy.visit(`/picking-list/picking/${pickingListId}`);
+  });
+
+  describe('Picking all product items', () => {
+    beforeEach(() => {
+      [pickingFragment.getProducts()].forEach((product) => {
+        const oryxCartQuantity = product.find('oryx-cart-quantity-input');
+        const increaseButton = oryxCartQuantity.find('[part="increase"]');
```
AndreyYevtukhov: `find()` calls should be moved in page object / page fragment (we don't want to have selectors in test files)

AndreyYevtukhov (changes requested):
Almost there, only 2 similar changes left

apps/fulfillment-e2e/src/integration/fully-picking-a-picklist.cy.ts
nsilva-spryker: Yup, we need the `.should('exist')` because some product cards could be in edit mode and not with 'done button' visible

AndreyYevtukhov (approved):

PR #282, opened 2023-04-19
feat: improve media queries generator

supproduction:
adds include\exclude generation
improves medias generation

closes [HRZ-2718](https://spryker.atlassian.net/browse/HRZ-2718)

[HRZ-2718]: https://spryker.atlassian.net/browse/HRZ-2718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/core/src/orchestration/theme/theme-tokens.ts
```diff
@@ -56,28 +56,76 @@ export class ThemeTokens {
...
+    const mediaRule = getPropByPath(this.mediaMapper, value);
+
+    if (isScreen) {
+      return this.generateScreenMedia(path[1] ?? value);
+    }
+
+    return `@media (${mediaRule})`;
+  }
+
+  generateScreenMedia(
```
tobi-or-not-tobi, line 71: I'd prefer to move this logic to the breakpoint service and use the plugin only to collect the breakpoint data

libs/platform/core/src/orchestration/theme/theme-tokens.ts
```diff
@@ -56,28 +56,76 @@ export class ThemeTokens {
...
+    include: string | string[],
```
tobi-or-not-tobi, line 72: Can we support `generateScreenMedia([],[])`? It should return `undefined` or `null`.

libs/platform/core/src/orchestration/theme/theme.model.ts
```diff
@@ -25,6 +25,8 @@ export interface ThemeBreakpoints {
   [Size.Lg]?: ScreenSize;
 }
 
+export type Breakpoint = keyof ThemeBreakpoints;
+
 export interface ThemeMediaQueries {
   [ThemeDefaultMedia.Mode]?: 'dark' | 'light';
   [ThemeDefaultMedia.Screen]?: keyof ThemeBreakpoints;
```
tobi-or-not-tobi:
```suggestion
  [ThemeDefaultMedia.Screen]?: Breakpoint;
```

libs/platform/core/src/orchestration/theme/theme-tokens.ts
tobi-or-not-tobi: we could still create a factory for it if we do not like to bloat the service code.

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #284, opened 2023-04-20
chore: checkout component refactor

tobi-or-not-tobi:
The checkout has been refactored, but also comes with some functional changes:
- The orchestration of the steps is done by the `CheckoutOrchestratorComponent`, using an inner composition. This allows more flexibility in the UX to rearrange steps and come up with alternative checkout flows (e.g. multi-step).
- The guest checkout is still optional (configurable through the customer checkout component), but is enabled by default; the user is not routed to an auth/login page. Instead, when not authenticated, the user lands at the checkout page and can switch to login at all times.

Technical refactor includes:
- The `CheckoutOrchestratorService` is removed;
- The checkout state is stored in the `CheckoutStateService` 
- The `CheckoutDataService` is only used to load the checkout options (i.e. shipments, payments, etc.). The specific `CheckoutShipmentService` and `CheckoutPaymentService` are removed.
- The components are simplified using signals and effects
- The validation of the checkout form data is triggered from the `CheckoutOrchestratorComponent`
- Stores have been completed for customer and link component.

closed [HRZ-2589](https://spryker.atlassian.net/browse/HRZ-2589)

[HRZ-2589]: https://spryker.atlassian.net/browse/HRZ-2589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/checkout/composition/src/composition.component.ts
```diff
@@ -34,6 +42,25 @@ export class CheckoutCompositionComponent extends ComponentMixin<CheckoutComposi
...
+
+  protected isGuest = signal(this.checkoutDataService.isGuestCheckout());
+  protected isAuthenticated = signal(this.authService.isAuthenticated());
+
+  protected checkoutLoginRoute = signal(
+    resolve(SemanticLinkService).get({ type: SemanticLinkType.CheckoutLogin })
+  );
+
+  protected eff = effect(() => {
+    if (!this.isAuthenticated() && !this.isGuest()) {
```
dunqan: you should stop effect at component disconnection.... we may think about a new directive to automate it. In general, this logic should probably be moved to connectedCallback also.

libs/domain/checkout/guest/guest.component.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  protected isGuest = signal(this.checkoutDataService.isGuestCheckout());
+  protected isAuthenticated = signal(this.authService.isAuthenticated());
+
+  protected checkoutRoute = signal(
+    this.linkService.get({ type: SemanticLinkType.Checkout })
+  );
+
+  protected eff = effect(() => {
+    if (this.isGuest() || this.isAuthenticated()) this.redirect();
```
dunqan: this one will also be running long time component will be disconnected, and it could cause some unexpected sideeffects.

libs/domain/checkout/src/mixins/checkout.mixin.ts
```diff
@@ -1,14 +1,21 @@
...
-  ContentComponentProperties,
-} from '@spryker-oryx/experience';
+import { signalAware } from '@spryker-oryx/utilities';
 import { LitElement } from 'lit';
 
-export const CheckoutComponentMixin = <T>(): Type<
-  LitElement & ContentComponentProperties<T>
-> => {
-  class CheckoutComponent extends ComponentMixin<T>() {}
+// export const CheckoutComponentMixin = <T>(): Type<
```
dunqan: 👀

libs/domain/checkout/composition/src/composition.component.ts
```diff
@@ -1,127 +1,41 @@
...
+
+  protected requiresRedirect = computed(
+    () =>
+      this.$options().checkoutAuthPage &&
+      !this.isGuest() &&
+      !this.isAuthenticated()
+  );
+
+  connectedCallback(): void {
+    if (this.requiresRedirect()) {
```
dunqan: I guess we should do it not only on connected callback? Or you think otherwise?

libs/domain/checkout/src/mocks/src/mock-checkout.providers.ts
```diff
@@ -23,14 +70,18 @@ export const mockCheckoutProviders: Provider[] = [
...
     provide: CheckoutOrchestrationService,
-    useClass: DefaultCheckoutOrchestrationService,
+    useClass: MockCheckoutOrchestrationService,
   },
   {
     provide: CheckoutService,
     useClass: MockCheckoutService,
   },
+  {
+    provide: CartService,
```
dunqan: this one may be problematic, when you'll use both cart and checkout mocks, as one can override CartService from the other. I'd consider only mocking services that are part of the domain.

libs/domain/checkout/src/services/default-checkout.service.ts
```diff
@@ -57,14 +56,7 @@ export class DefaultCheckoutService implements CheckoutService {
...
         switchMap((canCheckout) =>
-          canCheckout
-            ? this.preparePayload()
-            : throwError(
-                () =>
-                  new Error(
-                    'Cannot checkout, check your cart value limits and validations'
-                  )
-              )
+          canCheckout ? this.preparePayload() : of()
```
dunqan: not sure if we should fail silently? 
Throwing an error gives ability to decide in the component, what to do with it. 

But I agree, that the error message here is too specific.

libs/domain/checkout/composition/src/composition.component.ts
tobi-or-not-tobi: I had it first on an effect, but it was too aggressive, as it would kick in at:
- after placing the order
- when you use the ordinary logout
in both cases we'd land at checkout-login

libs/domain/checkout/src/services/default-checkout.service.ts
tobi-or-not-tobi: I noticed the error is thrown when expected; when some of the forms are not valid, we have form validation and this error seems too much.

libs/domain/checkout/delivery/src/delivery.component.ts
```diff
@@ -1,72 +1,67 @@
...
-        this.isAuthenticated$,
-        (isAuthenticated) =>
-          html` ${when(
-            !isAuthenticated,
-            () => html`<checkout-contact></checkout-contact>`
-          )}`
-      )}
-      <checkout-address></checkout-address>
-    `;
+  protected get forms(): NodeListOf<CheckoutForm> {
```
Sinucid: Can we use decorator instead?
```
@queryAll('oryx-checkout-contact, oryx-checkout-address')
protected forms: NodeListOf<CheckoutForm>
```

libs/domain/checkout/guest/guest.component.ts
```diff
@@ -0,0 +1,63 @@
...
+import { defaultOptions } from '@spryker-oryx/experience';
+import { RouterService } from '@spryker-oryx/router';
+import { SemanticLinkService, SemanticLinkType } from '@spryker-oryx/site';
+import { Size } from '@spryker-oryx/ui';
+import { hydratable, i18n, signal } from '@spryker-oryx/utilities';
+import { html, LitElement, TemplateResult } from 'lit';
+import { styles } from './guest.styles';
+
+@defaultOptions({
+  enableGuestCheckoutByDefault: true,
```
Sinucid: I don't see using of this options

libs/domain/checkout/guest/guest.component.ts
```diff
@@ -0,0 +1,63 @@
...
+})
+@hydratable('window:load')
+export class CheckoutGuestComponent extends CheckoutMixin(LitElement) {
+  static styles = [styles];
+
+  protected linkService = resolve(SemanticLinkService);
+
+  protected checkout = signal(
```
Sinucid: `checkoutLink` ?

libs/domain/checkout/guest/guest.component.ts
```diff
@@ -0,0 +1,63 @@
...
+    this.linkService.get({ type: SemanticLinkType.Checkout })
+  );
+
+  protected checkoutLogin = signal(
```
Sinucid: `checkoutLoginLink` ?

libs/domain/checkout/guest/guest.component.ts
```diff
@@ -0,0 +1,63 @@
...
+    this.linkService.get({ type: SemanticLinkType.CheckoutLogin })
+  );
+
+  connectedCallback(): void {
```
Sinucid: Can cause uncontrolled behaviour when called earlier then signals are resolved. I believe Stan suggested to use `effect` instead

libs/domain/checkout/guest/guest.component.ts
```diff
@@ -0,0 +1,63 @@
...
+    return html`<h1>${i18n('checkout.guest.guest-checkout')}</h1>
+      <p>${i18n('checkout.guest.continue-without-account')}</p>
+      <oryx-button .size=${Size.Md}>
+        <a href=${this.checkout()} @click=${() => (this.guestCheckout = true)}>
+          ${i18n('checkout.guest.open-checkout')}
+        </a>
+      </oryx-button>`;
+  }
+
+  protected set guestCheckout(value: boolean) {
```
Sinucid: New convention? Looks not very convenient

libs/domain/checkout/guest/guest.component.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  protected override render(): TemplateResult | void {
+    if (this.isAuthenticated()) return;
+
+    if (this.isGuest()) {
+      return html`<p class="have-an-account">${i18n(
+        'checkout.guest.have-an-account?'
+      )}</p>
+        <oryx-button outline size="md">
```
Sinucid: `size=${Size.Md}`

libs/domain/checkout/payment/src/payment.component.ts
```diff
@@ -118,6 +57,27 @@ export class CheckoutPaymentComponent extends ComponentMixin() {
...
 
+  // TODO: consider fallback strategies: none, first, cheapest
+  protected isSelected(methodId: string): boolean {
+    return this.selected()
+      ? this.selected() === methodId
+      : methodId === this.methods()?.[0]?.id;
+  }
+
+  protected validate(action: CheckoutTrigger): void {
+    const selected = this.renderRoot.querySelector('input[checked]');
```
Sinucid: `@query`?

libs/domain/checkout/place-order/src/place-order.component.ts
```diff
@@ -1,58 +1,31 @@
...
-import { BehaviorSubject, combineLatest, map } from 'rxjs';
+import { CheckoutMixin } from '@spryker-oryx/checkout';
+import { hydratable, i18n, signal } from '@spryker-oryx/utilities';
+import { html, LitElement, TemplateResult } from 'lit';
 
 @hydratable('window:load')
-export class CheckoutPlaceOrderComponent extends ComponentMixin() {
-  protected checkout = resolve(CheckoutService);
+export class CheckoutPlaceOrderComponent extends CheckoutMixin(LitElement) {
+  protected isBusy = signal(false);
```
Sinucid: Would this loading state not be a part of the service?

libs/domain/checkout/shipment/src/shipment.component.ts
```diff
@@ -1,95 +1,70 @@
...
+      )}`;
   }
 
-  protected renderMethod(
-    method: ShipmentMethod,
-    selected = false
-  ): TemplateResult {
-    return html`<oryx-tile ?selected="${selected}">
+  protected renderMethod(method: ShipmentMethod): TemplateResult {
+    const isSelected = this.isSelected(Number(method.id));
```
Sinucid: Why needs to convert the value to the Number?

libs/domain/checkout/composition/src/composition.component.ts
tobi-or-not-tobi: outdated

libs/domain/checkout/guest/guest.component.ts
tobi-or-not-tobi: outdated

libs/domain/checkout/src/services/default-checkout.service.ts
Sinucid: ` canCheckout ? this.preparePayload() : of(undefined)`

Sinucid (changes requested):

libs/domain/checkout/src/mocks/src/mock-checkout.providers.ts
tobi-or-not-tobi: Also dropped it from checkout mixin

libs/domain/checkout/guest/guest.component.ts
tobi-or-not-tobi: I replied to that suggestion (and we had effect before, but it's not working in this case)

libs/domain/checkout/guest/guest.component.ts
tobi-or-not-tobi: not sure why, but we have not made it a standard pattern, so i'll go back to 2 methods

Sinucid: I meant that this is not fully obvious to use setter in this way

libs/domain/checkout/shipment/src/shipment.component.ts
tobi-or-not-tobi: I am not entirely sure, but it's not been part of this PR. I've investigated a little as I wanted to make it string-only, but like to leave that for another time.

libs/domain/checkout/guest/guest.component.ts
tobi-or-not-tobi: discussed once more with Stan and we'll improve this after signal _decorator_ lands.

.gitignore
```diff
@@ -51,3 +51,6 @@ build-storybook.log
 # Ignoring env configuration
 **/.env*
 !**/.env.template
+
+
+/apps/storefront-e2e/cypress/downloads
```
supproduction: do we really need it?

libs/domain/checkout/address/address.component.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+      describe('and the form is invalid', () => {
+        beforeEach(() => {
+          const form = document.createElement('form');
+          form.innerHTML = '<input type="email" name="email" required>';
+          const addressForm = element.shadowRoot?.querySelector(
+            'oryx-address-form'
+          ) as AddressFormComponent;
+          addressForm.getForm = () => form;
+        });
+        it('should not return the form data', () => {
```
supproduction:
```suggestion
        });
        
        it('should return the form data', () => {
          expect(element.collectData()).toEqual({ email: 'foo@bar.com' });
        });
      });

      describe('and the form is invalid', () => {
        beforeEach(() => {
          const form = document.createElement('form');
          form.innerHTML = '<input type="email" name="email" required>';
          const addressForm = element.shadowRoot?.querySelector(
            'oryx-address-form'
          ) as AddressFormComponent;
          addressForm.getForm = () => form;
        });
        
        it('should not return the form data', () => {
```

Please, keep format

libs/domain/checkout/address/address.component.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  collectData(): Address | null {
+    const form = this.addressComponent?.getForm();
+    if (!this.selected && !form?.checkValidity()) {
+      form?.reportValidity();
+      return null;
+    }
+    return form
+      ? (Object.fromEntries(new FormData(form).entries()) as unknown as Address)
+      : (this.selected as Address | null);
```
supproduction:
```suggestion
      : this.selected;
```

libs/domain/checkout/auth/auth.component.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+        element = await fixture(
+          html`<oryx-checkout-auth></oryx-checkout-auth>`
+        );
+        guestComponent = element.shadowRoot?.querySelector(
+          'oryx-checkout-guest'
+        ) as CheckoutGuestComponent;
+        guestComponent.collectData = vi.fn();
+        callback().pipe(take(1)).subscribe();
+      });
+      it('should collect the data', () => {
```
supproduction:
```suggestion
      });
      
      it('should collect the data', () => {
        expect(userService.getUser).toHaveBeenCalled();
      });
    });
  });

  describe('when the user is not authenticated', () => {
    beforeEach(async () => {
      authService.isAuthenticated.mockReturnValue(of(false));
      userService.getUser.mockReturnValue(of(null));
      element = await fixture(html`<oryx-checkout-auth></oryx-checkout-auth>`);
    });

    it('should render the guest component', () => {
      expect(element).toContainElement('oryx-checkout-guest');
    });

    describe('and the collect callback is called', () => {
      let guestComponent: CheckoutGuestComponent;
      
      beforeEach(async () => {
        element = await fixture(
          html`<oryx-checkout-auth></oryx-checkout-auth>`
        );
        guestComponent = element.shadowRoot?.querySelector(
          'oryx-checkout-guest'
        ) as CheckoutGuestComponent;
        guestComponent.collectData = vi.fn();
        callback().pipe(take(1)).subscribe();
      });
      
      it('should collect the data', () => {
```

libs/domain/checkout/auth/auth.model.ts
supproduction, line 1: Do we need empty file?

libs/domain/checkout/auth/index.ts
```diff
@@ -1 +1,3 @@
-export * from './src';
+export * from './auth.component';
+export * from './auth.schema';
+export * from './auth.styles';
```
supproduction: We need to keep `src` folder for paths generation (if we still need this tool ofc)

libs/domain/checkout/delivery/delivery.component.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    it('should register the step at the checkout service', () => {
+      expect(checkoutService.register).toHaveBeenCalledWith(
+        'shippingAddress',
+        expect.anything(),
+        2
+      );
+    });
+  });
+
+  describe('if there is no list of addresses', () => {
```
supproduction, line 116: Here and 2 describes below, please fix format

libs/domain/checkout/orchestrator/orchestrator.component.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  describe('when the collect callback is called', () => {
+    beforeEach(async () => {
+      element = await fixture(
+        html`<oryx-checkout-orchestrator></oryx-checkout-orchestrator>`
+      );
+      callback().pipe(take(1)).subscribe();
+    });
+    it('should collect the address', () => {
```
supproduction:
```suggestion
    });
    
    it('should collect the address', () => {
```

libs/domain/checkout/orchestrator/stories/static/behavior.stories.ts
```diff
@@ -15,7 +13,7 @@ export default {
 
 const Template: Story<Props> = (props): TemplateResult => {
   toggleBehavior(props.behavior);
-  resolve(CheckoutDataService).setGuestCheckout(props.behavior === 'guest');
+  // resolve(CheckoutDataService).setGuestCheckout(props.behavior === 'guest');
```
supproduction: Leftover?

libs/domain/checkout/address/address.component.ts
Sinucid: Do we need type casting here? 
`: this.selected;`

libs/domain/checkout/place-order/place-order.component.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  protected onClick(): void {
+    this.checkoutService.placeOrder().subscribe({
+      next: (response) => {
+        this.redirect(response.redirectUrl, response.isExternalRedirect);
+      },
+      error: (error) => {
+        throw error;
+      },
```
supproduction: Without it we don't throw error?

libs/domain/checkout/auth/auth.model.ts
Sinucid, line 1: empty file

libs/domain/checkout/src/models/orchestration.model.ts
```diff
@@ -15,14 +15,14 @@ export enum CheckoutStepType {
   Payment = 'payment',
 }
 
-export interface CheckoutStep {
-  id: CheckoutStepType;
-  label?: string;
-}
+// export interface CheckoutStep {
```
supproduction: Leftover?

libs/domain/checkout/guest/guest.component.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  protected loginRoute = signal(
+    this.linkService.get({ type: SemanticLinkType.Login })
+  );
+
+  @query('form')
+  protected form?: HTMLFormElement;
+
+  protected override render(): TemplateResult | void {
+    const guestFields: FormFieldDefinition[] = [
```
Sinucid: It will re-define this object on each re-render. Move it to the model or separate property of the class

libs/domain/checkout/src/services/default-checkout.service.ts
```diff
@@ -1,140 +1,172 @@
...
-  protected postCheckout(response: CheckoutResponse): Observable<unknown> {
+  /**
+   * Clears the cart, stores the order in memory.
+   */
+  protected postCheckout(response: CheckoutResponse): void {
+    this.cartService.reload();
     if (response.orders) {
-      this.orderService.storeLastOrder(response.orders[0]);
+      console.log('TODO: store last order', response.orders[0]);
+      // this.orderService.storeLastOrder(response.orders[0]);
```
supproduction: Leftover?

libs/domain/checkout/src/services/payment/default-checkout-payment.service.ts
```diff
@@ -0,0 +1,84 @@
...
+    map((data) => data?.paymentMethods ?? null),
+    shareReplay({ bufferSize: 1, refCount: true })
+  );
+
+  getMethods(): Observable<PaymentMethod[] | null> {
+    return this.methods$;
+  }
+
+  select(key: string): Observable<unknown> {
+    // this.subject.next(key);
```
supproduction: leftover?

supproduction (changes requested):

libs/domain/checkout/manage-address/manage-address.component.ts
```diff
@@ -0,0 +1,111 @@
...
+  protected onRemove(ev: CustomEvent): void {
+    this.selected = ev.detail.address.id;
+  }
+
+  protected onStateChange(e: CustomEvent): void {
+    this.state = e.detail.state;
+  }
+
+  protected renderRemoveConfirmationModal(): TemplateResult | void {
+    if (!this.selected) return;
```
Sinucid, line 83: not sure whether this is a right way or not. It will mount and unmount the modal and it's content each time. Less performance and what is the benefit?

libs/domain/checkout/src/mocks/src/mock-checkout.providers.ts
```diff
@@ -1,33 +1,50 @@
...
   CheckoutShipmentService,
-  DefaultCheckoutDataService,
-  DefaultCheckoutOrchestrationService,
 } from '@spryker-oryx/checkout';
 import { Provider } from '@spryker-oryx/di';
-import { MockCheckoutService } from './mock-checkout.service';
+import { RouterService } from '@spryker-oryx/router';
+import { SemanticLinkService } from '@spryker-oryx/site';
+import { of } from 'rxjs';
+import { vi } from 'vitest';
```
Sinucid: not needed

libs/domain/checkout/src/mixins/checkout.mixin.ts
```diff
@@ -1,14 +1,44 @@
...
+  superClass: T
+): Type<CheckoutMixinInterface> & T => {
+  @signalAware()
+  class CheckoutMixinClass extends superClass {
+    protected checkoutService = resolve(CheckoutService);
+
+    protected isAvailable = signal(
+      this.checkoutService
+        .getProcessState()
+        .pipe(map((state) => state !== CheckoutProcessState.NotAvailable)),
```
Sinucid: `.pipe(map((state) => state !== CheckoutProcessState.Ready)),` ?

libs/domain/checkout/src/models/orchestration.model.ts
```diff
@@ -15,14 +15,14 @@ export enum CheckoutStepType {
...
+//   label?: string;
+// }
 
 export type ValidityReport = {
   id: CheckoutStepType;
   validity: Validity | undefined;
 };
 
-export type CheckoutConfiguration = CheckoutStep[];
+// export type CheckoutConfiguration = CheckoutStep[];
```
Sinucid: Still relevant?

libs/domain/checkout/src/services/adapter/normalizers/payments/payments.normalizer.spec.ts
```diff
@@ -14,9 +14,19 @@ const mockCheckoutData: DeserializedCheckout = {
...
 
-describe('Payments Normalizer', () => {
-  it('should transform ApiCheckoutModel.PaymentMethods Include into CheckoutData', () => {
-    const normalized = paymentsNormalizer(mockCheckoutData);
-    expect(normalized).toEqual(mockNormalizedPaymentMethods);
+describe('paymentsNormalizer', () => {
+  describe('when no data is provided', () => {
+    it('should return an empty array', () => {
+      expect(
+        paymentsNormalizer(null as unknown as DeserializedCheckout)
```
Sinucid: I believe normalizer should accept 'no data' without type casting

libs/domain/checkout/src/services/default-checkout.service.ts
```diff
@@ -1,140 +1,172 @@
...
 
-export class DefaultCheckoutService implements CheckoutService {
+// consider moving out steps, register and collect to CheckoutOrchestratorService
+// we'd keep register here as a proxy to "hide" the orchestrator from the components
+export class DefaultCheckoutService<T extends Checkout>
+  implements CheckoutService<T>
+{
+  protected state = new BehaviorSubject(CheckoutProcessState.Initializing);
+
+  protected steps: Array<{
```
Sinucid: Looks like a type for a model

libs/domain/checkout/src/services/default-checkout.service.ts
```diff
@@ -1,140 +1,172 @@
...
+    protected linkService = inject(SemanticLinkService)
   ) {}
 
-  protected canCheckout$ = this.orchestrationService.getValidity().pipe(
-    map((validity) =>
-      validity.every(({ validity }) => validity === Validity.Valid)
-    ),
-    shareReplay({ bufferSize: 1, refCount: true })
-  );
+  register<K extends keyof T>(
```
Sinucid: similar

libs/domain/checkout/src/services/default-checkout.service.ts
```diff
@@ -1,140 +1,172 @@
...
-            ...customer!,
-            salutation:
-              customer?.salutation ?? billingAddress!.salutation ?? '',
-          },
-          billingAddress: {
-            ...billingAddress!,
-          },
-        };
+  protected collect(): Observable<T> {
+    if (this.steps.length === 0) {
```
Sinucid: `if (!this.steps.length) {`

libs/domain/checkout/src/services/default-checkout.service.ts
```diff
@@ -1,140 +1,172 @@
...
+      throw new Error('No steps registered');
+    }
 
-        if (shipment?.idShipmentMethod) {
-          payload.shipment = { idShipmentMethod: shipment.idShipmentMethod };
-          payload.shippingAddress = { ...billingAddress! };
-        }
+    const collectStep = (
```
Sinucid: This recursion looks weird

Sinucid (changes requested):

libs/domain/checkout/auth/auth.model.ts
tobi-or-not-tobi: I kept it, but filled it with new option to enable guest checkout

libs/domain/checkout/auth/index.ts
tobi-or-not-tobi: works like a charm actually

libs/domain/checkout/manage-address/manage-address.component.ts
tobi-or-not-tobi: removing an address (very) rarely happens, which means that we have a cleaner DOM in most cases.

libs/domain/checkout/src/mocks/src/mock-checkout.providers.ts
tobi-or-not-tobi: we have TS issues if we drop it

libs/domain/checkout/src/mixins/checkout.mixin.ts
tobi-or-not-tobi: A busy checkout is still available.

libs/domain/checkout/src/models/orchestration.model.ts
tobi-or-not-tobi: nope dropped all types there already

tobi-or-not-tobi: nope dropped all types there already

libs/domain/checkout/src/services/default-checkout.service.ts
tobi-or-not-tobi: Have left a comment in the code to clarify.

dunqan: I agree it's complex, I'd try to replace it with either for loop or `concat` + `filter`.

On the other hand, we will still need to observe all steps, as they may become invalid at unpredictable time.

dunqan: We also want to trigger validity checks for all components on single page checkout (to mark all invalid fields in all steps, not only half of it)

tobi-or-not-tobi: No we do not, we like to only mark invalid fields step by step on single page as well, as otherwise the errors look to overwhelming. It might be opinionated and we could add an option to validate it all.

libs/domain/checkout/manage-address/manage-address.component.ts
Sinucid: Potentially can cause performance issue, but lets see how it will be in result

libs/domain/checkout/place-order/place-order.component.ts
```diff
@@ -0,0 +1,32 @@
...
+    </oryx-button>`;
+  }
+
+  protected onClick(): void {
+    this.checkoutService.placeOrder().subscribe((response) => {
+      this.redirect(response.redirectUrl, response.isExternalRedirect);
+    });
+  }
+
+  protected redirect(url?: string, external = false): void {
```
supproduction: Why do we need second params? Also not sure that this is ok to have functions with optional param. Mb better change url to required prop and add condition when we call this function?

libs/domain/checkout/src/mocks/src/mock-checkout.providers.ts
Sinucid: Wow, then wait a sec. Why do we need to use `vitest` for mocking data? Just:

```
export class MockCheckoutService implements Partial<CheckoutService> {
  register = () => undefined;
  getProcessState = () => of(CheckoutProcessState.Initializing);
  placeOrder = () => of();
}
```

and etc...

libs/domain/checkout/src/services/default-checkout.service.ts
```diff
@@ -1,140 +1,170 @@
...
-                  new Error(
-                    'Cannot checkout, check your cart value limits and validations'
-                  )
-              )
-        ),
-        switchMap((payload) =>
-          this.adapter.placeOrder({ attributes: payload })
-        ),
+      this.collect().pipe(
+        catchError((error) => {
```
supproduction: Please, delete `error` as we don't use it

supproduction (changes requested):

libs/domain/checkout/src/mocks/src/mock-checkout.providers.ts
tobi-or-not-tobi: yeah my bad, will clean it up as it doesn't belong here

supproduction (dismissed):

libs/base/ui/form/input/src/error/error.controller.ts
```diff
@@ -3,16 +3,17 @@ import { ErrorOptions } from './error.model';
 
 export class ErrorController implements ReactiveController {
   hostUpdated?(): void {
-    this.handleError();
+    // this.handleError();
```
dunqan: is it deliberate?

libs/base/ui/form/input/src/error/error.controller.ts
```diff
@@ -3,16 +3,17 @@ import { ErrorOptions } from './error.model';
...
 
   render(): TemplateResult {
     return html`<oryx-error-message .message=${this.message}>
-      <slot name="error" @slotchange=${this.handleError.bind(this)}></slot>
+      <slot name="error"></slot>
     </oryx-error-message>`;
   }
 
   protected handleError(): void {
+    console.log('handle error', this);
```
dunqan: 👀

libs/domain/checkout/src/services/default-checkout.service.ts
```diff
@@ -1,140 +1,104 @@
...
-          this.adapter.placeOrder({ attributes: payload })
-        ),
-        tap((response) => this.postCheckout(response))
+                ),
+
+                finalize(() => this.process.next(CheckoutState.Ready))
+              );
+          } else {
+            this.process.next(CheckoutState.Invalid);
+            return EMPTY;
```
dunqan: let's error the stream on fail with returning `throwError` here?

libs/domain/checkout/delivery/delivery.component.ts
```diff
@@ -0,0 +1,49 @@
...
+
+@hydratable()
+export class CheckoutDeliveryComponent
+  extends CheckoutMixin(LitElement)
+  implements isValid
+{
+  static styles = [styles];
+
+  protected selected = signal(this.checkoutStateService.get('shippingAddress'));
+  protected addresses = signal(resolve(AddressService).getAddresses());
```
dunqan: We discussed convention of not resolving inside signal factory

libs/domain/checkout/src/services/state/default-checkout-state.service.ts
```diff
@@ -0,0 +1,127 @@
...
+            const map = new Map<
+              keyof Checkout,
+              { data: Checkout[keyof Checkout] | null; valid: boolean }
+            >();
+            Object.keys(stored).forEach(() => {
+              const entries = Object.entries(stored);
+              for (const [key, value] of entries) {
+                map.set(key as keyof Checkout, {
+                  data: value as Checkout[keyof Checkout] | null,
+                  valid: false,
```
dunqan: We discussed that this one may not be needed and could simplify components.

libs/base/ui/form/input/src/error/error.controller.ts
tobi-or-not-tobi: NO!

Sinucid (dismissed):

libs/domain/checkout/src/services/state/default-checkout-state.service.ts
tobi-or-not-tobi: yes, this has improved the code a lot

libs/domain/checkout/guest/guest.component.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    it('should be an instance of CheckoutGuestComponent', () => {
+      expect(element).toBeInstanceOf(CheckoutGuestComponent);
+    });
+
+    it('should pass the a11y audit', async () => {
+      await expect(element).shadowDom.to.be.accessible();
+    });
+  });
+
+  // describe('when the form is invalid', () => {
```
dunqan: what about adding TODO comment, assuming we want to merge it, and fix in the followup?

libs/domain/checkout/src/models/checkout.model.ts
```diff
@@ -1,4 +1,36 @@
...
+  Invalid = 'invalid',
+}
+
+export interface CheckoutStepCallback<T> {
+  id: keyof T;
+  collectDataCallback: () => Observable<T[keyof T]>;
+  order?: number;
+}
+
+export interface CheckoutTrigger<T> {
```
dunqan: Do we need it?

libs/domain/checkout/src/models/checkout.model.ts
```diff
@@ -1,4 +1,36 @@
...
+import { Observable, Subject } from 'rxjs';
+
+export const enum CheckoutState {
+  Empty = 'empty',
+  Ready = 'ready',
+  Busy = 'busy',
+  Invalid = 'invalid',
+}
+
+export interface CheckoutStepCallback<T> {
```
dunqan: Do we still need it?

libs/domain/product/title/src/title.component.ts
```diff
@@ -24,7 +24,7 @@ export class ProductTitleComponent extends ProductMixin(
   protected override render(): TemplateResult | void {
     const { tag, as, asLg, asMd, asSm, maxLines } = this.$options();
 
-    return html`<oryx-heading
+    return html` <oryx-heading
```
dunqan: 👀

libs/domain/user/address-form/src/address-form.component.ts
```diff
@@ -54,6 +58,23 @@ export class AddressFormComponent
 
   @property({ type: Boolean }) enableDefaultShipping?: boolean;
   @property({ type: Boolean }) enableDefaultBilling?: boolean;
+  @property({ type: Object }) set address(value: Address) {
+    // TODO: introduce generic method to apply form data on form element
```
dunqan: do you want to keep it?

libs/domain/user/address-form/src/address-form.component.ts
```diff
@@ -191,4 +212,29 @@ export class AddressFormComponent
       ${this.fieldRenderer.buildForm(form, this.formValues as FormValues)}
     </form>`;
   }
+
+  @query('form')
+  protected formX?: HTMLFormElement;
```
dunqan, line 201: Would be nice to have a better name than `formX` :)

libs/domain/checkout/guest/guest.component.spec.ts
tobi-or-not-tobi: I'd rather drop all those tests for now, and then work on them this week (in the air is a good time, but we wont' block the merge(

libs/domain/user/address-form/src/address-form.component.ts
tobi-or-not-tobi: Tobias Ouwejan
  [4 minutes ago](https://spryker.slack.com/archives/D02DE33H509/p1684314547130819?thread_ts=1684314538.271339&cid=D02DE33H509)
I'm in favor of dropping it now

Tobias Ouwejan
  [4 minutes ago](https://spryker.slack.com/archives/D02DE33H509/p1684314564651039?thread_ts=1684314538.271339&cid=D02DE33H509)
only when we have refelcction of form data using form component (or something else), we can bring it back

Tobias Ouwejan
  [3 minutes ago](https://spryker.slack.com/archives/D02DE33H509/p1684314591497099?thread_ts=1684314538.271339&cid=D02DE33H509)
it woudl mean that for now we'd not have address form data being restored from memory, which is okish

libs/domain/user/address-form/src/address-form.component.ts
tobi-or-not-tobi: Yeah, but the whole component needs refactoring, so I prefer to leave it for later. We currently have a form property alrady.

libs/domain/checkout/guest/guest.component.spec.ts
dunqan: 👍

libs/domain/checkout/src/services/state/default-checkout-state.service.ts
```diff
@@ -0,0 +1,93 @@
...
+        data.forEach((item, key) => {
+          Object.assign(result, { [key]: item.value });
+        });
+        return this.populateData(result);
+      })
+    );
+  }
+
+  protected populateData(data: Partial<Checkout>): Partial<Checkout> {
+    console.log('inp', data);
```
dunqan: some leftovers

libs/domain/checkout/src/services/state/default-checkout-state.service.ts
```diff
@@ -0,0 +1,93 @@
+import { StorageService, StorageType } from '@spryker-oryx/core';
+import { inject } from '@spryker-oryx/di';
+import { BehaviorSubject, map, Observable, take, tap } from 'rxjs';
+import { Checkout, checkoutDataStorageKey } from '../../models';
+import { CheckoutStateService } from './checkout-state.service';
+
+type m = { valid?: boolean; value?: Partial<Checkout[keyof Checkout]> | null };
```
dunqan: Can we have some more meaningful type name? :)

libs/domain/checkout/src/services/state/default-checkout-state.service.ts
```diff
@@ -0,0 +1,93 @@
...
+      StorageType.SESSION
+    );
+  }
+
+  get<K extends keyof Checkout>(key: K): Observable<Checkout[K] | null> {
+    return this.subject.pipe(
+      map((data) => {
+        if (!data.get(key)) this.set(key, {});
+        return data.get(key)?.value as Checkout[K] | null;
+      })
```
dunqan: let's maybe add `distinctUntilChanged` to improve performance (now all subsribers to data will get emission whenever any data will change)

dunqan (approved):

PR #285, opened 2023-04-20
feat: set placeholder when there is no picking lists

nsilva-spryker:
Closes [HRZ-2736](https://spryker.atlassian.net/browse/HRZ-2736)

[HRZ-2736]: https://spryker.atlassian.net/browse/HRZ-2736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
+++ b/libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
@@ -76,9 +76,7 @@ export class PickingListsComponent extends LitElement {
         <oryx-heading as="h4">
           ${i18n('picking.no-results-found')}
         </oryx-heading>
-        <div class="img-wrap">
-          <oryx-image resource="no-orders"></oryx-image>
-        </div>
+        <oryx-image resource="no-orders"></oryx-image>
       </div>
     `;
   }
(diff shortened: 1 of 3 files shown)


libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
```diff
@@ -71,7 +71,16 @@ export class PickingListsComponent extends LitElement {
   }
 
   protected renderEmptyLists(): TemplateResult {
-    return html`<p>${i18n('picking.no-picking-lists-found')}</p>`;
+    return html`
+      <div class="no-items-fallback">
+        <oryx-heading as="h4">
+          ${i18n('picking.no-results-found')}
+        </oryx-heading>
+        <div class="img-wrap">
```
Sinucid: We can drop this unnecessary `img-wrap` wrapper for `oryx-image` if just improve a bit the styles:

```
.no-items-fallback {
    display: flex;
    flex-direction: column;
    align-items: center;
    height: 100%;
    margin: 45px;
    gap: 16px;
  }

  oryx-image {
    margin-inline: 40px;
    display: inline-block;
  }
```

Sinucid (changes requested):

Sinucid (approved):

PR #286, opened 2023-04-20
feat: index db decorators dual spec

Sinucid:
closes to [HRZ-2726](https://spryker.atlassian.net/browse/HRZ-2726)

[HRZ-2726]: https://spryker.atlassian.net/browse/HRZ-2726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/indexed-db/src/decorators/entity.decorator.ts
```diff
@@ -1,26 +1,49 @@
-import { IndexedDbEntity } from '../models';
+import { ClassContext, IndexedDbEntity, TargetContext } from '../models';
 import {
   IndexedDbIndexMetadata,
   IndexedDbSchemaMetadata,
 } from '../schema-metadata';
 
-export function indexedDbEntity(options?: IndexedDbEntity): ClassDecorator {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function addEntity(context: any, options?: IndexedDbEntity): void {
```
gund: We should not export internal parts of the decorator.

libs/platform/indexed-db/src/models/decorators.model.ts
gund, line 1: These types already exist in [oryx/libs/base/utilities/src/model/decorator.ts](https://github.com/spryker/oryx/blob/ed8f07e513a1e2dadb810a72b9785b3fab5fd375/libs/base/utilities/src/model/decorator.ts) please remove this one or move extra types to that module.

libs/domain/picking/src/components/decorators-test/stories/test.stories.ts
```diff
@@ -0,0 +1,42 @@
...
+    orderReferences: ['test'],
+    productSkus: ['test'],
+    requestedDeliveryDate: new Date(),
+    localStatus: 'test',
+    id: 'test',
+    status: 'test',
+    createdAt: new Date(),
+    updatedAt: new Date(),
+  };
+  const entity = new PickingListEntity(testOfflineData as PickingListOffline);
```
gund: I think it's not enough to just create an entity instance.
We need to validate that the entity metadata is properly collected and for that we should try to store it in the IndexedDb and retrieve it back from it.
You can use `IndexedDbService` for this.

gund (changes requested):

gund (dismissed):

gund (approved):

PR #288, opened 2023-04-21
feat: add ORYX_LABS env support

AndreyYevtukhov:
Closes https://spryker.atlassian.net/browse/HRZ-2778


apps/storefront/src/app.ts
```diff
@@ -1,8 +1,16 @@
 import { appBuilder } from '@spryker-oryx/core';
+import { labsFeature } from '@spryker-oryx/labs';
```
tobi-or-not-tobi: Can you rename it to `labsFeatures` (plural) inside labs package, to make it more consistent?

apps/storefront/.env.template
```diff
@@ -8,3 +8,6 @@ SCOS_BASE_URL=
 # The Cloudinary ID is not on a paid plan, which means that it cannot be heavily used. If you like to copy
 # this implementation for you production usage, make sure that you provide a reliable Cloudinary account.
 ORYX_CLOUDINARY_ID=
+
+# If set - turns on oryx-labs features
```
tobi-or-not-tobi: can you:
- add a warning about this not being a standard supported feature?  We might already have a section in docs that you can copy

tobi-or-not-tobi (changes requested):

apps/storefront/src/app.ts
AndreyYevtukhov: done

apps/storefront/.env.template
AndreyYevtukhov: Done (I've copied it from `libs/template/labs/src/feature.ts` file)

apps/storefront/.env.template
```diff
@@ -3,8 +3,17 @@ SCOS_BASE_URL=
 
 # The STORE env variable is optional and can be used to initialise the application to use a 
 # specific store. When this variable is not provided, the first store from the backend would be used. 
-# STORE=DE
+STORE=DE
```
tobi-or-not-tobi: Why did we make DE hardcoded again? 
- if we do not specify a STORE we take the 1st one, so we should be good
- a hardcoded store will block us from using multi-store

tobi-or-not-tobi (changes requested):

apps/storefront/.env.template
AndreyYevtukhov: this file is just a template, it is not used anywhere

---

I've uncommented it because all other params are not commented (to make it consistent as well)

AndreyYevtukhov: Commented it back)

tobi-or-not-tobi (approved):

PR #296, opened 2023-04-25
chore: expand fulfillment e2e login tests to verify more elements

aleespryker:
Closes [HRZ-2676](https://spryker.atlassian.net/browse/HRZ-2676)

[HRZ-2676]: https://spryker.atlassian.net/browse/HRZ-2676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/fulfillment-e2e/src/integration/login.cy.ts
+++ b/apps/fulfillment-e2e/src/integration/login.cy.ts
@@ -1,11 +1,29 @@
 import { PickingListsFragment } from '../support/page_fragments/picking-lists.fragment';
+import { LoginPage } from '../support/page_objects/login.page';
 
 const pickingListsFragment = new PickingListsFragment();
+const loginPage = new LoginPage();
 
 describe('Login Suite', () => {
-  it('should login successfully', () => {
-    cy.login();
+  describe('when logged out', () => {
+    it('should redirect to login page', () => {
+      cy.visit('/');
 
-    pickingListsFragment.getWrapper().should('be.visible');
+      cy.location('pathname').should('to.match', /^\/login/);
+    });
+  });
+  describe('when logging in', () => {
+    it('should login successfully', () => {
+      let isLoadingPage = false;
+
+      cy.on('url:changed', (newurl) => {
+        isLoadingPage = newurl.includes('oauth') ? true : isLoadingPage;
+      });
+
+      cy.login();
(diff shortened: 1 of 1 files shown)


gund (approved):

apps/fulfillment-e2e/src/integration/login.cy.ts
```diff
@@ -1,11 +1,29 @@
 import { PickingListsFragment } from '../support/page_fragments/picking-lists.fragment';
+import { LoginPage } from '../support/page_objects/login.page';
 
 const pickingListsFragment = new PickingListsFragment();
+const loginPage = new LoginPage();
```
AndreyYevtukhov, line 5: Do we need the `loginPage` here?

apps/fulfillment-e2e/src/integration/login.cy.ts
```diff
@@ -1,11 +1,29 @@
...
+    it('should login successfully', () => {
+      let isLoadingPage = false;
+
+      cy.on('url:changed', (newurl) => {
+        isLoadingPage = newurl.includes('oauth') ? true : isLoadingPage;
+      });
+
+      cy.login();
+      cy.then(() => expect(isLoadingPage).to.be.true);
+      pickingListsFragment.getPickingListsItem().should('have.length', 2);
```
AndreyYevtukhov, line 25: I also wonder about this check -> will we always have 2 pickingListItems, or it might change unexpectedly?

aleespryker: One of the ticket requirements was to check for the number of picking list items. Is there a better way to get this number in e2e tests?

AndreyYevtukhov: We can change the requirements to make them more stable :) 

We should check that the correct page (picking list page) is opened, and to prove that we can do 2 checks:
1. wrapper check (which we have on line 26)
2. URL check `cy.url().should('contain', '/picking-list')` or something similar

In that case we don't care about the number of products displayed (because it can be 0, 1, 2, 100500 - so flaky) but still we are able to prove that login logic and redirect works (and that's what we need here)

PR #303, opened 2023-04-26
feat: oryx-search small screens behaviour

Sinucid:
On small screens input becomes floated and it's visibility controls by icon-button

closes to [HRZ-2734](https://spryker.atlassian.net/browse/HRZ-2734)

[HRZ-2734]: https://spryker.atlassian.net/browse/HRZ-2734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/form/select/src/stories/select-demo.stories.ts
```diff
@@ -151,15 +151,15 @@ SelectDemo.argTypes = {
     table: { category: 'Layout' },
   },
   searchIconPosition: {
-    options: ['PREFIX', 'SUFFIX', 'NONE'],
+    options: ['prefix', 'suffix', 'none'],
```
tobi-or-not-tobi: we better use `SearchIconPosition` here

libs/base/ui/search/searchbox/src/searchbox.component.ts
```diff
@@ -9,17 +9,19 @@ import { SearchboxController } from './searchbox.controller';
...
 
 export class SearchboxComponent
   extends LitElement
-  implements SearchOptions, FormControlOptions, AffixOptions
+  implements SearchAttributes, FormControlOptions, AffixOptions
 {
   static override styles = baseStyles;
 
+  @property({ reflect: true, type: Boolean }) open?: boolean;
+  @property({ attribute: 'not-floated' }) notFloated?: boolean;
```
tobi-or-not-tobi: what about `float = true`?

tobi-or-not-tobi: also, let's add `type: Boolean`

tobi-or-not-tobi (changes requested):
Nice work. The floating search box is not visible due to a clash with css specificity since the not-floating selector was added.

libs/base/ui/search/searchbox/src/searchbox.component.ts
Sinucid: `xs-floated` and turned off by default

tobi-or-not-tobi (approved):

PR #309, opened 2023-04-28
feat: update adapter to support search

nsilva-spryker:
- Clean and refactor adapter to search by order reference, 
- update id on picking list card, 
- create new search qualifier
- update unit tests

Closes [HRZ-2731](https://spryker.atlassian.net/browse/HRZ-2731)

[HRZ-2731]: https://spryker.atlassian.net/browse/HRZ-2731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/offline/services/adapter/picking-list-offline.adapter.ts
```diff
@@ -50,18 +50,29 @@ export class PickingListOfflineAdapter implements PickingListAdapter {
...
+                ) {
+                  const searchValue = qualifier.searchOrderReference;
+                  andCollections.push(
+                    pickingListStore
+                      .where('localStatus')
+                      .equals(qualifier.status)
+                      .and((record: PickingListSerialized) =>
+                        record.orderReferences.some((ref) =>
+                          ref.startsWith(searchValue)
+                        )
```
gund: This is not good for performance as `and()` operation is executed on the whole collection on JS side returned by the `where()` operation which is much wider.
You should not mix the 2 queries and have the search by order refs done as part of `where()` operation so it's carried by the indexed-db engine directly and not JS. Just push them to `andCollections` separately which then will be filtered by intersection on JS side.

libs/domain/picking/src/models/picking-list.ts
```diff
@@ -7,6 +7,7 @@ export interface PickingList {
   cartNote?: string;
   createdAt: Date;
   updatedAt: Date;
+  orderReferences: string[];
```
gund: Why is it added on the standard model? It should not be used here as we are not supporting this search in online mode because BAPI should be doing it which it doesn't yet. But frontend model does not need this info.

libs/domain/picking/src/services/adapter/picking-list-default.adapter.ts
```diff
@@ -188,6 +188,7 @@ export class PickingListDefaultAdapter implements PickingListAdapter {
       cartNote: cardNote,
       createdAt: new Date(data.createdAt),
       updatedAt: new Date(data.updatedAt),
+      orderReferences: data.orderReferences,
```
gund: This is not needed for default adapter.

gund (changes requested):

libs/domain/picking/offline/services/adapter/picking-list-offline.adapter.ts
nsilva-spryker: Updated after our 1on1 alignment

libs/domain/picking/src/models/picking-list.ts
nsilva-spryker: This will be cover on: https://github.com/spryker/oryx/pull/317

libs/domain/picking/src/services/adapter/picking-list-default.adapter.ts
nsilva-spryker: this will be cover on: https://github.com/spryker/oryx/pull/317

libs/base/utilities/src/misc/intersect-arrays.ts
```diff
@@ -19,17 +19,32 @@ export function intersectArrays<T>(
...
   }
 
-  const intersectingMap = new Map<unknown, T>();
+  const intersectionArray: T[] = [];
+  const [firstArray, ...restArrays] = arrays;
 
-  for (const array of arrays) {
-    for (const item of array) {
-      const id = identityFn(item);
+  firstArray.filter((value) => {
```
gund: This implementation seems like a mess, there are functional methods used but not as expected, filter should be filtering items but it's not, map should be mapping items but it's not.
Let's fix this function to properly use functional style:
1. Use filter to actually filter items of `firstArray` and assign the result to `intersectionArray`
2. Replace `restArrays.map` with `restArrays.every` to check that in every other array the item is found
3. Replace `array.map` with `array.some` to check that item exists in specific array

All of the mutable let variables are no longer needed as the state is managed by the functional array methods which you can now use. So on high level the pipeline should be like this:
```ts
firstArray.filter -> restArrays.every -> array.some
```

gund (changes requested):

libs/base/utilities/src/misc/intersect-arrays.ts
nsilva-spryker: Updated

gund (approved):

PR #312, opened 2023-05-02
feat: add radix ui support for color system

supproduction:
adds radix support for color system

closes [HRZ-2788](https://spryker.atlassian.net/browse/HRZ-2788)

[HRZ-2788]: https://spryker.atlassian.net/browse/HRZ-2788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/themes/design-tokens/src/storefront/color.tokens.ts
```diff
@@ -1,22 +1,39 @@
...
 export const color: ColorDesignTokens = {
+  // TODO: decide which color
   ink: '#121212',
 
   canvas: colorPalette.white,
+  // TODO: decide which color
+  // canvasA: newColorPalette.gray,
+
   neutral: colorPalette.gray,
+  neutralA: newColorPalette.gray,
```
tobi-or-not-tobi: 1. Let's drop old color palette.
2. Let's drop canvas (it is part of grays)
3. Let's temporarily reassign CSS vars in `root-app.styles.ts`, e.g. (see mapping in POC). We need this for FA as well when the old color palette is dropped.
4. What about separating grays and colors? This will help us to organize the colors in EB but it's also nicer for typesafety when selecting an allterative color


Add 3:
```css
host {
    --oryx-color-primary-100: var(--oryx-color-primary-3);
}
```

Add 4:

```ts
export const color: ColorDesignTokens = {
  neutral: colorPalette.grays.mauve,
  primary: colorPalette.colors.crimson,
}
```

tobi-or-not-tobi (changes requested):

libs/platform/experience/src/color/color.model.ts
```diff
@@ -0,0 +1,25 @@
+import { ColorId } from '@spryker-oryx/utilities';
+
+export const enum ColorTone {
```
tobi-or-not-tobi: I'm not so sure about this semantic list. These are quite opinionated and none of our compnents have implemented their designs according to radix semantics. Can we move this to a numbered list for now?

Also, canvast is likely not correct. We're using the additional 0-color for the text color.

libs/platform/experience/src/plugins/theme/theme-tokens.ts
```diff
@@ -201,10 +202,32 @@ export class ThemeTokens {
       let tokenMedia = DesignTokenGlobal.Global as string;
 
       if (!tokens.media && tokens.color) {
-        tokensList.push({
-          media: { [DefaultMedia.Mode]: 'light' },
-          color: tokens.color,
-        });
+        for (const [key, color] of Object.entries(tokens.color)) {
+          // TODO: delete when all colors will be replaced
```
tobi-or-not-tobi, line 206: still needed?

tobi-or-not-tobi (dismissed):

libs/platform/experience/src/plugins/theme/theme-tokens.ts
supproduction: Yes, until we have these fallbacks (plain string, not an object)
image

tobi-or-not-tobi (approved):

PR #317, opened 2023-05-03
fix: picking list card shows wrong order ID and pick up time

tarashaliuk:
closes: [HRZ-2834](https://spryker.atlassian.net/browse/HRZ-2834)

[HRZ-2834]: https://spryker.atlassian.net/browse/HRZ-2834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/models/picking-list.ts
```diff
@@ -2,6 +2,8 @@ import { PickingProduct } from './picking-product';
 
 export interface PickingList {
   id: string;
+  orderReference: string;
+  requestedDeliveryDate?: Date;
```
gund: Why is it optional? Every picking list should always have a delivery date.

libs/domain/picking/src/models/picking-list.ts
```diff
@@ -2,6 +2,8 @@ import { PickingProduct } from './picking-product';
 
 export interface PickingList {
   id: string;
+  orderReference: string;
```
gund: We already have all `orderReferences` as an array, please use it instead of adding a single string.

libs/domain/picking/src/services/adapter/picking-list-default.adapter.ts
```diff
@@ -154,6 +154,10 @@ export class PickingListDefaultAdapter implements PickingListAdapter {
     products: PickingProduct[]
   ): PickingList {
     const cardNote = data.pickingListItems[0].salesOrders[0].cartNote;
+    const orderReference =
+      data.pickingListItems[0].salesOrders[0].orderReference;
+    const requestedDeliveryDate: string =
+      data.pickingListItems[0].salesShipments[0].requestedDeliveryDate;
```
gund: Please remove this logic from offline adapter as it's now moved to here.

gund (changes requested):

libs/domain/picking/src/models/picking-list.ts
tarashaliuk: but we have something different in orderReferences 
image

libs/domain/picking/src/services/adapter/picking-list-default.adapter.ts
tarashaliuk: there is no logic regarding `orderReference` and `requestedDeliveryDate` in `picking-list-offline.adapter.ts`. Or do you mean something different?

libs/domain/picking/src/models/picking-list.ts
gund: Ok if the order references are wrong you should fix those and not create a new one because it's basically the same thing so we don't need to duplicate `orderReferences` and `orderReference` props.

libs/domain/picking/src/services/adapter/picking-list-default.adapter.ts
gund: Check offline adapter there is already a code that parses `requestedDeliveryDate` and entity also has it.
That should only be done in one place - either in offline or in default adapter but not in both.
I would keep it in offline feature because to support them in online mode we need BAPI to support filtering which it does not do yet so for now we can keep it as is in the offline adapter.

gund (changes requested):

gund (dismissed):

gund (dismissed):

gund (approved):

PR #318, opened 2023-05-03
feat: expand fulfillment login e2e tests

aleespryker:
Closes [HRZ-2826](https://spryker.atlassian.net/browse/HRZ-2826)

[HRZ-2826]: https://spryker.atlassian.net/browse/HRZ-2826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/fulfillment-e2e/src/integration/login.cy.ts
```diff
@@ -1,27 +1,81 @@
...
     });
   });
+
+  describe('when login fails', () => {
+    it('should display error notification', () => {
+      cy.intercept('POST', '**/authorize', {
+        status: 400,
+        code: 'invalid_grant',
+        message: 'The user credentials were incorrect.',
+      }).as('authorize');
```
gund, line 60: Why not just enter actually wrong credentials instead of mocking the response?

libs/template/application/src/auth/bapi/plugin.ts
```diff
@@ -34,7 +34,7 @@ export class BapiPlugin implements AppPlugin {
             (currentRoute.startsWith(oauthConfig.loginRoute) ||
               currentRoute.startsWith(redirectUrl.pathname))
           ) {
-            resolve(RouterService).navigate('/');
+            setTimeout(() => resolve(RouterService).navigate('/'), 500);
```
gund: I don't think it's a good idea to add delay to the app just for E2E tests.
Maybe you can instead intercept and add delay to a network request to achieve the same result without slowing down the app?

gund (changes requested):

apps/fulfillment-e2e/src/integration/login.cy.ts
aleespryker: Because the mock API never returns errors no matter what credentials you give it.

apps/fulfillment-e2e/src/integration/login.cy.ts
```diff
@@ -1,27 +1,81 @@
...
+        password: 'change123',
       });
 
-      cy.login();
-      cy.then(() => expect(isLoadingPage).to.be.true);
+      cy.wait('@token');
+
+      cy.location('pathname').should('match', /^\/oauth/);
+      oauthHandler.getWrapper().should('be.visible');
+      oauthHandler
```
AndreyYevtukhov: We can skip CSS checks (they are not reliable and not needed in E2E tests)

apps/fulfillment-e2e/src/integration/login.cy.ts
AndreyYevtukhov: looks nice as a temporary solution :)

libs/template/application/src/auth/bapi/plugin.ts
AndreyYevtukhov: yep, that's debatable actually... 

from user perspective thinks happen to fast and he might miss that loading screen appearance and be confused

as for me - we should have a delay here (for better UX) but make it like 250-300ms (to just have it as a "blink" but still be able to see loader)

apps/fulfillment-e2e/src/integration/login.cy.ts
```diff
@@ -1,27 +1,81 @@
...
 
 describe('Login Suite', () => {
   describe('when logged out', () => {
     it('should redirect to login page', () => {
       cy.visit('/');
 
       cy.location('pathname').should('to.match', /^\/login/);
+
+      loginPage.getWrapper().should('be.visible');
+      loginPage.getWrapper().find('oryx-image[resource="logo"]');
```
AndreyYevtukhov: We should avoid using `.get('')` and `.find('')` in test files, let's move then in page objects as UI elements.

It will make our tests more stable and easy to maintain, because selectors will be decoupled (meaning if selector changes - we should update it only in page object and not in 10 test files)

AndreyYevtukhov (changes requested):
Nice job, clean and small, I like these tests)
Left some small comments)

libs/template/application/src/auth/bapi/plugin.ts
gund: We should not introduce artificial delays to the app - if it's fast it should behave fast, the loader screen purpose is to communicate to a user that something is still happening and if there is nothing happening we should not display it just for the sake of it looking "nicer".
However the fact that loader screen pops too fast seems like an implementation bug as it should be visible until after the data has been fully loaded from BAPI into local DB which clearly does not happen, so I would suggest instead to create a bug ticket to fix the root cause rather then adding extra delays to simulate loading.

gund: Created a bug ticket for this. Let's remove the delay from this PR and comment out related E2E tests for now until the bug it resolved.

gund (dismissed):

AndreyYevtukhov (dismissed):

libs/domain/picking/offline/data-plugin.ts
```diff
@@ -28,14 +31,30 @@ export class OfflineDataPlugin extends ExecPlugin {
       const env = injector.inject(AppEnvironment, {} as AppEnvironment);
 
       const authService = injector.inject(OauthService);
+      const routerService = injector.inject(RouterService);
+      const oauthConfig = injector.inject(OauthServiceConfig);
 
       this.subscription = authService
         .isAuthenticated()
         .pipe(
+          distinctUntilChanged(),
```
gund: I don't think this will prevent multiple DB cleanups/populations as if you just reload the page it will still perform full DB drop and repopulate the data.
As we discussed before this data fetching logic should be performed _only_ after the login and not generically on the application startup.

gund (changes requested):

AndreyYevtukhov (dismissed):
Looks good!

AndreyYevtukhov (dismissed):

gund (approved):

PR #327, opened 2023-05-09
chore: checkout login refactor orchestrate

tobi-or-not-tobi:
- Checkout orchestration is done from the checkout component. This allows to implement various orchestration and validation strategies going forward.
- Checkout Data Service is used to (re)load checkout data
- Checkout State Service is used to persist checkout state; checkout state is the data collected from checkout forms.


libs/base/ui/form/input/src/error/error.controller.ts
```diff
@@ -3,16 +3,17 @@ import { ErrorOptions } from './error.model';
 
 export class ErrorController implements ReactiveController {
   hostUpdated?(): void {
-    this.handleError();
+    // this.handleError();
```
dunqan, line 6: 👀

libs/domain/checkout/delivery/delivery.component.ts
```diff
@@ -0,0 +1,48 @@
...
+          html`<oryx-checkout-manage-address></oryx-checkout-manage-address>`
+      )}
+      <oryx-checkout-address
+        @selectedAddress=${this.onChangeAddress}
+        .address=${this.selected()}
+      ></oryx-checkout-address>
+    `;
+  }
+
+  validate(report: boolean): boolean {
```
dunqan: as discussed, we should consider making it bit more specific, also having an interface to better adjust typesafety of it could make sense

libs/domain/checkout/src/models/checkout.model.ts
```diff
@@ -1,4 +1,61 @@
...
+  Invalid = 'invalid',
+
+  // Initializing = 'initializing',
+  // NotAvailable = 'n/a',
+
+  // Validate = 'validate',
+  // Busy = 'busy',
+}
+
+// export const enum CheckoutProcess {
```
dunqan: 👀

libs/domain/checkout/src/services/default-checkout-data.service.ts
```diff
@@ -1,155 +1,189 @@
...
+  );
+
+  protected load$ = this.cartId.pipe(
+    switchMap((cartId) =>
+      cartId
+        ? this.adapter.get({
+            cartId,
+            // TODO: consider making this configurable by the type of checkout we're using
+            include: [
+              ApiCheckoutModel.Includes.Shipments,
```
dunqan: Ideally we will move backend specific details (and usage of backend models) to adapters

libs/domain/checkout/src/services/default-checkout.service.ts
```diff
@@ -1,140 +1,88 @@
...
-                  new Error(
-                    'Cannot checkout, check your cart value limits and validations'
-                  )
-              )
-        ),
-        switchMap((payload) =>
-          this.adapter.placeOrder({ attributes: payload })
-        ),
-        tap((response) => this.postCheckout(response))
+      this.dataService.collect().pipe(
```
dunqan: Ideally we would use command here, let's maybe add a ToDo?

libs/domain/checkout/src/services/default-checkout-data.service.ts
```diff
@@ -1,155 +1,189 @@
...
+  implements CheckoutDataService<CheckoutData, Checkout>
 {
-  protected isGuestCheckout$ = new BehaviorSubject<boolean>(false);
-  protected contactDetails$ = new BehaviorSubject<ContactDetails | null>(null);
-  protected addressDetails$ = new BehaviorSubject<Address | null>(null);
-  protected shipmentDetails$ = new BehaviorSubject<Shipment | null>(null);
-  protected paymentDetails$ = new BehaviorSubject<PaymentMethod | null>(null);
-
-  protected subscription?: Subscription;
+  protected subject = new BehaviorSubject<
```
dunqan: Ideally we will use query for this.

PR #328, opened 2023-05-10
feat: implement picking in progress modal

aleespryker:
Closes [HRZ-2688](https://spryker.atlassian.net/browse/HRZ-2688)

[HRZ-2688]: https://spryker.atlassian.net/browse/HRZ-2688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/services/picking-list-default.service.ts
```diff
@@ -16,8 +16,9 @@ export class PickingListDefaultService implements PickingListService {
   startPicking(pickingList: PickingList): Observable<PickingList | null> {
     this.upcomingPickingListId$.next(pickingList.id);
     return this.adapter.startPicking(pickingList).pipe(
-      catchError(() => {
+      catchError((e) => {
         this.upcomingPickingListId$.next(null);
+        throw e;
         return of(null);
```
gund: You should not throw and return as both do not makes sense together.
Instead you can return an error stream using `throwError(() => e)` operator.

libs/domain/picking/src/services/adapter/picking-list-default.adapter.ts
```diff
@@ -41,12 +42,25 @@ export class PickingListDefaultAdapter implements PickingListAdapter {
...
+        map((response) => {
+          if (response.errors) {
+            for (let i = 0; i < response.errors.length; i++) {
+              const error = response.errors[i];
+              const pickingListError = new Error(
+                error.message
+              ) as PickingListError;
+              pickingListError.status = error.status;
+              pickingListError.code = error.code;
+              throw pickingListError;
```
gund: Because you are throwing an error immediately you do not need a for loop, just grab the first element directly.

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
```diff
@@ -76,6 +82,25 @@ export class PickingListsComponent extends LitElement {
     `;
   }
 
+  protected renderPickingInProgressModal(): TemplateResult {
+    return html`<oryx-modal
```
gund: Since you have the same modal duplicated in 2 places maybe it's good idea to move it into a component to reuse?

gund (changes requested):

libs/domain/picking/src/components/picking-in-progress/picking-in-progress.component.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export class PickingInProgressModalComponent extends LitElement {
+  @state()
+  protected pickingInProgress?: boolean;
+
+  open(): void {
+    this.pickingInProgress = true;
+  }
+
+  protected override render(): TemplateResult {
```
Sinucid, line 9: What about simplify a bit the `oryx-picking-in-progress-modal`?
1) I believe that it should have property that will control visibility of the modal inside
2) On click on close button it will dispatches `oryx.close` event that allows adding some side effect in parent component and close it

like this:
```
export class PickingInProgressModalComponent extends LitElement {
  @property({type: Boolean}) open?: boolean;

  return html`<oryx-modal
      ?open=${this.open}
      ...
      @oryx.close=${this.onClose}
    >
        ...
        <button @click=${this.onClose}>
        ...
    </oryx-modal>`;

    protected onClose(): void {
       this.dispatchEvent(new CustomEvent('oryx.close', {composed: true, bubbles: true}))
    }
```

And parent component will control the rest

Sinucid (changes requested):

libs/domain/picking/src/components/picking-in-progress/picking-in-progress.component.ts
aleespryker: The events bubble without the need to further dispatch them, as it works already in the customer note component.

Sinucid (dismissed):

libs/domain/picking/src/components/customer-note/customer-note.component.ts
```diff
@@ -32,7 +31,9 @@ export class CustomerNoteComponent extends PickingListMixin(LitElement) {
         ),
         catchError((e) => {
           if (e.status === 409) {
-            this.pickingInProgress = true;
+            const modal = this.pickingInProgressModal
+              .value as PickingInProgressModalComponent;
```
gund: You can move the type to the `createRef()` so the type cast is not needed here:
```
createRef<PickingInProgressModalComponent>()
```
Also the ref may be undefined so you need to check that it exists before assigning anything to it.

libs/domain/picking/src/services/picking-http-default.service.ts
```diff
@@ -54,10 +48,22 @@ export class PickingHttpDefaultService implements PickingHttpService {
...
+
+  protected expandContentType<T = unknown>(
+    options?: RequestOptions<T>
+  ): RequestOptions<T> {
+    return {
+      ...options,
+      headers: {
+        ...options?.headers,
+        'Content-Type': 'application/vnd.api+json',
+      },
```
gund: Btw headers may be of different type then simple object, it may be nested array or [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) instance, in which case this logic will be broken.
You should probably convert headers to a Headers instance as it will take care of handling any format type and then use it's method [`Headers.set()`](https://developer.mozilla.org/en-US/docs/Web/API/Headers/set) to set Content-Type header.

libs/domain/picking/src/services/picking-list-default.service.ts
```diff
@@ -54,8 +54,7 @@ export class PickingListDefaultService implements PickingListService {
     return this.adapter.startPicking(pickingList).pipe(
       catchError((e) => {
         this.upcomingPickingListId$.next(null);
-        throw e;
-        return of(null);
+        return throwError(e);
```
gund: This function expects a function factory of the error observable:
```suggestion
        return throwError(() => of(e));
```

gund (changes requested):

gund (dismissed):

libs/domain/picking/src/services/picking-list-default.service.ts
gund: This should actually return just the error, not observable of error:
```
return throwError(() => e);
```

Sinucid (approved):

PR #331, opened 2023-05-11
feat: FA filters

Sinucid:
Tests will be added in the separate sub-task

closes to [HRZ-2862](https://spryker.atlassian.net/browse/HRZ-2862), [HRZ-2863](https://spryker.atlassian.net/browse/HRZ-2863)

[HRZ-2862]: https://spryker.atlassian.net/browse/HRZ-2862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HRZ-2863]: https://spryker.atlassian.net/browse/HRZ-2863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/overlays/modal/src/modal.component.ts
```diff
@@ -136,15 +136,13 @@ export class ModalComponent extends LitElement implements ModalProperties {
...
-                type="button"
-                aria-label="navigate back"
-                @click=${this.onGoBack}
-              >
-                <oryx-icon type="back"></oryx-icon>
-              </button>
-            </oryx-icon-button>
+            <slot name="navigate-back" @click=${this.onGoBack}>
+              <oryx-icon-button>
+                <button type="button" aria-label="navigate back">
```
gund: We should use i18n to translate aria labels.

libs/domain/picking/src/components/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    this.filters?.toggleAttribute('open');
+  }
+
+  protected override render(): TemplateResult {
+    return html`
+      <oryx-toggle-icon size=${Size.Sm}>
+        <input
+          type="checkbox"
+          placeholder="toggle filters"
```
gund: Please use i18n here as well.

libs/domain/picking/src/components/filters/filters.model.ts
```diff
@@ -0,0 +1,33 @@
...
+export const fields: FormFieldDefinition[] = [
+  {
+    id: 'sortBy',
+    type: FormFieldType.RadioList,
+    label: i18n('picking.filter.sort-by') as string,
+    attributes: { direction: 'vertical' },
+    options: [
+      {
+        value: 'requestedDeliveryDate.desc',
+        text: 'Earliest pick-up time',
```
gund: These texts should also be translated by i18n.

libs/domain/picking/src/models/picking-list-qualifier.ts
```diff
@@ -11,5 +11,6 @@ export interface PickingListQualifier
 }
 
 export enum PickingListQualifierSortBy {
-  deliveryDate = 'deliveryDate',
+  RequestedDeliveryDate = 'requestedDeliveryDate',
+  ItemsCount = 'itemsCount',
```
gund: It's better to keep the original name as `DeliveryDate` as the `requestedDeliveryDate` is implementation detail.
Same goes for a new value `ItemsCount` better to rename to `OrderSize` to reflect the domain and not implementation details.
You can then map these values to actual fields in the adapter.

libs/domain/picking/src/services/picking-list-default.service.ts
```diff
@@ -1,16 +1,56 @@
...
+  ): void {
+    this.sortingQualifier$.next(qualifier);
+  }
 
   get(qualifier: PickingListQualifier): Observable<PickingList[]> {
-    return this.adapter.get(qualifier);
+    return this.sortingQualifier$.pipe(
+      switchMap((sortingQualifier) =>
+        this.adapter.get({
+          ...qualifier,
```
gund, line 46: More specific qualifier should win over the generic qualifier so please swap them around.

libs/domain/picking/src/services/picking-list-default.service.ts
```diff
@@ -1,16 +1,56 @@
...
+    protected adapter = inject(PickingListAdapter),
+    protected defaultSortingQualifier = {
+      sortBy: PickingListQualifierSortBy.RequestedDeliveryDate,
+      sortDesc: false,
+    }
+  ) {}
+
+  protected sortingQualifier$ =
+    new BehaviorSubject<SortableQualifier<PickingListQualifierSortBy> | null>(
+      null
```
gund: You do not need `null` here as you have a default qualifier - just put it there.

gund (changes requested):

libs/base/ui/form/input-list/src/input-list.styles.ts
```diff
@@ -36,3 +37,14 @@ export const styles = css`
...
   }
 `;
+
+export const screenStyles = screenCss({
+  sm: css`
+    legend {
+      margin-block-end: 20px;
+      font-weight: 500;
+      font-size: 14px;
+      line-height: 22px;
```
tobi-or-not-tobi: Both large and small screens should use standard body typography. Let's use design tokens here.

libs/base/ui/form/input-list/src/component.ts
```diff
@@ -4,4 +4,9 @@ export const inputListComponent = componentDef({
   name: 'oryx-input-list',
   impl: () =>
     import('./input-list.component').then((m) => m.InputListComponent),
+  stylesheets: [
+    {
+      rules: () => import('./input-list.styles').then((m) => m.screenStyles),
+    },
+  ],
```
tobi-or-not-tobi, line 11: after using design tokens for typography, we have only 1 rule left, let's inline it here rather than lazy load the css on small screen.

libs/domain/picking/src/components/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  @subscribe()
+  protected selectedFilter = resolve(PickingListService)
+    .getSortingQualifier()
+    .pipe(
+      tap((sortingQualifier) => {
+        if (!this.input) return;
+        this.input.checked = !!sortingQualifier;
+      })
+    );
```
tobi-or-not-tobi: Can we change to:

```ts
protected selected = signal(
  resolve(PickingListService).getSortingQualifier()
);

<input ?checked=${!!this.selected()} />
```

Nitpick, can we move onclick below render?

libs/domain/picking/src/components/index.ts
```diff
@@ -1,4 +1,6 @@
 export * from './customer-note/customer-note.def';
+export * from './filter-button/filter-button.def';
+export * from './filters/filters.def';
```
tobi-or-not-tobi: we need to export the component as well:
- add an index.ts to each component dir
- export the components, styles, def, etc in each barrel
- export the barrels in this component barrel

libs/domain/picking/src/components/picking-lists/picking-lists.component.ts
```diff
@@ -87,6 +88,19 @@ export class PickingListsComponent extends LitElement {
     `;
   }
 
+  protected renderSorting(): TemplateResult {
+    return html`
+      <div class="filters">
+        <span>
+          ${i18n('picking.filter.<value>-open-orders', {
+            value: this.pickingLists!.length ?? 0,
```
tobi-or-not-tobi:
```suggestion
            value: this.pickingLists?.length ?? 0,
```

libs/domain/picking/src/components/filters/filters.styles.ts
```diff
@@ -0,0 +1,21 @@
...
+    font-weight: 600;
+  }
+
+  oryx-modal::part(close) {
+    --oryx-icon-size: 24px;
+    --_margin: 0;
+
+    padding: 0;
+    max-height: 26px;
+  }
```
tobi-or-not-tobi: can we make this small window driven? Meaning, icon button will have increased size on small screen? Or modal will show increased icon button small screen?

libs/domain/picking/src/components/filters/filters.model.ts
tobi-or-not-tobi: yeah, i'd make those text a i18n key

tobi-or-not-tobi (changes requested):

libs/domain/picking/src/services/picking-list-default.service.ts
Sinucid: Yeah, the idea of null was to simplify the logic. We need to know whether some filter selected or not. If we set default value, we need to make this default qualifier public for comparison. But now we can just check whether it something or null

libs/base/ui/form/input-list/src/component.ts
Sinucid: Replaced values by tokens

libs/domain/picking/src/components/filter-button/filter-button.component.ts
Sinucid: Unfortunately - not. Checked attr just and initial value, it does not have affect on the property after.

Sinucid: switched to signals

gund (dismissed):

libs/domain/picking/src/components/filter-button/filter-button.component.ts
tobi-or-not-tobi: I've asked @dunqan to double check as this is expected to work.

libs/domain/picking/src/components/index.ts
tobi-or-not-tobi: My bad, we should not export def file in the main barrel as it will break lazy loading of the components.

libs/base/ui/form/input-list/src/component.ts
tobi-or-not-tobi: You could consider to use:

```ts
${headingUtil(HeadingTag.Subtitle)}
```

libs/domain/picking/src/components/filter-button/filter-button.component.ts
Sinucid: Don't you mind if I fix (or not) this later, to not to block this pr?

tobi-or-not-tobi (dismissed):

libs/domain/picking/src/components/filter-button/filter-button.component.ts
Sinucid: Discussed with Stan. We justified the signal's implementation, but as I mentioned before this does not affecting the property, that why I added mixed solution that triggers the property and attribute

tobi-or-not-tobi (dismissed):

gund (approved):

PR #333, opened 2023-05-11
chore: semantic split layouts

tobi-or-not-tobi:
Introduced three semantic split layout types: 
- equal
- main
- aside

The layouts are still based on the global grid system, but you no longer need to come up with a split factor. Instead, the split column size is given by design tokens. 

closes [HRZ-2895](https://spryker.atlassian.net/browse/HRZ-2895)

[HRZ-2895]: https://spryker.atlassian.net/browse/HRZ-2895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/src/services/layout/default-layout.service.ts
```diff
@@ -98,9 +98,23 @@ export class DefaultLayoutService implements LayoutService {
           )
         );
 
-      case CompositionLayout.SplitColumn:
+      case CompositionLayout.Split:
```
supproduction, line 101: Please, update unit testing also with new types

supproduction (changes requested):

supproduction (approved):

PR #334, opened 2023-05-11
fix: fa customer note e2e test flakiness

AndreyYevtukhov:
closes https://spryker.atlassian.net/browse/HRZ-2894

### DIFF
--- a/apps/fulfillment-e2e/src/integration/customer-note.cy.ts
+++ b/apps/fulfillment-e2e/src/integration/customer-note.cy.ts
@@ -7,7 +7,7 @@ const customerNoteFragment = new CustomerNoteModalFragment();
 describe('When the user opens customer note', () => {
   beforeEach(() => {
     cy.login();
-    pickingListsFragment.getCustomerNoteButton().click();
+    pickingListsFragment.getCustomerNoteButtons().eq(0).click();
   });
 
   describe('and user clicks on close button', () => {
--- a/apps/fulfillment-e2e/src/integration/picking-lists.cy.ts
+++ b/apps/fulfillment-e2e/src/integration/picking-lists.cy.ts
@@ -10,6 +10,6 @@ describe('Picking Lists', () => {
 
   it('should display picking lists', () => {
     pickingListsFragment.getWrapper().should('be.visible');
-    pickingListsFragment.getPickingListsItem().should('be.visible');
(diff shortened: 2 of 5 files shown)


apps/fulfillment-e2e/src/integration/customer-note.cy.ts
```diff
@@ -7,7 +7,7 @@ const customerNoteFragment = new CustomerNoteModalFragment();
 describe('When the user opens customer note', () => {
   beforeEach(() => {
     cy.login();
-    pickingListsFragment.getCustomerNoteButton().click();
+    pickingListsFragment.getCustomerNoteButtons().eq(0).click();
```
tolerants, line 10: What about to add `getCustomerNoteFirstButton`?

tarashaliuk (dismissed):

tolerants (dismissed):

apps/fulfillment-e2e/src/integration/customer-note.cy.ts
AndreyYevtukhov: In that case, in the future, we might have `getCustomerNoteSecondButton`, `getCustomerNoteThirdButton` etc. buttons, which is not nice :)

AndreyYevtukhov: Actually, we can improve the `PickingListsFragment` class, which will allow us to remove that `.eq()` call, but it will require some time (e.g. 3-4 hours) which I don't have ))

tarashaliuk (approved):

apps/fulfillment-e2e/src/integration/customer-note.cy.ts
tolerants: I see your point, and totally agree that `FirstButton` not the best name, but was is the purpose of this button? So my suggestion is to name this variable according to it purpose but not position in the DOM. `getCustomerNoteSubmitButton` for example (The issue is I don't know what this button is doing :))
So in the future if the place of this button will be changed then we shouldn't touch every place where it's used but only one place where are getting `getCustomerNoteSubmitButton` (for ex.).

AndreyYevtukhov: Everything should be solved by refactoring of the PickingListsFragment class :) 
it will make it more clear)

tolerants (approved):

PR #335, opened 2023-05-11
feat: layout divider

tobi-or-not-tobi:
The divider is used to divide layout items by a thin line. It is used inside the footer component to separate out footer links, but is can potentially also be used to separate out larger layout elements such as cart entries, checkout steps, etc. 

closes [HRZ-2893](https://spryker.atlassian.net/browse/HRZ-2893)

[HRZ-2893]: https://spryker.atlassian.net/browse/HRZ-2893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront-e2e/src/integration/spa-navigation.cy.ts
```diff
@@ -5,19 +5,30 @@ const landingPage = new LandingPage();
...
-  it('must navigate to Contact page from header', () => {
-    landingPage.visit();
-    landingPage.header.getContactLink().click();
+  describe('when the user clicks on the "Contact us" link', () => {
+    beforeEach(() => {
+      landingPage.visit();
+      // cy.wait(2000);
+      // cy.scrollTo('bottom');
+      // const link = landingPage.footer.getLink('/contact');
+      // link.click();
```
supproduction: Is it temporary?

libs/platform/experience/src/mixins/layout.mixin.ts
```diff
@@ -47,6 +50,8 @@ export const LayoutMixin = <T extends Type<LitElement & LayoutAttributes>>(
     @signalProperty({ type: Boolean }) bleed?: boolean;
     @signalProperty({ type: Boolean }) sticky?: boolean;
     @signalProperty({ type: Boolean }) overlap?: boolean;
+    @signalProperty({ type: Boolean }) divider?: boolean;
+    @property({ type: Boolean, reflect: true }) vertical?: boolean;
```
supproduction, line 54: Should we use `signalProperty` here?

libs/platform/experience/src/services/layout/default-layout.service.ts
```diff
@@ -70,6 +70,13 @@ export class DefaultLayoutService implements LayoutService {
           )
         );
 
+      case 'divider':
```
supproduction, line 73: Should we test this case?

supproduction (changes requested):

libs/platform/experience/src/mixins/layout.mixin.ts
tobi-or-not-tobi: No need to afaik, as the vertical is only used as a reflected property in various layout rules rather than being observed to load style rules. I did however improved a few small things

supproduction (dismissed):

apps/storefront-e2e/src/integration/spa-navigation.cy.ts
tobi-or-not-tobi: yup, should be fixed by now

supproduction (approved):

PR #336, opened 2023-05-12
chore: apply radix colors in UI system

tobi-or-not-tobi:
We had a few places where we did not correctly leverage the colors. 


libs/template/presets/src/b2c/experience/header.ts
```diff
@@ -34,7 +34,7 @@ export const HeaderTemplate: StaticComponent = {
           rules: [
             {
               layout: 'flex',
-              background: 'var(--oryx-color-primary-500)',
```
tolerants, line 37: Hm, is it a good idea to use direct color value?

tolerants (dismissed):

libs/template/presets/src/b2c/experience/header.ts
tobi-or-not-tobi: It's an exception; when we use the design token, we get a very ugly header with all colors in dark mode.

Sinucid (dismissed):

tobi-or-not-tobi:
we'll retry

tobi-or-not-tobi:
recreated at #338

PR #339, opened 2023-05-13
chore: apply radix colors in UI system

tobi-or-not-tobi:
We had a few places where we did not correctly leverage the colors. 


apps/storybook/project.json
```diff
@@ -38,7 +38,7 @@
     "visual-regression-pr": {
       "executor": "@nrwl/workspace:run-commands",
       "options": {
-        "command": "chromatic --ci --only-changed --exit-zero-on-changes"
```
supproduction: Assume we should revert it back

libs/platform/experience/src/color/colors/colored/mint.ts
```diff
@@ -2,7 +2,7 @@ import { Color } from '../../color.model';
 
 export const mint: Color = {
   light: {
-    0: 'hsl(0, 0%, 100%)',
+    0: 'hsl(0, 0%, 0%)',
```
supproduction, line 5: Is it correct that for dark and light mode we have the same color at `0` position?

supproduction (dismissed):

libs/platform/experience/src/color/colors/colored/mint.ts
tobi-or-not-tobi: yes. The same "solid" color is used for the middle position (used to be our 300 color, how it is position 9). In both dark and light, the text in the solid color is white on most colours, but black on the bright colours.

apps/storybook/project.json
tobi-or-not-tobi: Thx for the heads up. Chromatic did not report the regressions, hence we found this workaround. Will revert it.

supproduction (approved):

PR #342, opened 2023-05-16
feat: picking in progress e2e tests

aleespryker:
Closes [HRZ-2687](https://spryker.atlassian.net/browse/HRZ-2687)

[HRZ-2687]: https://spryker.atlassian.net/browse/HRZ-2687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/fulfillment-e2e/src/integration/start-picking-picklist-in-progress.cy.ts
```diff
@@ -0,0 +1,56 @@
...
+            status: 409,
+            code: '5310',
+          },
+        ],
+      },
+    }).as('picking-in-progress');
+
+    pickingListsFragment.getStartPickingButtons().eq(1).should('be.visible');
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(300);
```
AndreyYevtukhov: Why do we need this `wait`? Can we change it to something else? (Each wait makes tests less stable, that's why we should add them very carefully)

apps/fulfillment-e2e/src/integration/start-picking-picklist-in-progress.cy.ts
```diff
@@ -0,0 +1,56 @@
...
+    pickingListsFragment.pickingInProgressModal
+      .getModal()
+      .should('have.attr', 'open');
+  });
+
+  it('should render picking in progress text', () => {
+    pickingListsFragment.pickingInProgressModal
+      .getModal()
+      .should('contain.text', 'Already in progress');
+  });
```
AndreyYevtukhov: Let's combine these in 1 test (because we are paying for each `it` block - and we don't want to have hundreds of super small tests run ~30 times per day, right?)

apps/fulfillment-e2e/src/integration/start-picking-picklist-with-customer-note.cy.ts
```diff
@@ -73,4 +73,52 @@ describe('Start picking a picklist with customer note', () => {
       pickingFragment.getProducts().should('have.length.at.least', 1);
     });
   });
+
+  describe('and picking is already in progress', () => {
```
AndreyYevtukhov, line 77: same `wait` and small `it` tests that should be combined here :)

apps/fulfillment-e2e/src/integration/start-picking-picklist-in-progress.cy.ts
AndreyYevtukhov: Maybe it can be replaced with `cy.wait('@picking-in-progress')`? Cause I see that we add a `picking-in-progress` alias, but we don't use it

apps/fulfillment-e2e/src/integration/start-picking-picklist-with-customer-note.cy.ts
```diff
@@ -73,4 +73,52 @@ describe('Start picking a picklist with customer note', () => {
...
+        body: {
+          errors: [
+            {
+              message: 'Picklist is already being picked by another user.',
+              status: 409,
+              code: '5310',
+            },
+          ],
+        },
+      }).as('picking-in-progress');
```
AndreyYevtukhov: Maybe we can refactor it somehow to not have 2 similar interceptors in 2 files? 
It might be some kind of global function or cypress action

AndreyYevtukhov (changes requested):
Left some comments

apps/fulfillment-e2e/src/integration/start-picking-picklist-with-customer-note.cy.ts
aleespryker: I still ended up having to keep the wait because it didn't work without it. It's using the same format as the `beforeEach` above it: https://github.com/spryker/oryx/blob/488ca865257b8e92c8ef95331a24519719bb152a/apps/fulfillment-e2e/src/integration/start-picking-picklist-with-customer-note.cy.ts#L47-L48

gund (dismissed):

AndreyYevtukhov (approved):

PR #345, opened 2023-05-16
refactor: move icon types list

supproduction:
moves icon types list
adds font icons replacers
removes chromatic injectables
drop asyncValue from injectables
splits data-clit into revealers
adds icon list into schema

closes [HRZ-2885](https://spryker.atlassian.net/browse/HRZ-2885) [HRZ-2889](https://spryker.atlassian.net/browse/HRZ-2889)

[HRZ-2885]: https://spryker.atlassian.net/browse/HRZ-2885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HRZ-2889]: https://spryker.atlassian.net/browse/HRZ-2889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/presets/resources/icons/backoffice/font-icons.ts
```diff
@@ -0,0 +1,102 @@
+import { IconMapper } from '@spryker-oryx/experience';
+import { IconTypes } from '../icon.model';
+
+export const backofficeFontAwesomeIcons: IconMapper = {
```
tobi-or-not-tobi: I'd say this is not the `backofficeFontAwesomeIcons`, but just the `fontAwesomeIcons`.

tobi-or-not-tobi: we have `fontAwesomeIcons` in the `font-icon.ts` file, should we not keep all mappings there?

apps/storefront/src/index.html
```diff
@@ -1,8 +1,6 @@
 <!DOCTYPE html>
 <html>
-  <head>
-    <meta charset="UTF-8" />
-  </head>
+  <meta charset="UTF-8" />
```
tobi-or-not-tobi: What happened here? Customers will quit likely add some boilerplate head tags, so we should be prepared for the head to be available.

libs/template/themes/src/storefront.theme.ts
```diff
@@ -1,18 +1,17 @@
...
-  //   resources: [
-  //     {
-  //       resource: fontAwesomeIcons,
-  //       types: [IconTypes.User],
-  //     },
-  //   ],
-  // },
+  icons: {
+    resource: storefrontMaterialIcons,
+    resources: storefrontIconSources,
```
tobi-or-not-tobi: I'd say the storefront theme imports `materialDesignIcons` and  `storefrontIconSources`

tobi-or-not-tobi (changes requested):

apps/storefront/src/index.html
supproduction: We prepared for it. I fixed meta service, so it works as with header as without

libs/template/presets/resources/icons/backoffice/font-icons.ts
supproduction: I moved mappings per theme, so we can change icons

apps/storybook/.storybook/app/chromatic-styles.ts
```diff
@@ -62,7 +62,6 @@ export const chromaticStyledComponents = [
...
           }
-
           :host([type='critical'][loading]) ::slotted(*)::before {
             background-image: url('data:image/svg+xml,%3Csvg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="red"%3E%3Cg%3E%3Cellipse opacity="0.7" rx="3.48584" ry="3.4859" transform="matrix(0.80444 0.594034 -0.593986 0.804475 11.925 4.87503)" /%3E%3Cellipse rx="3.48584" ry="3.4859" transform="matrix(0.80444 0.594034 -0.593986 0.804475 20.1242 10.9278)" /%3E%3Cellipse opacity="0.4" rx="3.48584" ry="3.4859" transform="matrix(0.80444 0.594034 -0.593986 0.804475 5.87224 13.0723)" /%3E%3Cellipse opacity="0.3" rx="3.48584" ry="3.4859" transform="matrix(0.80444 0.594034 -0.593986 0.804475 14.0695 19.125)" /%3E%3C/g%3E%3C/svg%3E');
```
tobi-or-not-tobi: FYI this fix is no longer moving since we moved to colored icons for the loading effect...

libs/domain/search/pagination/src/pagination.schema.ts
```diff
@@ -0,0 +1,22 @@
...
+    icon: IconTypes.ArrowsOutward,
+    options: {
+      max: {
+        type: FormFieldType.Number,
+        label: 'Max number of visible pages',
+        width: 100,
+      },
+      enableControls: {
+        type: FormFieldType.Boolean,
+        label: 'Enable controls',
```
tobi-or-not-tobi: We've added resolving a standard name from the type name (splitting camel case and add a space). We should avoid label for most of the field definitions.

tobi-or-not-tobi: please fix everywhere

apps/storefront/src/index.html
tobi-or-not-tobi: i'm in favour of keeping the `head` element, to make it more obvious that we use standard html, but i'm not sure about motivation here.

libs/template/themes/src/storefront.theme.ts
tobi-or-not-tobi: can we change those names?

apps/storefront/src/index.html
supproduction: We had this template at start. Head was adding  by me in case of `page head service`, I thought it was a limitation so I fixed this moment.
https://github.com/spryker/oryx/commit/860ad6f524feb19b8776bb572827790de26d6f00#diff-c87a2a62fcdc8aa20f501092830edff7a5a5184162cfbb287717cfff2c464bed

libs/template/themes/src/storefront.theme.ts
supproduction: storefrontMaterialIcons => materialDesignIcons
storefrontIconSources => which name ?

tobi-or-not-tobi (dismissed):

libs/template/presets/resources/icons/storefront/font-icons.ts
```diff
@@ -0,0 +1,123 @@
...
+    [IconTypes.User]: 'person',
+    [IconTypes.Image]: 'photo_library',
+    [IconTypes.ModeLight]: 'light_mode',
+    [IconTypes.ModeDark]: 'dark_mode',
+    [IconTypes.Ratings]: 'reviews',
+    [IconTypes.Planet]: 'language',
+    [IconTypes.Locker]: 'lock',
+    [IconTypes.Images]: 'images',
+  },
+};
```
tobi-or-not-tobi: This is the material design icon set. Can we:
- rename it to `materialDesignIcons`
- move it to `icons/material-design.icons.ts` (or similar name)

The intend of this definition is that it is not storefront related.

libs/template/presets/resources/icons/storefront/font-icons.ts
```diff
@@ -0,0 +1,123 @@
...
+    [IconTypes.NavigationArrow]: 'arrow_right_alt',
+    [IconTypes.Actions]: 'more_vert',
+    [IconTypes.User]: 'person',
+    [IconTypes.Image]: 'photo_library',
+    [IconTypes.ModeLight]: 'light_mode',
+    [IconTypes.ModeDark]: 'dark_mode',
+    [IconTypes.Ratings]: 'reviews',
+    [IconTypes.Planet]: 'language',
+    [IconTypes.Locker]: 'lock',
+    [IconTypes.Images]: 'images',
```
tobi-or-not-tobi: Can we add the following missing types:
```
    [IconTypes.Visible]: 'visibility',
    [IconTypes.Invisible]: 'visibility_off',
    [IconTypes.Parcel]: 'deployed_code',
```

libs/template/presets/resources/icons/storefront/font-icons.ts
```diff
@@ -0,0 +1,123 @@
...
+    [IconTypes.BulletList]: '&#xf0ca;',
+    [IconTypes.ArrowsOutward]: '&#xf337;',
+    [IconTypes.ViewList]: '&#xf00b;',
+    [IconTypes.InputStepper]: '&#xf0d8;',
+    [IconTypes.Imports]: '&#xf019;',
+    [IconTypes.Copy]: '&#xf0c5;',
+    [IconTypes.File]: '&#xf15b;',
+    [IconTypes.Imports]: '&#xf019;',
+  },
+};
```
tobi-or-not-tobi: Storefront should not use any font awesome fonts. We should:
- rename to `fontAwesomeIcons`
- move to `icons/fontawesome.icons.ts`

libs/template/presets/resources/icons/storefront/font-icons.ts
```diff
@@ -0,0 +1,123 @@
...
+      IconTypes.ViewList,
+      IconTypes.InputStepper,
+      IconTypes.Imports,
+      IconTypes.Copy,
+      IconTypes.File,
+      IconTypes.Imports,
+    ],
+  },
+];
+
```
tobi-or-not-tobi: This definition should not exist, and after introducing visible and visible_off to material design I have not seen any regression. Can we remove it? Or do we depend on it in the icons for EB? If the latter is true, we can keep it for now and find a way to drop it going forward.

libs/template/themes/src/storefront.theme.ts
tobi-or-not-tobi: As suggested elsewhere in this PR we should endup with:

```ts
icons: {
    resource: materialDesignIcons,
}
```

Only if EB requires to keep storefrontIconSources, but for clearness we could rename it.

tobi-or-not-tobi (changes requested):

libs/template/presets/resources/icons/storefront/font-icons.ts
```diff
@@ -0,0 +1,123 @@
...
+export const storefrontFontIcons: ResourceFonts = {
+  [storefrontMaterialIcons.id]:
+    'https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@20..48,100..700,0..1,-50..200&display=block',
+  [storefrontFontAwesomeIcons.id]:
+    'https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.4.0/css/all.min.css',
```
tobi-or-not-tobi: I'd move the font definitions inside the materialDesign and fontAwesome icon definitions, and reference them in the app resources:

```ts
export const storefrontResources: Resources = {
  graphics: { ...commonGraphics, ...brandGraphics },
  icons: {
    list: storefrontIcons,
    types: IconTypes,
  },
  fonts: [materialDesignFont, fontAwesomeFont],
};
```

However, driven by other comments in this PR, we should drop `fontAwesomeFont`.

tobi-or-not-tobi (changes requested):

libs/template/presets/resources/icons/storefront/font-icons.ts
supproduction: In order that we use IconTypes as fallback so we should define all icons or we will see just text in storybook. I fixed it to add all definitions into materialDesign icons, as we don't use `storefront\bo` prefix I guess it's ok when we have the full icon list of our IconTypes

libs/template/themes/src/storefront.theme.ts
```diff
@@ -1,18 +1,13 @@
 import { Theme } from '@spryker-oryx/experience';
+import { materialDesignIcons } from '@spryker-oryx/presets/resources';
 import { defaultBreakpoints } from '@spryker-oryx/themes/breakpoints';
 
 export const storefrontTheme: Theme = {
```
tobi-or-not-tobi, line 5: nitpick: no need to export

tobi-or-not-tobi (approved):

PR #355, opened 2023-05-22
fix: use click password visibility for picking login

aleespryker:
Closes [HRZ-2914](https://spryker.atlassian.net/browse/HRZ-2914)

[HRZ-2914]: https://spryker.atlassian.net/browse/HRZ-2914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AndreyYevtukhov (approved):

libs/domain/picking/src/components/login/login.component.ts
```diff
@@ -13,7 +14,10 @@ export class LoginPageComponent extends LitElement {
         </h3></oryx-heading
       >
       <oryx-auth-login
-        .options="${{ enableRememberMe: false }}"
+        .options="${{
+          enableRememberMe: false,
+          passwordVisibility: PasswordVisibilityStrategy.Click,
```
tobi-or-not-tobi, line 19: Can we change this in the defaultOptions of the component. I actually think it's the best standard option, though some security folks won't like it. It's configurable so that's fine.

tobi-or-not-tobi (changes requested):

libs/domain/picking/src/components/login/login.component.ts
aleespryker: Looks like it already is the default. For some reason, it was on mousedown before, but now netlify prod is using the click strategy anyways. I think we can close this.

aleespryker:
password visibility strategy is already click by default.

PR #357, opened 2023-05-22
feat: add icon list into schema

supproduction:
splits data-clit into revealers
adds icon list into schema

closes [HRZ-2889](https://spryker.atlassian.net/browse/HRZ-2889)

[HRZ-2889]: https://spryker.atlassian.net/browse/HRZ-2889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/content/link/src/link.schema.ts
```diff
@@ -69,28 +70,11 @@ export const linkComponentSchema: ContentComponentSchema<ContentLinkComponent> =
...
-          {
-            value: 'mark',
-            text: 'mark',
-          },
-        ],
+        options:
+          iconInjectable
+            .get()
+            ?.getIcons()
+            .map((i) => ({ value: i, text: i })) ?? [],
```
tobi-or-not-tobi, line 78: what about sorting them?

libs/domain/search/src/services/revealers/products-experience-data.revealer.ts
```diff
@@ -0,0 +1,34 @@
...
+  ExperienceDataRevealer,
+  MessageType,
+  postMessage,
+} from '@spryker-oryx/experience';
+import { Observable, of, switchMap, tap } from 'rxjs';
+import { Suggestion } from '../../models';
+
+import { SuggestionService } from '../suggestion';
+
+export class ProductsExperienceDataRevealer implements ExperienceDataRevealer {
```
tobi-or-not-tobi, line 13: I've tried to use the product sku editor inside EB and it no longer works. not sure if it's related to this PR though.

libs/domain/search/src/services/search.providers.ts
```diff
@@ -44,3 +46,10 @@ export const searchProviders: Provider[] = [
     useClass: SearchPageTitleMetaResolver,
   },
 ];
+
+export const searchPreviewProviders: Provider[] = [
+  {
+    provide: ExperienceDataRevealer,
```
tobi-or-not-tobi, line 52: I prefer to not hang the revealer up on Experience Builder, as other integrations can benefit from this as well.

tobi-or-not-tobi: I do like this revealer pattern though!

tobi-or-not-tobi (changes requested):

libs/domain/search/src/services/revealers/products-experience-data.revealer.ts
supproduction: It's not releated to this PR. we have bug in EB

tobi-or-not-tobi (approved):

PR #358, opened 2023-05-23
feat: implement push notification handling

tarashaliuk:
closes: [HRZ-2841](https://spryker.atlassian.net/browse/HRZ-2841)

[HRZ-2841]: https://spryker.atlassian.net/browse/HRZ-2841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/offline/feature.ts
```diff
@@ -30,13 +45,30 @@ export class OfflinePickingFeature implements AppFeature {
         PickingSyncAction,
         PickingSyncActionHandlerService
       ),
+      {
+        provide: PushProvider,
+        useFactory: () =>
+          new WebPushProvider({
+            applicationServerKey:
+              'BGqNWbv0hWM5CQ1-KwAfSQBMC6TMVFyrnh3vQp37oGCNvQ6eG_HyMjxBFJRWeCPTbzDoxcjhxLJS8Ck8r1G2oFw',
```
nsilva-spryker: Maybe we should put this on environment variable

nsilva-spryker (changes requested):
Left review
(Update branch with dev, that fix this issue: https://spryker.atlassian.net/browse/HRZ-2928)

apps/fulfillment-e2e/src/support/index.ts
```diff
@@ -10,3 +10,14 @@
...
 import './commands';
+
+Cypress.on('uncaught:exception', (err) => {
+  const ignoreErrors = [
+    'Registration failed - push service not available',
+    'Background Sync is disabled',
+  ];
+
+  if (ignoreErrors.some((ignoreError) => err.message.includes(ignoreError))) {
+    return false;
```
gund, line 21: Why are push errors ignored in E2E?

apps/fulfillment/mock/server.js
```diff
@@ -37,6 +37,34 @@ exports.createMockServer = function createMockServer() {
...
+          },
+          group: {
+            name: 'warehouse',
+            identifier: 'e84b3cb8-a94a-5a7e-9adb-cc5353f7a73f',
+          },
+        },
+        links: {
+          self: '/push-notification-subscriptions/adf77c5b-17c6-5aa0-bc87-50558d79b03b',
+        },
+      },
```
gund: This data should be sent by the app to this endpoint but the response should be different one from the BAPI.
Also please move it to the JSON file from here so we do not pollute the implementation code.

apps/fulfillment/mock/server.js
```diff
@@ -129,7 +157,14 @@ function createProxyRouter(proxyUrl, proxyRoutes, basePath) {
       xfwd: true,
       logLevel: 'debug',
       selfHandleResponse: true,
-      onProxyReq: fixRequestBody,
+      onProxyReq: (proxyReq, req, res) => {
+        if (req.url === '/push-notification-subscriptions') {
+          proxyReq.setHeader('Content-Type', 'application/vnd.api+json');
+          proxyReq.setHeader('Accept-Language', 'en');
```
gund: These headers must be set by the application and not proxy server, please move it.

libs/domain/picking/offline/feature.ts
```diff
@@ -9,19 +14,29 @@ import {
...
+  BapiPushNotificationService,
   PickingListOfflineAdapter,
   PickingListOnlineAdapter,
   PickingListOnlineDefaultAdapter,
   PickingSyncAction,
   PickingSyncActionHandlerService,
 } from './services';
+import { BapiPushNotificationDefaultAdapter } from './services/adapter/bapi-push-notification-default.adapter';
+import { BapiPushNotificationAdapter } from './services/adapter/bapi-push-notification.adapter';
+import { BapiPushNotificationDefaultService } from './services/bapi-push-notification-default.service';
```
gund: Please re-export these from `./services` barrel file so they are all imported from a single location.

libs/domain/picking/offline/feature.ts
```diff
@@ -30,13 +46,31 @@ export class OfflinePickingFeature implements AppFeature {
...
+        useClass: BapiPushNotificationDefaultAdapter,
+      },
       { provide: PickingListService, useClass: PickingListDefaultService },
       { provide: PickingListAdapter, useClass: PickingListOfflineAdapter },
       {
         provide: PickingListOnlineAdapter,
         useClass: PickingListOnlineDefaultAdapter,
       },
       { provide: PickingHttpService, useClass: PickingHttpDefaultService },
+      { provide: StorageService, useClass: DefaultStorageService },
```
gund: This is already provided by the core feature, why do you need it here?

libs/domain/picking/offline/push-notification-plugin.ts
```diff
@@ -0,0 +1,31 @@
...
+        .isAuthenticated()
+        .pipe()
+        .subscribe((isAuthenticated) => {
+          if (isAuthenticated) {
+            pushNotificationService.initSubscription().subscribe(() => {
+              console.log('Push notifications subscribed');
+            });
+          } else {
+            pushNotificationService.unsubscribe().subscribe(() => {
+              console.log('Unsubscribed from push notifications');
```
gund: Remove logs?

libs/domain/picking/offline/services/adapter/bapi-push-notification-default.adapter.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(protected pickingHttpService = inject(PickingHttpService)) {}
+  sendSubscription(subscription: PushSubscriptionJSON): Observable<void> {
+    return this.pickingHttpService.post('/push-notification-subscriptions', {
+      data: {
+        type: 'push-notification-subscriptions',
+        attributes: {
+          providerName: 'web-push-php',
+          group: {
+            name: 'warehouse',
+            identifier: 'e84b3cb8-a94a-5a7e-9adb-cc5353f7a73f',
```
gund: Please leave a TODO comment to update warehouse info once we ingrate this feature with BAPI.

libs/domain/picking/offline/services/bapi-push-notification-default.service.ts
```diff
@@ -0,0 +1,53 @@
...
+          // catch error when subscription already exists
+          catchError((error: any) => {
+            if (
+              error.body.errors[0].status === 400 &&
+              error.body.errors[0].code === 5003
+            ) {
+              return of(undefined);
+            }
+
+            this.unsubscribe();
```
gund: This stream is not subscribed so it's not going to execute properly.
You should instead return it and then you can map it to error so it's executed.

libs/domain/picking/offline/services/picking-sync-action-handler.service.ts
```diff
@@ -59,4 +69,35 @@ export class PickingSyncActionHandlerService
...
+      );
+    }
+
+    return this.onlineAdapter.get({ ids: sync.payload.ids }).pipe(
+      combineLatestWith(this.indexedDbService.getStore(PickingListEntity)),
+      switchMap(async ([pickingLists, store]) => {
+        // TODO: Remove filtering when BE will be able to retrieve the list of picking lists by provided "ids" list
+        const filteredPickingLists = pickingLists.filter(({ id }) =>
+          sync.payload.ids.includes(id)
+        );
```
gund: This filtering must be part of the adapter and not here.

libs/domain/picking/offline/services/picking-sync-action-handler.service.ts
```diff
@@ -59,4 +69,35 @@ export class PickingSyncActionHandlerService
...
+
+        const addedOrUpdatedKeys = await store.bulkPut(filteredPickingLists, {
+          allKeys: true,
+        });
+
+        if (addedOrUpdatedKeys.length !== filteredPickingLists.length) {
+          throw new Error(
+            `PickingSyncActionHandlerService: Could not add or update all PickingLists after push!`
```
gund: Let's add as much info as possible to error messages:
```suggestion
            `PickingSyncActionHandlerService: Could not save ${addedOrUpdatedKeys.length} out of ${filteredPickingLists.length} PickingLists after push!`
```

libs/domain/picking/src/models/picking-list-qualifier.ts
```diff
@@ -8,6 +8,7 @@ export interface PickingListQualifier
     PageableQualifier,
     SearchQualifier {
   id?: string;
+  ids?: string[];
```
gund: Just replace `id` prop with `ids` we do not need both.

libs/domain/picking/offline/push-notification-plugin.ts
```diff
@@ -0,0 +1,31 @@
...
+      const injector = app!.requirePlugin(InjectionPlugin).getInjector();
+
+      const authService = injector.inject(OauthService);
+      const pushNotificationService = injector.inject(
+        BapiPushNotificationService
+      );
+
+      authService
+        .isAuthenticated()
+        .pipe()
```
Sinucid, line 17: `pipe` is useless here

libs/domain/picking/offline/sw-feature.ts
```diff
@@ -0,0 +1,11 @@
+import { AppFeature, AppInitializer } from '@spryker-oryx/core';
+import { SwPushInitializerService } from './services';
+
+export class SwOfflinePickingFeature implements AppFeature {
```
gund: This feature should extend `OfflinePickingFeature` and override plugins prop to remove them.
Then in the main SW feature you should only use `SwOfflinePickingFeature` without a hacking of `OfflinePickingFeature` currently done there.

libs/domain/picking/offline/services/adapter/bapi-push-notification-default.adapter.ts
Sinucid, line 19: I believe identifier should be configurable

gund (changes requested):

libs/domain/picking/offline/services/picking-sync-action-handler.service.ts
Sinucid: @gund Should we schedule another try for the failed ones?

libs/domain/picking/offline/services/sw-push-initializer.service.ts
```diff
@@ -0,0 +1,29 @@
...
+import { firstValueFrom, Observable } from 'rxjs';
+
+declare let self: ServiceWorkerGlobalScope;
+
+export class SwPushInitializerService implements AppInitializer {
+  constructor(private syncSchedulerService = inject(SyncSchedulerService)) {}
+
+  initialize(): void | Observable<void> | Promise<void> {
+    self.addEventListener('push', (event: PushEvent) => {
+      const payload: PushSyncPayload = event.data?.json();
```
Sinucid: Here is optional chaining in event data. Can be a case when there is not data? If so - should we schedule such sync?

libs/platform/offline/src/services/sync-scheduler-default.service.ts
```diff
@@ -15,12 +15,17 @@ import { SyncSchedulerService } from './sync-scheduler.service';
...
 export class SyncSchedulerDefaultService implements SyncSchedulerService {
   protected scheduleSyncTimer?: ReturnType<typeof setTimeout>;
 
   constructor(
     protected indexedDbService = inject(IndexedDbService),
-    protected serviceWorker = navigator.serviceWorker
+    // navigator.serviceWorker is available only in main thread and this service is running in SW thread
+    protected serviceWorker:
+      | ServiceWorkerContainer
+      | undefined = navigator.serviceWorker
```
Sinucid: What about:

```
protected async getServiceWorker() {
    return navigator.serviceWorker ? await navigator.serviceWorker.ready : self.registration:
}

```

and then:

`const sync = (await this.getServiceWorker()).sync;`

Sinucid (changes requested):

apps/fulfillment-e2e/src/support/index.ts
tarashaliuk: After the call with @AndreyYevtukhov we decided to extract e2e tests into [separate task](https://spryker.atlassian.net/browse/HRZ-2946) because we didn't find a proper way to "Allow" push notifications in headless mode.

libs/domain/picking/src/models/picking-list-qualifier.ts
tarashaliuk: But currently, BAPI is not supporting `ids`. I will left TODO above the `id?: string;` to remove it when BE will support `ids`

libs/domain/picking/offline/services/adapter/bapi-push-notification-default.adapter.ts
tarashaliuk: Added TODO as Alex suggested in prev review

libs/domain/picking/offline/push-notification-plugin.ts
Sinucid: still here, but ok, I will drop it in my pr

libs/domain/picking/offline/services/sw-push-initializer.service.ts
Sinucid:
```suggestion
      const payload: PushSyncPayload = event.data.json();
```

Sinucid (changes requested):

Sinucid (dismissed):

libs/domain/picking/offline/services/picking-sync-action-handler.service.ts
gund: No, background sync will automatically do that if the promise is rejected.

libs/domain/picking/src/models/picking-list-qualifier.ts
gund: We discussed with Taras that for now we should implement filtering by ids in the adapter and once BAPI supports it we will just update adapter code.
So we do not need both `id` and `ids` only the `ids` prop.

apps/fulfillment/mock/db.json
```diff
@@ -979,5 +979,30 @@
...
+          "expirationTime": null,
+          "keys": {
+            "p256dh": "p256dh",
+            "auth": "auth"
+          }
+        },
+        "group": {
+          "name": "warehouse",
+          "identifier": "identifier"
+        }
```
gund, line 1001: This is a payload that FE sends, please replace this with a response that BAPI sends back.

gund (changes requested):
A few comments were not resolved still.

apps/fulfillment/mock/db.json
tarashaliuk: No, it is the response that BAPI sends back. Payload is a bit different.
image

image

libs/domain/picking/offline/services/adapter/picking-list-offline.adapter.ts
```diff
@@ -45,9 +45,8 @@ export class PickingListOfflineAdapter implements PickingListAdapter {
             'readonly',
             [pickingListStore, productStore],
             async () => {
-              if (qualifier.id) {
-                const data = await pickingListStore.get(qualifier.id);
-
+              if (qualifier.ids?.length) {
+                const data = await pickingListStore.get(qualifier.ids[0]);
```
gund: You should use `bulkGet` to get array of picking-lists.

gund (changes requested):

gund (dismissed):

nsilva-spryker (approved):

PR #359, opened 2023-05-23
fix: provide product contenxt

supproduction:
fixes providing product context


libs/domain/product/card/src/card.component.ts
```diff
@@ -42,8 +48,15 @@ export class ProductCardComponent extends ProductMixin(
     })
   );
 
+  protected context = new ContextController(this);
+
+  skuController = effect(() => {
```
tobi-or-not-tobi: protected?

tobi-or-not-tobi (dismissed):

tobi-or-not-tobi (approved):

PR #367, opened 2023-05-24
fix: sw auth

Sinucid:
Required [358](https://github.com/spryker/oryx/pull/358) is being merged first

closes to [HRZ-2916](https://spryker.atlassian.net/browse/HRZ-2916)

[HRZ-2916]: https://spryker.atlassian.net/browse/HRZ-2916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/offline/push-notification-plugin.ts
```diff
@@ -12,20 +12,17 @@ export class PushNotificationPlugin extends ExecPlugin {
...
-          }
-        });
+      authService.isAuthenticated().subscribe((isAuthenticated) => {
+        if (isAuthenticated) {
+          pushNotificationService.initSubscription().subscribe(() => {
+            console.log('Push notifications subscribed');
+          });
+        } else {
+          pushNotificationService.unsubscribe().subscribe(() => {
+            console.log('Unsubscribed from push notifications');
```
gund: Please remove logs.

libs/platform/auth/src/oauth/oauth.service.spec.ts
```diff
@@ -306,5 +298,44 @@ describe('OauthService', () => {
...
+              provide: OauthServiceConfig,
+              useValue: {
+                ...mockOauthServiceConfig,
+                defaultProvider,
+              },
+            },
+          ],
+        });
+
+        getService().loginWith = vi.fn();
```
gund: Please use [`spyOn()` API](https://vitest.dev/guide/mocking.html#functions) for spying on methods instead of overriding them:
```suggestion
        spyOn(getService(), 'loginWith');
```

libs/platform/core/src/services/storage/default-storage.service.ts
```diff
@@ -1,36 +1,54 @@
...
+import { IndexedStorage, StorageType, StoredValue } from './model';
 import { StorageService } from './storage.service';
 
 export class DefaultStorageService implements StorageService {
-  protected getStorage(type = StorageType.LOCAL): Storage {
-    return type === StorageType.LOCAL ? localStorage : sessionStorage;
-  }
+  constructor(protected defaultStorageType = StorageType.Local) {}
+
+  protected ibdStorage = new IndexedDbStorage();
```
gund: You should use DI to get instances instead of manual object creation.

libs/platform/core/src/services/storage/indexed-db-storage.ts
```diff
@@ -0,0 +1,56 @@
...
+        liveQuery<StoredValue>(async () => {
+          return (await db.table(indexedDbTableName).get(key))?.value;
+        })
+      )
+    );
+  }
+
+  setItem(key: string, value: string): void {
+    this.db.subscribe(
+      async (db) => await db.table(indexedDbTableName).put({ key, value })
```
gund: This async/await are not needed as nobody can listen it so please drop it to a simple sync function.

libs/platform/core/src/services/storage/indexed-db-storage.ts
```diff
@@ -0,0 +1,56 @@
...
+    subscriber.next(db);
+
+    return () => {
+      db.close();
+    };
+  }).pipe(
+    switchMap(async (db) => {
+      await db.open();
+
+      return db;
```
gund: You can return the table here so you don't have to do it every time in other places:
```suggestion
      return db.table(indexedDbTableName);
```

libs/platform/auth/src/oauth/oauth.service.ts
```diff
@@ -115,6 +117,13 @@ export class OauthService implements AuthService, AuthTokenService {
     );
   }
 
+  invokeStoredToken(): void {
+    this.storageService
+      .get<OauthServiceState>(OauthService.STATE_KEY)
+      .pipe(distinctUntilChanged())
+      .subscribe((state) => this.state$.next({ ...state }));
+  }
```
gund: Please invoke this via `AppInitializer` feature or in constructor directly instead of manually in the `SwAuthPlugin` to remove unnecessary code.

gund (changes requested):

libs/template/presets/src/fulfillment/offline.ts
```diff
@@ -67,7 +66,7 @@ export function offlineServiceWorkerFulfillmentFeatures(
     new IndexedDbFeature(config?.indexedDb),
     new RouterFeature(),
     new (class extends BapiAuthFeature {
```
gund:
```suggestion
    new BapiAuthFeature()
```

libs/platform/core/src/services/storage/default-storage.service.ts
```diff
@@ -1,12 +1,13 @@
...
-import { IndexedStorage, StorageType, StoredValue } from './model';
+import { IndexedDBStorageService } from './indexed-db-storage.service';
+import { StorageType, StoredValue } from './model';
 import { StorageService } from './storage.service';
 
 export class DefaultStorageService implements StorageService {
   constructor(protected defaultStorageType = StorageType.Local) {}
 
-  protected ibdStorage = new IndexedDbStorage();
+  protected ibdStorage = resolve(IndexedDBStorageService);
```
gund: Why not using `inject()` API?

libs/domain/picking/offline/sw-feature.ts
```diff
@@ -1,15 +1,9 @@
...
-import { SwPushInitializerService } from './services';
 
-export class SwOfflinePickingFeature extends OfflinePickingFeature {
-  plugins = [];
-
-  providers = [
-    ...super.getProviders(),
-    {
-      provide: AppInitializer,
-      useClass: SwPushInitializerService,
```
gund: Why did you remove `SwPushInitializerService` from providers? It's not going to initialize anymore the service.

libs/domain/picking/offline/services/sw-push-initializer.service.ts
```diff
@@ -13,6 +14,9 @@ export class SwPushInitializerService implements AppInitializer {
   constructor(private syncSchedulerService = inject(SyncSchedulerService)) {}
 
   initialize(): void | Observable<void> | Promise<void> {
+    //make subscription on auth token
+    resolve(OauthService).invokeStoredToken();
```
gund: Why not `inject()` it in the constructor?

gund (changes requested):

libs/domain/picking/offline/sw-feature.ts
Sinucid: This is not my change, it's from source Taras's branch. He moved it to the offline feature

libs/template/presets/src/fulfillment/offline.ts
Sinucid: But we need to drop `BapiAuthPlugin` for SW or I missed something?

libs/domain/picking/offline/services/sw-push-initializer.service.ts
Sinucid: I'm not sure that there is a big difference, due to we make resolving only once, but ok

gund: It's mostly a convention to have all dependencies in the class constructor =)

libs/template/presets/src/fulfillment/offline.ts
gund: Ah yes, let's then have this in a separate class/file, you can call it `SwBapiAuthPlugin` which removed the plugins.

gund (changes requested):

gund (dismissed):

gund (dismissed):

apps/fulfillment-e2e/cypress.config.js
```diff
@@ -23,5 +23,7 @@ module.exports = defineConfig({
   },
   env: {
     ORYX_FULFILLMENT_BACKEND_URL: process.env.ORYX_FULFILLMENT_BACKEND_URL,
+    ORYX_FULFILLMENT_PUBLIC_VAPID_KEY:
```
AndreyYevtukhov: Why do we need this variable here? (We have discussed it with @tarashaliuk and it should have been removed)

apps/fulfillment-e2e/src/support/commands.ts
```diff
@@ -109,3 +103,35 @@ Cypress.Commands.add('mockPickingInProgress', () => {
...
   });
 });
+
+async function indexedDbExists(dbName) {
+  const dbs = await indexedDB.databases();
+  if (!dbs.map((db) => db.name).includes(dbName)) {
+    throw 'DB not found';
+  }
+
+  console.log('DB found');
```
AndreyYevtukhov: let's also log db name here (it will make debug easier in the future)

libs/platform/core/src/services/storage/default-indexed-db-storage.service.spec.spec.ts
```diff
@@ -0,0 +1,137 @@
+import { createInjector, destroyInjector, getInjector } from '@spryker-oryx/di';
```
AndreyYevtukhov: This file name has a `spec` word duplication - should we remove it?

AndreyYevtukhov (changes requested):
Small comments left

apps/fulfillment-e2e/cypress.config.js
Sinucid: Probably some issues after merging to the dev Taras's branch. This branch one was created based on his one. Dropped

AndreyYevtukhov (approved):

PR #372, opened 2023-05-25
feat: fa filters e2es

Sinucid:
closes to [HRZ-2865](https://spryker.atlassian.net/browse/HRZ-2865)

[HRZ-2865]: https://spryker.atlassian.net/browse/HRZ-2865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/fulfillment-e2e/src/integration/filters.cy.ts
```diff
@@ -0,0 +1,83 @@
...
+
+const pickingListsFragment = new PickingListsFragment();
+const filtersFragment = new FiltersFragment();
+
+describe('When user opens the filters', () => {
+  const largeAndEarliestOrderFirst = (exists = true) =>
+    pickingListsFragment
+      .getPickingListsItems()
+      .eq(0)
+      .find('.total oryx-icon-button button')
```
AndreyYevtukhov: All selectors should exist only in page objects or page fragments. 

When we put selectors in spec files - we violate that pattern + we make our tests less stable and increase maintenance time and costs

apps/fulfillment-e2e/src/integration/filters.cy.ts
```diff
@@ -0,0 +1,83 @@
+import { FiltersFragment } from '../support/page_fragments/filters.fragment';
+import { PickingListsFragment } from '../support/page_fragments/picking-lists.fragment';
+
+const pickingListsFragment = new PickingListsFragment();
+const filtersFragment = new FiltersFragment();
+
+describe('When user opens the filters', () => {
+  const largeAndEarliestOrderFirst = (exists = true) =>
```
AndreyYevtukhov:
```
//should not change the order
largeAndEarliestOrderFirst();
```

can we rename it to show exactly what it does? like `checkOrderSorting(isChanged: boolean)` or something?

Also, we can move this function out from the `describe` block to make it visible to everyone

apps/fulfillment-e2e/src/integration/filters.cy.ts
```diff
@@ -0,0 +1,83 @@
...
+  beforeEach(() => {
+    cy.clearIndexedDB();
+    cy.login();
+
+    filtersFragment.getFilterButtonTrigger().click();
+  });
+
+  it('should apply correct sorting configuration', () => {
+    //should open the filters modal
+    filtersFragment.getFiltersModal().should('have.attr', 'open');
```
AndreyYevtukhov: `.should('have.attr', 'open');` -> `.should('be.visible);`

A visibility check is better here because it is "user" centred and checks if block is ACTUALLY visible to user

`open` attribute check is not a prove of visibility. Element might have it, but still be not visible to the end user

AndreyYevtukhov: You actually can read more about visibility check here https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Visibility, it is quite complex, but everything there makes sense :)

apps/fulfillment-e2e/src/integration/filters.cy.ts
```diff
@@ -0,0 +1,83 @@
...
+    //should have default selected sorting option
+    filtersFragment.getFiltersModal().find('input').eq(0).should('be.checked');
```
AndreyYevtukhov: `filtersFragment.getFiltersModal().find('input')` -> `find` has a selector as a parameter, so should be moved in page object / page fragment

apps/fulfillment-e2e/src/integration/filters.cy.ts
```diff
@@ -0,0 +1,83 @@
...
+    filtersFragment.getFilterButtonTrigger().should('not.be.checked');
+
+    //should reset the form
+    filtersFragment.getFiltersModal().find('input').eq(0).should('be.checked');
+
+    //should not change the order
+    largeAndEarliestOrderFirst();
+
+    //When sorting is applied
+    [false, true, false].forEach((expect, index) => {
```
AndreyYevtukhov: It seems to me that we have a lot of checks in 1 test 

As far as I see in https://spryker.atlassian.net/browse/HRZ-2777 we should have at least 3:

1.  Sorting cancelling 
2. Successful sorting application
3. Sorting reset
4. + I'd say we should add 1 more to check if default sorting is applied

In this case, tests will be smaller, easier to maintain and it will be easier to understand what exactly does not work

apps/fulfillment-e2e/src/support/page_fragments/filters.fragment.ts
```diff
@@ -0,0 +1,19 @@
+export class FiltersFragment {
+  getFilterButton = () => cy.get('oryx-picking-filter-button');
+  getFilterButtonTrigger = () =>
+    this.getFilterButton().find('oryx-toggle-icon input');
+  getFiltersModal = () =>
+    this.getFilterButton().find('oryx-picking-filters').find('oryx-modal');
+  getFiltersReset = () =>
```
AndreyYevtukhov: `getFiltersReset` -> `getFiltersResetButton` or `getFiltersResetBtn`

the same is true for Modals, Inputs, Checkboxes etc. It is a good practice to use HTML element tag in getter because it will increase tests readability

 Just compare `getFiltersReset().click()` with `getFiltersResetButton().click()`

libs/domain/picking/src/components/filter-button/filter-button.def.ts
```diff
@@ -4,4 +4,6 @@ export const filterButtonComponent = componentDef({
   name: 'oryx-picking-filter-button',
   impl: () =>
     import('./filter-button.component').then((m) => m.FilterButtonComponent),
+  schema: () =>
+    import('./filter-button.schema').then((m) => m.filterButtonComponentSchema),
```
AndreyYevtukhov, line 8: I wonder about the name here. Why is it a `filter-button` and not a `sorting` button? 

The task description is about sorting + it is `Sort` on UI. It is possible that we will have inconsistency in the terminology used in the future? (Maybe we have to update `sort` in UI to `filter` as well?)

AndreyYevtukhov (changes requested):
GJ, but can be improved ;)

libs/domain/picking/src/components/filter-button/filter-button.def.ts
Sinucid: This is true. According the designs it will be sort + filters modal in the future: https://www.figma.com/file/OgvBbM6xRMVfy6VYouxONy/PA-Spryker?type=design&node-id=1598-27429&t=cLDDDXFRvIQeG6wb-0

apps/fulfillment-e2e/src/integration/filters.cy.ts
Sinucid: Done. Default configuration is checking as a part of the first test case

AndreyYevtukhov (approved):
Nice one!

PR #374, opened 2023-05-25
fix: failing fa tests

AndreyYevtukhov:
fixes falling e2e tests for FA


apps/fulfillment-e2e/src/support/page_fragments/picking-in-progress-modal.fragment.ts
```diff
@@ -1,6 +1,7 @@
 export class PickingInProgressModalFragment {
   getWrapper = () => cy.get('oryx-picking-in-progress-modal');
 
+  getProceedToPickingBtn = () => cy.contains('Proceed to picking');
```
nsilva-spryker: Should we get the proper button, instead of via text? This could change in the future.

nsilva-spryker (dismissed):

Sinucid (approved):

PR #384, opened 2023-05-28
refactor: neutral colors

tobi-or-not-tobi:
Neutral colors.


docs/drafts/design-system/color-mapping.md
```diff
@@ -0,0 +1,34 @@
+## Neutrals
+
+| Old         | New        |
+| ----------- | ---------- |
+| canvas-100  | neutral-1  |
+| canvas-200  | neutral-3  |
+| canvas-300  | neutral-4  |
+| canvas-400  | neutral-5  |
+| canvas-500  | neutral-6  |
```
Sinucid, line 9: Hmm, not really clear. For some colors primary has `9` index, for some `4`. Is here some reason not to reserve `9` for all of them?

libs/base/ui/form/input-list/src/input-list.styles.ts
```diff
@@ -15,7 +15,7 @@ export const styles = css`
 
     display: none;
     margin-block-end: 8px;
-    color: var(--oryx-color-ink);
+    color: var(--oryx-color-neutral-12);
```
Sinucid: looks like `inkColor`

docs/drafts/design-system/color-mapping.md
tobi-or-not-tobi: It's because of the wrong decision we made in the past to separate canvas and neutral

libs/base/ui/form/input-list/src/input-list.styles.ts
tobi-or-not-tobi: I have dropped the color all together, so that the ink color is inherited. I've also left a note on the Figma to get better inputs on the color (think it should be ink, but it's not clear)

Sinucid (dismissed):

Sinucid (approved):

Sinucid (approved):

PR #399, opened 2023-05-31
fix: product description hydration

Sinucid:
- fixed prehydration directive of oryx-text component to make sure that its entire content is accessible when it is applied
- oryx-text was refactored a bit and fixed bug with truncation of short texts 

closes to [HRZ-2939](https://spryker.atlassian.net/browse/HRZ-2939)

[HRZ-2939]: https://spryker.atlassian.net/browse/HRZ-2939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tobi-or-not-tobi (dismissed):

dunqan (dismissed):

libs/base/ui/structure/text/prehydrate.ts
```diff
@@ -0,0 +1,34 @@
...
+      if (linesCount <= lineClampValue) {
+        return;
+      }
+
+      const isDefaultExpanded = host.hasAttribute('defaultExpanded');
+      host.toggleAttribute('truncated', !isDefaultExpanded);
+      host.toggleAttribute('truncation', true);
+    }
+
+    container.removeEventListener('slotchange', handler);
```
dunqan, line 30: shouldn't it be 
```suggestion
    slot?.removeEventListener('slotchange', handler);
```
?

dunqan (approved):

libs/base/ui/structure/text/prehydrate.ts
Sinucid: Misunderstanding happened, obsolete

PR #400, opened 2023-05-31
feat: Product availability component

tobi-or-not-tobi:
The product availability is exposed in 3 states:
- Available
- Low availability
- Not available

The component can be configured with the following options:
- threshold – the threshold when the availability becomes `LowStock`
- enable Indicator (using the swatch UI component)
- enable exact count – show the number of available items after the text (e.g. "Available online (10)")

closes [HRZ-2974](https://spryker.atlassian.net/browse/HRZ-2974)

[HRZ-2974]: https://spryker.atlassian.net/browse/HRZ-2974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/graphical/swatch/src/swatch.component.spec.ts
```diff
@@ -10,17 +11,73 @@ useComponent(swatchComponent);
 describe('SwatchComponent', () => {
   let element: SwatchComponent;
 
-  beforeAll(async () => {
-    element = await fixture(
-      html`<oryx-swatch .color=${mockColor}></oryx-swatch>`
-    );
-  });
+  describe('when a AlertType is provided', () => {
```
Sinucid:
```suggestion
  describe('when type is provided', () => {
```

libs/base/ui/graphical/swatch/src/swatch.component.ts
```diff
@@ -6,6 +7,12 @@ import { swatchBaseStyle } from './swatch.styles';
 export class SwatchComponent extends LitElement implements SwatchAttributes {
   static styles = swatchBaseStyle;
 
+  @property()
+  set type(value: AlertType) {
```
Sinucid, line 11:
```
this.color = `var(--oryx-color-${value}-9)`;
```

libs/domain/product/availability/availability.component.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+import { customElement, property } from 'lit/decorators.js';
+import { of } from 'rxjs';
+import { ProductAvailabilityComponent } from './availability.component';
+import { productAvailabilityComponent } from './availability.def';
+
+class MockProductService implements Partial<ProductService> {
+  get = vi.fn();
+}
+
+@customElement('oryx-swatch')
```
Sinucid: Why do you need to mock the component? Just check the property instead of attributes:

```
const getSwatch = () => element.renderRoot.querySelector<SwatchComponent>('oryx-swatch');
 
it('should render oryx-swatch with type success', () => {
        expect(getSwatch()?.type).toBe(AlertType.Success);
});
```

libs/domain/product/availability/availability.component.ts
```diff
@@ -0,0 +1,80 @@
...
+@defaultOptions({ threshold: 5, enableIndicator: true })
+export class ProductAvailabilityComponent extends ProductMixin(
+  ContentMixin<CartItemAvailabilityOptions>(LitElement)
+) {
+  static styles = availabilityStyles;
+
+  protected status = computed(() => {
+    const { isNeverOutOfStock, quantity, availability } =
+      this.$product()?.availability ?? {};
+    const { threshold: threshold = Infinity } = this.$options();
```
Sinucid: `const { threshold = Infinity } = this.$options();`
and whether it is reasonable to use Infinity as default value?

libs/domain/product/availability/availability.component.ts
```diff
@@ -0,0 +1,80 @@
...
+  });
+
+  protected override render(): TemplateResult | void {
+    return html`${this.renderIndicator()}${this.renderText()}`;
+  }
+
+  protected renderIndicator(): TemplateResult | void {
+    if (!this.$options().enableIndicator) return;
+
+    const stock = this.status();
```
Sinucid:
```
switch(this.status()) {
    case StockAvailability.InStock:
         return html`<oryx-swatch .type=${AlertType.Success}></oryx-swatch>`;
    case StockAvailability.LowStock:
         return html`<oryx-swatch .type=${AlertType. Warning}></oryx-swatch>`;
    default:
         return html`<oryx-swatch .type=${AlertType. Error}></oryx-swatch>`;
}
```

libs/domain/product/availability/availability.component.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  protected renderText(): TemplateResult | void {
+    const stock = this.status();
+    const availableQuantity = this.$product()?.availability?.quantity;
+
+    if (stock === StockAvailability.OutOfStock)
+      return html`${i18n('product.availability.none')}`;
+
+    if (this.$options().enableExactStock && availableQuantity !== undefined) {
```
Sinucid:
```suggestion
    if (this.$options().enableExactStock && availableQuantity) {
```

libs/domain/product/availability/availability.model.ts
```diff
@@ -0,0 +1,11 @@
+export interface CartItemAvailabilityOptions {
+  threshold?: number;
```
Sinucid, line 6: some doc?

Sinucid (changes requested):

libs/domain/product/availability/availability.component.spec.ts
tobi-or-not-tobi: because the type is not reflected

Sinucid: Reflect works for the attribute, but prop should be presented all time, even in not-defined custom element

libs/domain/product/availability/availability.model.ts
Sinucid: And what about short descriptions for the rest options?

Sinucid (changes requested):

libs/domain/product/availability/availability.component.spec.ts
tobi-or-not-tobi: actually, no need to do it. apparently, when the element is not registered, lit will reflect it as an attribute by default anyway.

Sinucid: Yes, in case of standard attribute binding ie `attr="value"`, but not dot binding `.attr="value"`

Sinucid (dismissed):

Sinucid (approved):

libs/domain/product/availability/availability.component.spec.ts
tobi-or-not-tobi: actually it does. When there's no registered version of the element, the attributes can be queried. It actually makes sense, as the element can be registered later on, and is this case the property would be lost.

PR #401, opened 2023-05-31
refactor: use signals in user components

tobi-or-not-tobi:
Refactor address mixin and move to signals

closes [HRZ-2986](https://spryker.atlassian.net/browse/HRZ-2986)

[HRZ-2986]: https://spryker.atlassian.net/browse/HRZ-2986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/user/address-list/src/address-list.component.ts
+++ b/libs/domain/user/address-list/src/address-list.component.ts
@@ -33,7 +33,7 @@ export class AddressListComponent extends AddressMixin(
   protected selectedAddressId?: string;
 
   protected autoSelectAddress = effect(() => {
-    const addresses = this.addresses();
+    const addresses = this.$addresses();
     if (
       this.$options().selectable &&
       !addresses?.find((address) => address.id === this.selectedAddressId)
@@ -45,7 +45,7 @@ export class AddressListComponent extends AddressMixin(
   });
 
   protected override render(): TemplateResult | void {
-    const addresses = this.addresses();
+    const addresses = this.$addresses();
     if (!addresses?.length) return this.renderEmpty();
 
     return html`
@@ -90,7 +90,7 @@ export class AddressListComponent extends AddressMixin(
   }
 
   protected dispatchSelectedAddress(addressId?: string): void {
-    const address = this.addresses()?.find(
+    const address = this.$addresses()?.find(
       (address) => address.id === addressId
     );
     if (address) {
@@ -130,8 +130,8 @@ export class AddressListComponent extends AddressMixin(
   }
 
   protected createKey(addressId?: string): string {
-    return `${addressId}-${this.addresses()?.findIndex(
+    return `${addressId}-${this.$addresses()?.findIndex(
       (a) => a.id === addressId
-    )}-${this.addresses?.length}`;
+    )}-${this.$addresses?.length}`;
   }
 }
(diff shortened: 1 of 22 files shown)


libs/domain/user/address-list/src/address-list.component.ts
```diff
@@ -22,39 +16,37 @@ import { styles } from './address-list.styles';
...
 
-  @asyncState()
-  protected addresses = valueType(resolve(AddressService).getAddresses());
+  protected add = signal(resolve(AddressService).getAddresses());
 
   @state()
   protected selectedAddressId?: string;
 
   protected willUpdate(changedProperties: PropertyValues): void {
+    const addresses = this.addresses();
```
dunqan: Did you consider to move it to an effect? Or it didn't work?

libs/domain/user/src/mixins/address-component.mixin.ts
```diff
@@ -34,30 +27,18 @@ export const AddressMixin = <
...
-
-    @observe()
-    protected addressId$ = new BehaviorSubject(this.addressId);
-
-    @property({ type: Object }) address?: Address;
-
-    @observe('address')
-    protected addressProp$ = new BehaviorSubject(this.address);
+    @signalProperty() addressId?: string;
+    @signalProperty() addressData?: Address;
```
dunqan: I think attribute name is more important than usable property (as it lands in DOM), and we don't have this convention of `somethingData`. Not saying it's wrong, but maybe worth reconsidering...

libs/domain/user/address-list/src/address-list.component.ts
tobi-or-not-tobi: have not considered, good point

libs/platform/experience/src/mixins/content.mixin.ts
```diff
@@ -21,10 +21,24 @@ export declare class ContentMixinInterface<OptionsType, ContentType>
 
   protected data: { options: OptionsType; content: ContentType };
 
+  /**
+   * @deprecated
```
dunqan, line 25: I assume we want to drop those soon? (when finishing refactor I guess) As ideally we won't have deprecared in 1.0.

libs/domain/user/address/src/address.component.ts
```diff
@@ -20,29 +19,20 @@ export class AddressComponent extends AddressMixin(
...
-      }
+    if (multiline)
+      return this.renderMultiline(schema ?? defaultMultilineSchema, address);
 
-      return html`${this.formatLine(
-        options.schema ?? defaultSinglelineSchema,
-        address
-      )}`;
-    })}`;
+    return this.formatLine(schema ?? defaultSinglelineSchema, address);
```
dunqan: this one could be implemented as computed, but not a big deal

dunqan (dismissed):

libs/platform/experience/src/mixins/content.mixin.ts
tobi-or-not-tobi: For sure. I just want a mechanism during the refactors to highlight these thigns

dunqan (approved):

PR #417, opened 2023-06-06
feat: picking header back button logic

aleespryker:
Closes [HRZ-2988](https://spryker.atlassian.net/browse/HRZ-2988)

---

### Checklist before requesting a code review

- [x] PR title is aligned with [PR titles and conventional commits](https://spryker.atlassian.net/wiki/spaces/FA/pages/3821175291/PR+titles+and+conventional+commits)
- [x] I have performed a self-review of my code and have fixed all jobs failures

[HRZ-2988]: https://spryker.atlassian.net/browse/HRZ-2988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/components/picking-header/picking-header.component.ts
```diff
@@ -1,19 +1,37 @@
...
 import { PickingListMixin } from '../../mixins';
+import { DiscardPickingComponent } from '../discard-modal';
 import { styles } from './picking-header.styles';
 
 export class PickingHeaderComponent extends PickingListMixin(LitElement) {
   static styles = styles;
 
   protected routerService = resolve(RouterService);
 
+  protected discardModal = createRef<DiscardPickingComponent>();
```
Sinucid: use `@query` decorator instead ref, please

libs/domain/picking/src/components/picking-header/picking-header.component.ts
```diff
@@ -1,19 +1,37 @@
...
+
   @state() isCartNoteVisible?: boolean;
 
+  @asyncState()
```
Sinucid: use signals, please

libs/domain/picking/src/components/picking-header/picking-header.component.ts
```diff
@@ -55,12 +73,20 @@ export class PickingHeaderComponent extends PickingListMixin(LitElement) {
...
-      ></oryx-site-navigation-item>`;
+      ></oryx-site-navigation-item>
+      <oryx-discard-picking
+        ${ref(this.discardModal)}
+        @oryx.back=${this.back}
+      ></oryx-discard-picking>`;
+  }
+
+  protected openDiscardModal(): void {
+    const modal = this.discardModal.value;
```
Sinucid: when change ref to query, just:
`this.discardModal?.toggleAttribute('open', true)`

libs/domain/picking/src/components/picking-header/picking-header.component.ts
```diff
@@ -55,12 +73,20 @@ export class PickingHeaderComponent extends PickingListMixin(LitElement) {
...
+    modal && (modal.open = true);
   }
 
   protected back(): void {
-    //TODO - display discard modal
-    this.routerService.back();
+    this.routerService.back(this.pickingList?.cartNote ? 2 : 1);
```
Sinucid: looks not very stable. I believe we should find exact step, where we would like to navigate back

libs/template/presets/fulfillment/plugin.ts
```diff
@@ -0,0 +1,59 @@
...
+              (
+                resolve(RouterService).routeGuard() as BehaviorSubject<string>
+              ).next('');
+              resolve(StorageService).remove(
+                CAN_LEAVE_ROUTE,
+                StorageType.SESSION
+              );
+              return;
+            }
+            if (current.startsWith('/picking-list')) {
```
Sinucid: hardcoded route's parts also looks not very stable

Sinucid (dismissed):

libs/domain/picking/src/components/picking-header/picking-header.component.ts
```diff
@@ -56,12 +69,23 @@ export class PickingHeaderComponent extends PickingListMixin(LitElement) {
...
+  }
+
+  protected closeDiscardModal(): void {
+    this.discardModal?.toggleAttribute('open', false);
   }
 
   protected back(): void {
-    //TODO - display discard modal
-    this.routerService.back();
+    this.routerService.back(this.pickingList?.cartNote ? 2 : 1);
```
dunqan: we did discuss that we should not go back 2 or 1 times, as it's unreliable, and maybe not needed, anything changed?

libs/domain/picking/src/routes.ts
```diff
@@ -10,6 +13,9 @@ export const defaultPickingRoutes: RouteConfig[] = [
     path: '/picking-list/picking/:id',
     render: ({ id }) =>
       html`<oryx-picking .pickingListId="${id}" mode-light></oryx-picking>`,
+    leave: (): Observable<boolean> => {
+      return resolve(RouterService).shouldGuardRoute();
```
dunqan: Why we delegate this to router service? 
Shouldn't leave callback of specific have all the info it needs to decide, if we should lave (or not) this route?

libs/platform/router/src/services/router.service.ts
```diff
@@ -13,6 +13,7 @@ export interface RouterService {
   acceptParams(params: RouteParams): void;
   getUrl(route?: string, extras?: NavigationExtras): Observable<string>;
   getPathId(id: string): string | undefined;
+  shouldGuardRoute(): Observable<boolean>;
```
dunqan: I'm wondering why we need this public on the RouterService. 
Ideally router will handle navigation or not inside, depending on the leave calback result, without exposing anything.

dunqan (changes requested):

supproduction (dismissed):

libs/platform/router/lit/lit-router.ts
```diff
@@ -97,12 +125,60 @@ export class LitRouter extends Router {
...
+
+    const location = window.location;
+    if (anchor.origin !== origin) {
+      return;
+    }
+
+    e.preventDefault();
+    if (href !== location.href) {
+      window.history.pushState(
+        { counter: this.routerService.getCounter() + 1 },
```
dunqan: Can't we put conter here? We can leave it for the merging routing followup and unit test coverage.

libs/platform/router/lit/lit-routes.ts
```diff
@@ -0,0 +1,382 @@
...
+      const route = this._getRoute(pathname);
+      if (route === undefined) {
+        throw new Error(`No route found for ${pathname}`);
+      }
+      const pattern = getPattern(route);
+      const result = pattern.exec({ pathname });
+      const params = result?.pathname.groups ?? {};
+      tailGroup = getTailGroup(params);
+      const counter =
+        globalThis.history?.state?.counter ?? this.routerService.getCounter();
```
dunqan: one more argument to put in here

libs/platform/router/src/services/default-router.service.ts
```diff
@@ -56,7 +57,7 @@ export class DefaultRouterService implements RouterService {
   }
 
   navigate(route: string): void {
-    globalThis.history.pushState({}, '', route);
+    globalThis.history.pushState({ counter: this.counter + 1 }, '', route);
```
dunqan: this logic should happen somewhere inside lit-router

libs/platform/router/src/services/router.service.ts
```diff
@@ -13,6 +13,8 @@ export interface RouterService {
   acceptParams(params: RouteParams): void;
   getUrl(route?: string, extras?: NavigationExtras): Observable<string>;
   getPathId(id: string): string | undefined;
+  getCounter(): number;
```
dunqan: we don't need it as a public api of router services (related to ohter comments)

libs/platform/router/src/services/default-router.service.ts
aleespryker: `go` can be called for various reasons unrelated to `navigate`, so it's here instead of there.

dunqan: but should't we add counter to state on any navigation, not only programatic one?

aleespryker: We don't need to worry about this logic anymore when switching to timestamps.

libs/domain/picking/src/components/picking-header/picking-header.component.ts
```diff
@@ -41,7 +57,7 @@ export class PickingHeaderComponent extends PickingListMixin(LitElement) {
           aria-label=${i18n('oryx.picking.back-to-pick-lists')}
           class="back"
           href="#"
-          @click=${this.back}
+          @click=${this.routerService.back}
```
dunqan: let's not bind component event handlers directly to services

dunqan (dismissed):

libs/platform/router/src/services/default-router.service.ts
dunqan: I meant the whole adding state to route logic (not only counter increment).

dunqan (approved):

PR #419, opened 2023-06-06
fix: provide default value in cart add component

aleespryker:
Closes [HRZ-2944](https://spryker.atlassian.net/browse/HRZ-2944)

[HRZ-2944]: https://spryker.atlassian.net/browse/HRZ-2944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/cart/add/src/add.component.ts
+++ b/libs/domain/cart/add/src/add.component.ts
@@ -38,6 +38,7 @@ export class CartAddComponent extends ProductMixin(
     return html`<oryx-cart-quantity-input
       .min=${this.min()}
       .max=${this.max()}
+      value="1"
       @update=${this.onUpdate}
       @submit=${this.onSubmit}
     ></oryx-cart-quantity-input>`;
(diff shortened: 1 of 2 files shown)


libs/domain/cart/add/src/add.component.ts
```diff
@@ -38,6 +38,7 @@ export class CartAddComponent extends ProductMixin(
     return html`<oryx-cart-quantity-input
       .min=${this.min()}
       .max=${this.max()}
+      value="1"
```
Sinucid: I think that need to fix `oryx-cart-quantity-input` component, seems to be there is a bug with disabling of min button when value is not provided

Sinucid (changes requested):

Sinucid (approved):

PR #422, opened 2023-06-07
feat: add content and logic on user profile

nsilva-spryker:
Closes [HRZ-2897](https://spryker.atlassian.net/browse/HRZ-2897)

[HRZ-2897]: https://spryker.atlassian.net/browse/HRZ-2897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/components/user-profile/user-profile.component.ts
```diff
@@ -1,7 +1,124 @@
...
+  protected loading: boolean | null = null;
+
+  @asyncState()
+  protected currentRoute = valueType(this.routerService.currentRoute());
+
+  @asyncState()
+  protected route = valueType(this.routerService.route());
+
+  @asyncState()
+  protected pendingSyncs = valueType(this.syncSchedulerService.hasPending());
```
gund: Aren't we supposed to use signals instead of `@asyncState`?

libs/domain/picking/src/components/user-profile/user-profile.component.ts
```diff
@@ -1,7 +1,124 @@
...
+                </button>
+              </oryx-button>
+            `
+        )}
+      </div>
+    `;
+  }
+
+  protected onReceiveData(): void {
+    const app = resolve(AppRef);
```
gund: Please move this to the class property as usual DI.

libs/domain/picking/src/components/user-profile/user-profile.component.ts
```diff
@@ -1,7 +1,124 @@
...
+    const injector = app?.requirePlugin(InjectionPlugin).getInjector();
+    const injectorDataPlugin = app?.requirePlugin(OfflineDataPlugin);
+
+    injectorDataPlugin
+      .clearDb(injector)
+      .pipe(
+        tap(() => (this.loading = true)),
+        delay(500),
+        switchMap(() => injectorDataPlugin.populateDb(injector))
```
gund: I suggest to combine `clearDb` with `populateDb` into a single method to simplify component code.

libs/domain/picking/src/components/user-profile/user-profile.component.ts
```diff
@@ -1,7 +1,124 @@
...
+        );
+
+        this.loading = false;
+        this.routerService.navigate('/');
+      });
+  }
+
+  protected onLogOut(): void {
+    this.authService.logout().subscribe(() => {
+      this.routerService.navigate('/login');
```
gund: Logout method already does the redirect so you do not need it.

gund (changes requested):

libs/domain/picking/src/components/user-profile/user-profile.component.ts
```diff
@@ -1,39 +1,46 @@
...
+  protected loading: boolean | null = null;
 
-  @asyncState()
-  protected pendingSyncs = valueType(this.syncSchedulerService.hasPending());
+  protected route = signal(this.routerService.route());
+  protected pendingSyncs = signal(
+    this.syncSchedulerService
+      .hasPending()
+      .pipe(tap((pending) => (this.hasPendingSyncs = pending)))
+      .subscribe()
```
gund: Why do you need to subscribe to observable manually? Isn't `signal` going to subscribe to observable internally?

libs/domain/picking/offline/services/adapter/picking-list-online-default.adapter.ts
```diff
@@ -1,16 +1,15 @@
...
 import { isDefined } from '@spryker-oryx/utilities';
 // Add full import because of issue with naming exports from cjs.
 import * as jsonapi from 'jsonapi-serializer';
 import { firstValueFrom, map, Observable } from 'rxjs';
+import { PickingList, PickingListQualifier } from '../../../src/models';
+import {
+  GetPickingListResponse,
+  PickingHttpService,
+  PickingListDefaultAdapter,
+} from '../../../src/services';
```
gund: This is not correct as `offline` is a secondary entry point and it cannot import from the main entry point as relative imports.
Please fix other imports in `offline` modules that self-importing from itself to a relative imports, also change such imports in all `.spec` files as they should always import local files from relative imports.
Basically there should never be imports from `@spryker-oryx/picking/offline` in the offline modules and it's spec files.

gund (changes requested):

libs/domain/picking/src/components/user-profile/user-profile.styles.ts
```diff
@@ -0,0 +1,37 @@
...
+  :host {
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+  }
+
+  .info-block {
+    flex-grow: 1;
+
+    .info-label {
```
tolerants: Why we have this nesting in css? And how it's working? :)

libs/domain/picking/offline/services/adapter/picking-list-online-default.adapter.ts
nsilva-spryker: Updated files on offline modules with relative paths

libs/domain/picking/src/components/user-profile/user-profile.styles.ts
nsilva-spryker: On this case we are using only native elements:
```
 <div class="info-block">
        <dl>
          <dt class="info-label">${i18n('user.profile.employee-id')}</dt>
          <dd class="info-value">admin@spryker.com</dd>
        </dl>
      </div>
```

libs/domain/picking/src/components/user-profile/user-profile.component.ts
nsilva-spryker: @dunqan can you help on this one? it's the case we talk last week, that it's forcing to subscribe on the component.
This is something to avoid completely, or it could be necessary in some cases?

libs/domain/picking/src/components/user-profile/user-profile.component.ts
```diff
@@ -32,16 +31,17 @@ export class UserProfileComponent extends LitElement {
 
   protected route = signal(this.routerService.route());
   protected pendingSyncs = signal(
-    this.syncSchedulerService
+    resolve(SyncSchedulerService)
       .hasPending()
       .pipe(tap((pending) => (this.hasPendingSyncs = pending)))
```
gund: Btw why do you need a side effect that sets `this.hasPendingSyncs` property if you can just use `this.pendingSyncs` signal?

libs/domain/picking/offline/services/adapter/picking-list-online-default.adapter.ts
```diff
@@ -1,16 +1,15 @@
...
-  GetPickingListResponse,
-  PickingHttpService,
-  PickingList,
-  PickingListDefaultAdapter,
-  PickingListQualifier,
-} from '@spryker-oryx/picking';
 import { isDefined } from '@spryker-oryx/utilities';
 // Add full import because of issue with naming exports from cjs.
+import { PickingList, PickingListQualifier } from '@spryker-oryx/picking';
 import * as jsonapi from 'jsonapi-serializer';
```
gund: I think this comment is related to the `jsonapi-serializer` import.

libs/domain/picking/offline/services/adapter/picking-list-online-default.adapter.ts
```diff
@@ -1,16 +1,15 @@
...
 import { firstValueFrom, map, Observable } from 'rxjs';
+import {
+  GetPickingListResponse,
+  PickingHttpService,
+  PickingListDefaultAdapter,
+} from '../../../src/services';
```
gund: This import is still not fixed to be a from a package instead if a relative.

gund (changes requested):

libs/domain/picking/offline/services/adapter/picking-list-online-default.adapter.ts
nsilva-spryker: Still got this error, when using module '@spryker-oryx/picking'  to import `PickingListDefaultAdapter`:
```
TypeError: Class extends value undefined is not a constructor or null
 ❯ offline/services/adapter/picking-list-online-default.adapter.ts:24:11
     22| 
     23| export class PickingListOnlineDefaultAdapter
     24|   extends PickingListDefaultAdapter
       |           ^
     25|   implements PickingListOnlineAdapter
     26| {
```

gund: Ok let's keep it as is for now.

gund (dismissed):

libs/domain/picking/src/components/user-profile/user-profile.styles.ts
tolerants: Ok, let me take a look from a different point of view:
- Why do we need a nesting here? I mean exact `.info-block .info-label` selector.
- Is it working in FF?

nsilva-spryker: We need the nesting because `.info-block` could have in the future an iteraction of more than one `.info-label`.
Currently is working on fulfillment app.

tolerants (approved):

PR #425, opened 2023-06-08
refactor: color facet align with component conventions

tobi-or-not-tobi:
closes [HRZ-2979](https://spryker.atlassian.net/browse/HRZ-2979)

[HRZ-2979]: https://spryker.atlassian.net/browse/HRZ-2979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/search/facet-color/facet-color.component.ts
```diff
@@ -2,24 +2,21 @@ import { resolve } from '@spryker-oryx/di';
...
-import { SearchFacetComponent } from '../../facet/src/facet.component';
+import { SearchFacetComponent } from '../facet/src/facet.component';
 import { styles } from './facet-color.styles';
 
 export class SearchColorFacetComponent extends SearchFacetComponent {
-  static styles = styles;
+  static styles = [SingleFacetControlStyles, styles];
 
-  protected colorsMapping = resolve(FacetColorsMapping).reduce(
+  protected facetColorsMapping = resolve(FacetColorsMapping);
```
Sinucid, line 13: Do we need such super specific service? I would make it generic, something like `ColorService` or even just moved it as a static field to this class

tobi-or-not-tobi: The idea is that you can configure custom color facet values in a web color system. By default, we fallback to the named color (e.g. `red`), but we can set up specific colors when needed. It is very common to have custom color values, and this would only require a custom provided color list and no customisation of the component. 

I'll fix the export of the mapping, as this wasn't done.

Sinucid (approved):

PR #436, opened 2023-06-12
chore: clean up product mixin

tobi-or-not-tobi:
- fixed issue with default value for product
- drop the usage of value state in the product package


dunqan (dismissed):

libs/domain/cart/add/src/add.component.ts
```diff
@@ -21,7 +29,9 @@ export class CartAddComponent extends ProductMixin(
   static styles = styles;
 
   protected cartService = resolve(CartService);
+
   @state() protected isInvalid = false;
+  protected isInvalid2 = signal(false);
```
dunqan: 👀

libs/domain/cart/add/src/add.component.ts
```diff
@@ -58,22 +68,29 @@ export class CartAddComponent extends ProductMixin(
...
-    if (this.$product()) {
-      this.shadowRoot
-        ?.querySelector<QuantityInputComponent>('oryx-cart-quantity-input')
-        ?.reset?.();
-    }
-    return value;
+  @elementEffect()
+  protected reset = effect(() => {
+    if (this.$product()) this.input?.reset?.();
+  });
```
dunqan:
```suggestion
  protected reset = () => {
    if (this.$product()) this.input?.reset?.();
  };
```

libs/domain/product/src/mixins/product.mixin.ts
```diff
@@ -34,15 +30,23 @@ export const ProductMixin = <
...
+    protected productService = resolve(ProductService, null);
+    protected contextController = new ContextController(this);
+
+    protected $context = signal(
+      this.contextController.get<string>(ProductContext.SKU)
+    );
 
-    @asyncState()
-    protected product = valueType(this.productController.getProduct());
+    protected setProductContext = effect(() => {
```
dunqan: `@elementEffect`?

libs/domain/product/src/mixins/product.mixin.ts
```diff
@@ -34,15 +30,23 @@ export const ProductMixin = <
   class ProductMixinClass extends superClass {
     @signalProperty({ reflect: true }) sku?: string;
 
-    protected productController = new ProductController(this);
+    protected productService = resolve(ProductService, null);
+    protected contextController = new ContextController(this);
+
+    protected $context = signal(
```
dunqan: `$context` could be confusing here, as it's too generic term... what about `$sku`? (I'm aware we already have sku property)...

libs/domain/product/src/mixins/product.mixin.ts
```diff
@@ -34,15 +30,23 @@ export const ProductMixin = <
...
+      this.contextController.get<string>(ProductContext.SKU)
+    );
 
-    @asyncState()
-    protected product = valueType(this.productController.getProduct());
+    protected setProductContext = effect(() => {
+      if (this.sku && !this.$context()) {
```
dunqan: I'm wondering if we should check for `!this.$context()` as my reasoning is, that if we pass `sku` as a property, we should override context (from parent element)

dunqan (changes requested):

libs/domain/cart/add/src/add.component.ts
tobi-or-not-tobi: We need to reset the input whenever the SKU is changing

libs/domain/product/src/mixins/product.mixin.ts
```diff
@@ -34,15 +30,25 @@ export const ProductMixin = <
...
+      const sku = this.$productContext() || this.sku;
+      return sku ? this.productService?.get({ sku }) : of(null);
     });
+
+    @elementEffect()
+    protected setProductContext = () => {
+      if (this.sku && this.$productContext() !== this.sku) {
+        this.contextController.provide(ProductContext.SKU, this.sku);
+      }
+    };
```
supproduction: is it correct that we should provide context sku inside mixin?
In this case we are going to pollute dom

As far I know we need it only for card component, don't we?

supproduction (dismissed):

libs/template/labs/src/bazaarvoice/rating/rating.component.ts
```diff
@@ -12,22 +11,19 @@ export class BazaarvoiceRatingComponent extends ProductMixin(
...
-      if (product) {
-        loadBvScript().then((bv) => {
-          const summaryRefId = `${this.uid}-average-rating`;
-          this.createContainer(summaryRefId);
-
-          bv.ui('rr', 'show_reviews', {
-            productId: product?.sku,
-            summaryContainerDiv: summaryRefId,
-          });
+  protected loadScript = effect(() => {
```
dunqan:
```suggestion
  @elementEffect()
  protected loadScript = () => {
```

libs/template/labs/src/bazaarvoice/review-list/review-list.component.ts
```diff
@@ -13,23 +12,19 @@ export class BazaarvoiceReviewListComponent extends ProductMixin(
...
-        loadBvScript().then((bv) => {
-          this.hideContainer();
-          const reviewsRefId = `${this.uid}-reviews`;
-          this.createContainer(reviewsRefId);
-
-          bv.ui('rr', 'show_reviews', {
-            productId: product?.sku,
-            contentContainerDiv: reviewsRefId,
-          });
+  protected loadScript = effect(() => {
```
dunqan:
```suggestion
  @elementEffect()
  protected loadScript = () => {
```

dunqan (approved):

PR #438, opened 2023-06-12
fix: set correct paths

supproduction:
Update relative paths for picking package


libs/domain/picking/offline/services/adapter/bapi-push-notification-default.adapter.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { createInjector, destroyInjector } from '@spryker-oryx/di';
-import { PickingHttpService } from '@spryker-oryx/picking';
+import { PickingHttpService } from '../../../src/services';
```
gund, line 2: `offline` is a secondary entry point and it cannot import stuff from main entry point using relative paths, it should always import stuff from the package name.
What should be fixed is internal imports in the `offline` modules which import other `offline` modules - those should be relative. This also should be fixed in the `.spec` files of the `offline` modules.

gund (changes requested):

tolerants:
It's a `chore`, not `fix`.

nsilva-spryker:
Changes apply on `offline` module on: https://github.com/spryker/oryx/pull/422

PR #439, opened 2023-06-12
chore: content text component

tobi-or-not-tobi:


tobi-or-not-tobi:
closing as it is replaced by #504

PR #442, opened 2023-06-12
fix(fa): stub window in service worker

gund:
- Stub window in FA service worker using recursive proxy
- Remove lit-stub from FA build

Closes [HRZ-3028](https://spryker.atlassian.net/browse/HRZ-3028)

[HRZ-3028]: https://spryker.atlassian.net/browse/HRZ-3028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/fulfillment/sw/main.ts
```diff
@@ -1,5 +1,7 @@
 /// <reference lib="WebWorker" />
 
+import './window-stub';
```
dunqan, line 4: Can you add a line of explanation, why are we stubbing window in service worker?

tobi-or-not-tobi (dismissed):

dunqan (approved):

PR #444, opened 2023-06-12
feat: enableList option for checkout address

tobi-or-not-tobi:
Addresses on the checkout (shipping or billing) can now be shown as a list or just the selected details. 

closes [HRZ-2586](https://spryker.atlassian.net/browse/HRZ-2586)

[HRZ-2586]: https://spryker.atlassian.net/browse/HRZ-2586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/checkout/address/address.component.spec.ts
```diff
@@ -154,34 +155,67 @@ describe('CheckoutAddressComponent', () => {
...
+          expect(element).toContainElement('oryx-user-address-list');
+        });
+
+        it('should not render the oryx-user-address', () => {
+          expect(element).not.toContainElement('oryx-user-address');
+        });
+
+        describe('and an change event is dispatched on the list', () => {
+          beforeEach(() => {
+            element.dispatchEvent = vi.fn();
```
Sinucid: Why do you need to mock the method?

libs/domain/checkout/address/address.component.spec.ts
```diff
@@ -154,34 +155,67 @@ describe('CheckoutAddressComponent', () => {
...
+            const list = element.renderRoot.querySelector(
+              'oryx-user-address-list'
+            ) as HTMLElement;
+            list.dispatchEvent(
+              new CustomEvent('change', {
```
Sinucid, line 209: And according to our convention it is public event and it should have `oryx` prefix, ie `oryx.change`

libs/domain/checkout/address/address.component.ts
```diff
@@ -6,30 +6,36 @@ import {
...
-        .addressId=${this.$addressId()}
-        .options=${{
-          selectable: true,
-          addressDefaults: AddressDefaults.All,
-        } as UserAddressListOptions}
-        @change=${this.onChange}
-      ></oryx-user-address-list>`;
+    if (this.$addresses()?.length) {
+      const { enableList } = this.$options();
+      if (enableList) {
```
Sinucid:
```suggestion
      if (this.$options().enableList) {
```

libs/domain/checkout/address/address.model.ts
```diff
@@ -0,0 +1,3 @@
+export interface CheckoutAddressOptions {
+  enableList?: boolean;
```
Sinucid, line 6: doc?

libs/domain/checkout/address/stories/demo.stories.ts
```diff
@@ -0,0 +1,44 @@
...
+  argTypes: {
+    addressId: {
+      options: mockNormalizedAddresses.map((a) => a.id),
+      control: { type: 'select' },
+      table: { category: 'Demo' },
+    },
+  },
+} as Meta;
+
+interface Props {
```
Sinucid:
```suggestion
interface Props extends CheckoutAddressOptions {
```

libs/domain/checkout/address/stories/static.stories.ts
```diff
@@ -0,0 +1,45 @@
...
+export default {
+  title: `${storybookPrefix}/Address`,
+} as Meta;
+
+const service = resolve<MockAddressService>(AddressService);
+service.changeMockAddressType(MockAddressType.ThreeWithDefaults);
+
+const Template: Story = (): TemplateResult => {
+  return html`
+    <h3>(multiline)</h3>
```
Sinucid: Do we need brackets it each title? I don't remember us using something like this in other stories

libs/domain/checkout/billing-address/billing-address.component.ts
```diff
@@ -45,7 +45,7 @@ export class CheckoutBillingAddressComponent
     if (
       deliveryAddress &&
       this.$isSameAsShippingAddress() &&
-      !this.$addresses()?.length
+      this.$addresses()?.length === 0
```
Sinucid, line 48: Is it important to check exact equality to 0 instead of just falsy check?

Sinucid (changes requested):

libs/domain/checkout/billing-address/billing-address.component.ts
tobi-or-not-tobi: yup, otherwise it returns false if the addresses are not yet loaded

libs/domain/checkout/address/stories/static.stories.ts
tobi-or-not-tobi: We do not, will drop it as this is static anyway

libs/domain/checkout/address/stories/demo.stories.ts
tobi-or-not-tobi: I actually use the type at the argument

tobi-or-not-tobi: `Story<Props & CheckoutAddressOptions>`

Sinucid: Yes, I saw that. But it is more clear and readable then extending on place

libs/domain/checkout/address/address.component.spec.ts
tobi-or-not-tobi: yeah, we started to drop that convention. We could discuss it, but it's no so much part of this PR

libs/domain/checkout/address/address.component.spec.ts
tobi-or-not-tobi: copy-paste I guess

libs/domain/checkout/address/address.component.spec.ts
Sinucid: Won't they then overlap with standard HTML events?

Sinucid (dismissed):

PR #448, opened 2023-06-13
chore: scroll to top on page change

aleespryker:
Closes [HRZ-2573](https://spryker.atlassian.net/browse/HRZ-2573)

[HRZ-2573]: https://spryker.atlassian.net/browse/HRZ-2573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/router/src/services/default-router.service.ts
+++ b/libs/platform/router/src/services/default-router.service.ts
@@ -48,7 +48,7 @@ export class DefaultRouterService implements RouterService {
     this.router$.next(url[0]);
     this.urlSearchParams$.next(queryParams);
     this.routerEvents$.next({ route, type: RouterEventType.NavigationEnd });
-    window.scrollTo(0, 0);
+    globalThis.scrollTo?.(0, 0);
   }
 
   navigate(route: string): void {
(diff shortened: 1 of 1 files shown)


supproduction (approved):

libs/platform/router/src/services/default-router.service.ts
```diff
@@ -48,6 +48,7 @@ export class DefaultRouterService implements RouterService {
     this.router$.next(url[0]);
     this.urlSearchParams$.next(queryParams);
     this.routerEvents$.next({ route, type: RouterEventType.NavigationEnd });
+    globalThis.scrollTo?.(0, 0);
```
tobi-or-not-tobi, line 51: can we make this work with https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior? It means that JS code only scrolls to the position, and css wil dicate the behviaor

tobi-or-not-tobi (changes requested):

PR #449, opened 2023-06-13
feat: confirm registered user data on checkout page

tobi-or-not-tobi:
When a registered user lands at the checkout page, we confirm the name and email address. 

Additionally, we've added some default values that can be resolved from the user and the users address book:
- copy first name over to new address
- copy last name over to new address
- copy salutation over to new address
- set default shipping address to true if there's no default shipping address
- set default billing address to true if there's no default billing address

closes [HRZ-3057](https://spryker.atlassian.net/browse/HRZ-3057)

[HRZ-3057]: https://spryker.atlassian.net/browse/HRZ-3057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/checkout/customer/customer.component.spec.ts
+++ b/libs/domain/checkout/customer/customer.component.spec.ts
@@ -108,15 +108,21 @@ describe('CheckoutAuthComponent', () => {
   describe('when the user is authenticated', () => {
     beforeEach(async () => {
       authService.isAuthenticated.mockReturnValue(of(true));
-      userService.getUser.mockReturnValue(of({ email: 'foo@bar.com' } as User));
+      userService.getUser.mockReturnValue(
+        of({ firstName: 'Foo', lastName: 'Bar', email: 'foo@bar.com' } as User)
+      );
       element = await fixture(
         html`<oryx-checkout-customer></oryx-checkout-customer>`
       );
     });
 
-    it('should render the heading', () => {
-      const h1 = element.renderRoot.querySelector('h1');
-      expect(h1?.textContent).toBe('Checkout');
+    it('should render the user name', () => {
+      const h1 = element.renderRoot.querySelector('h3');
+      expect(h1?.textContent).toBe('Checking out as');
+
+      const customerData = element.renderRoot.querySelector('h3 + p');
+      expect(customerData?.textContent).contain('Foo Bar');
+      expect(customerData?.textContent).contain('foo@bar.com');
     });
 
     it('should not render the guest component', () => {
--- a/libs/domain/checkout/customer/customer.component.spec.ts
+++ b/libs/domain/checkout/customer/customer.component.spec.ts
@@ -120,7 +120,7 @@ describe('CheckoutAuthComponent', () => {
     });
 
     it('should render the user name', () => {
-      const customerData = element.renderRoot.querySelector('p');
+      const customerData = element.renderRoot.querySelector('p')?.textContent;
       expect(customerData).toContain(`${user.firstName} ${user.lastName}`);
       expect(customerData).toContain(user.email);
     });
--- a/libs/domain/checkout/customer/customer.component.spec.ts
+++ b/libs/domain/checkout/customer/customer.component.spec.ts
@@ -106,23 +106,23 @@ describe('CheckoutAuthComponent', () => {
   });
 
   describe('when the user is authenticated', () => {
+    const user = {
+      firstName: 'Foo',
+      lastName: 'Bar',
+      email: 'foo@bar.com',
+    } as User;
     beforeEach(async () => {
       authService.isAuthenticated.mockReturnValue(of(true));
-      userService.getUser.mockReturnValue(
-        of({ firstName: 'Foo', lastName: 'Bar', email: 'foo@bar.com' } as User)
-      );
+      userService.getUser.mockReturnValue(of(user));
       element = await fixture(
         html`<oryx-checkout-customer></oryx-checkout-customer>`
       );
     });
 
     it('should render the user name', () => {
-      const h1 = element.renderRoot.querySelector('h3');
-      expect(h1?.textContent).toBe('Checking out as');
-
-      const customerData = element.renderRoot.querySelector('h3 + p');
-      expect(customerData?.textContent).contain('Foo Bar');
-      expect(customerData?.textContent).contain('foo@bar.com');
+      const customerData = element.renderRoot.querySelector('p');
+      expect(customerData).toContain(`${user.firstName} ${user.lastName}`);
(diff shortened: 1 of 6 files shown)


libs/domain/checkout/customer/customer.component.spec.ts
```diff
@@ -108,15 +108,21 @@ describe('CheckoutAuthComponent', () => {
   describe('when the user is authenticated', () => {
     beforeEach(async () => {
       authService.isAuthenticated.mockReturnValue(of(true));
-      userService.getUser.mockReturnValue(of({ email: 'foo@bar.com' } as User));
+      userService.getUser.mockReturnValue(
```
Sinucid: I do not see any reason to check heading's content, it is not conditional. I would suggest checking it this way:

```
const user = { firstName: 'Foo', lastName: 'Bar', email: 'foo@bar.com' } as User;
 beforeEach(async () => {
      authService.isAuthenticated.mockReturnValue(of(true));
      userService.getUser.mockReturnValue(
        of(user)
      );
      element = await fixture(
        html`<oryx-checkout-customer></oryx-checkout-customer>`
      );
    });

    it('should render user`s data', () => {
      const customerData = element.renderRoot.querySelector('p')?.textContent;
      expect(customerData).toContain(user.firstName);
      expect(customerData).toContain(user.lastName);
      expect(customerData).toContain(user.email);
    });
```

libs/domain/user/address-form/address-form.component.spec.ts
```diff
@@ -276,26 +296,94 @@ describe('UserAddressFormComponent', () => {
...
+        );
+        element = await fixture(html` <oryx-user-address-form
+          enableDefaultShipping
+        ></oryx-user-address-form>`);
+      });
+
+      it('should set isDefaultShipping to true in the default values', () => {
+        expect(renderer.buildForm).toHaveBeenCalledWith(
+          expect.anything(),
+          expect.not.objectContaining({ isDefaultShipping: true })
```
Sinucid, line 316: Might be a mistake. Should set isDefaultShipping to true, but expects that it is not

libs/domain/user/address-form/address-form.component.spec.ts
```diff
@@ -276,26 +296,94 @@ describe('UserAddressFormComponent', () => {
...
+        );
+        element = await fixture(html` <oryx-user-address-form
+          enableDefaultBilling
+        ></oryx-user-address-form>`);
+      });
+
+      it('should set isDefaultBilling to true in the default values', () => {
+        expect(renderer.buildForm).toHaveBeenCalledWith(
+          expect.anything(),
+          expect.not.objectContaining({ isDefaultBilling: true })
```
Sinucid, line 367: the same, expects that not

libs/domain/user/address-form/address-form.component.spec.ts
```diff
@@ -313,4 +401,23 @@ describe('UserAddressFormComponent', () => {
       );
     });
   });
+
+  describe('when a registered user creates a new address', () => {
+    beforeEach(async () => {
```
Sinucid, line 411:
```
describe('when a registered user creates a new address', () => {
const user = { firstName: 'First', lastName: 'Last', salutation: 'Mr' } as User;
    beforeEach(async () => {
      userService.getUser.mockReturnValue(
        of(user)
      );
      element = await fixture(html`
        <oryx-user-address-form></oryx-user-address-form>
      `);
    });

    it('should feed in the user`s data', () => {
      expect(renderer.buildForm).toHaveBeenCalledWith(expect.anything(), user);
    });
  });
```

libs/domain/user/address-form/address-form.component.ts
```diff
@@ -42,45 +44,69 @@ export class UserAddressFormComponent
...
+        this.getFormValues()
+      )}
     </form>`;
   }
 
+  protected getFormValues(): FormValues {
+    const values = this.values ?? {};
+    const { firstName, lastName, salutation } = this.$user() ?? {};
+
+    const a = this.$addresses();
```
Sinucid:
```suggestion
    const address = this.$addresses();
```

libs/domain/user/address-form/address-form.component.ts
```diff
@@ -42,45 +44,69 @@ export class UserAddressFormComponent
...
+      values.isDefaultBilling = true;
+    }
+    if (
+      this.enableDefaultShipping &&
+      !a?.find((address) => address.isDefaultShipping)
+    ) {
+      values.isDefaultShipping = true;
+    }
+
+    if (firstName) values.firstName ??= firstName;
```
Sinucid: Seems to be it is pretty similar, but without if
```suggestion
   values.firstName ??= firstName;
```

libs/domain/user/address-form/address-form.component.ts
```diff
@@ -42,45 +44,69 @@ export class UserAddressFormComponent
...
+    if (lastName) values.lastName ??= lastName;
+    if (salutation) values.salutation ??= salutation;
+    return values;
+  }
+
   protected getFormFields(): FormFieldDefinition[] {
-    const form = this.formModel();
+    const form = this.$formModel();
```
Sinucid, line 109: Oh, I still not get should computed have $ prefix or not?

Sinucid (changes requested):

libs/domain/user/address-form/address-form.component.ts
tobi-or-not-tobi: not yet sure tbh, we'll discuss today in arch catchup

libs/domain/user/address-form/address-form.component.ts
tobi-or-not-tobi: 100%

libs/domain/checkout/customer/customer.component.spec.ts
```diff
@@ -106,17 +106,23 @@ describe('CheckoutAuthComponent', () => {
...
+      userService.getUser.mockReturnValue(of(user));
       element = await fixture(
         html`<oryx-checkout-customer></oryx-checkout-customer>`
       );
     });
 
-    it('should render the heading', () => {
-      const h1 = element.renderRoot.querySelector('h1');
-      expect(h1?.textContent).toBe('Checkout');
+    it('should render the user name', () => {
```
Sinucid: desc is not relevant

Sinucid (changes requested):

Sinucid (approved):

PR #452, opened 2023-06-14
refactor: command and queries for address management

dunqan:
Closes [HRZ-3048](https://spryker.atlassian.net/browse/HRZ-3048)

[HRZ-3048]: https://spryker.atlassian.net/browse/HRZ-3048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/checkout/billing-address/billing-address.component.ts
```diff
@@ -52,12 +59,13 @@ export class CheckoutBillingAddressComponent
...
       });
     }
-  });
+  };
 
   @query('oryx-checkout-address')
   protected checkoutAddress?: CheckoutAddressComponent;
 
-  protected override render(): TemplateResult {
+  protected override render(): TemplateResult | undefined {
```
tobi-or-not-tobi:
```suggestion
  protected override render(): TemplateResult | void {
```

I've been using void everywhere, but wasn't aware undefined could be used as well

libs/domain/user/src/services/address.service.ts
```diff
@@ -4,7 +4,7 @@ import { Address } from '../models';
 export interface AddressService {
   getCurrentAddress(): Observable<Address | null>;
   getAddress(addressId: string): Observable<Address | null>;
-  getAddresses(): Observable<Address[] | null>;
+  getAddresses(): Observable<Address[] | undefined>;
```
tobi-or-not-tobi: What about following CRUD?
What about avoiding the domain in the methods?

```ts
create(data: Address): Observable<Address | null>;
read(data: Address): Observable<Address | null>;
update(data: Address): Observable<Address | null>;
delete(data: Address): Observable<Address | null>;

//
readAll(): Observable<Address[] | null | undefined>; 
(or a single read with conditional args)
readCurrent(): Observable<Address | null>;
or
current(): Observable<Address | null>;
```

tobi-or-not-tobi (dismissed):

libs/domain/checkout/billing-address/billing-address.component.ts
dunqan: I'd use `void` when you don't return or don't use return result from the function (or result is not important). In this case it's debatable, as the name of the method suggests doing something (`render`) while technically it's just returning TemplateResult for the rendering logic. That's why I went with TemplateResult | undefined.

libs/domain/user/src/services/address.service.ts
dunqan: 👍

libs/domain/checkout/billing-address/billing-address.component.ts
dunqan: ...but I actually got convinced, void looks bit nicier even in this context. And we don't write `return undefined` explicitely anyway.

tobi-or-not-tobi (approved):

PR #453, opened 2023-06-14
chore: facet component refactoring

Sinucid:
closes to [HRZ-2980](https://spryker.atlassian.net/browse/HRZ-2980)

[HRZ-2980]: https://spryker.atlassian.net/browse/HRZ-2980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/form/select/src/stories/select-demo.stories.ts
```diff
@@ -42,7 +42,7 @@ const Template: Story<Props> = (props: Props): TemplateResult => {
       (ev.detail.selected as HTMLOptionElement).value
     );
   };
-  const logSearch = (ev: CustomEvent<SearchEvent>): void => {
+  const logSearch = (ev: CustomEvent<SearchPayload>): void => {
```
tobi-or-not-tobi: not sure about payload as it associates with rest APIs. What about `SearchEventDetail`?

libs/domain/search/facet/controllers/facet.controller.ts
```diff
@@ -1,49 +1,71 @@
...
+    switchMap((name) =>
+      name
+        ? (this.facetListService.getFacet({
+            name,
+          }) as Observable<SingleMultiFacet>)
+        : of(null)
+    )
+  );
+
+  protected $facet = signal<SingleMultiFacet | null>(this.facet$);
```
tobi-or-not-tobi: can we move those into a single computed?

libs/domain/search/facet/controllers/facet.controller.ts
```diff
@@ -1,49 +1,71 @@
...
 export class FacetController implements ReactiveController {
-  protected observe: ObserveController<LitElement & FacetComponentAttributes>;
+  protected observe = new ObserveController(this.host);
   protected facetListService = resolve(FacetListService);
-  protected facet$: Observable<Facet | null>;
-  protected showAll$ = new BehaviorSubject(false);
-  protected searchedValue$ = new BehaviorSubject('');
 
-  constructor(protected host: LitElement & FacetComponentAttributes) {
+  protected facet$ = this.observe.get('name').pipe(
```
tobi-or-not-tobi: we should not need to observe now that name is a signalProperty

libs/domain/search/facet/facet.component.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  protected facet = computed(() => this.controller.getFacet());
+
+  protected onChange(e: InputEvent): void {
+    const { value, checked: selected } = e.target as HTMLInputElement;
+    this.controller.dispatchSelectEvent({
+      value,
+      selected,
+    });
+  }
```
tobi-or-not-tobi, line 35: I think we should move the method back to component

tobi-or-not-tobi (changes requested):

libs/base/ui/form/select/src/stories/select-demo.stories.ts
Sinucid: I believe that `detail` is close enough to requests `data`, but agree, this can confuse. What about `[eventNameOrSomething]EventPayload` -> SearchEventPayload?

libs/domain/search/facet/controllers/facet.controller.ts
Sinucid: But this is controller, in theory it can be applied to any other facet-like component. Lets continue observing to keep it a bit more generic and flexible

libs/domain/search/facet/facet.component.ts
Sinucid: It is no the best idea due to controller also uses this dispatcher inside onClear event handler. If we move to the component then we should duplicate the dispatcher for the controller or make it public and create unnucessary circular dependency between controller and componen

tobi-or-not-tobi: ah i missed that

libs/base/ui/form/select/src/stories/select-demo.stories.ts
tobi-or-not-tobi: That could work although I do not see the issue with `SearchEventDetail`.

tobi-or-not-tobi (approved):

PR #454, opened 2023-06-14
feat: implement references component

tarashaliuk:
closes: [HRZ-2838](https://spryker.atlassian.net/browse/HRZ-2838)

[HRZ-2838]: https://spryker.atlassian.net/browse/HRZ-2838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/product/references/src/references.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  ProductMixin,
+} from '@spryker-oryx/product';
+import { computed, hydratable } from '@spryker-oryx/utilities';
+import { html, LitElement, TemplateResult } from 'lit';
+import { repeat } from 'lit/directives/repeat.js';
+import { unsafeHTML } from 'lit/directives/unsafe-html.js';
+import { of } from 'rxjs';
+
+@hydratable()
+export class ReferencesComponent extends ProductMixin(LayoutMixin(LitElement)) {
```
tobi-or-not-tobi:
```suggestion
export class ProductReferencesComponent extends ProductMixin(LayoutMixin(LitElement)) {
```

libs/domain/product/references/src/references.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  protected alternativeProductsListService = resolve(
+    AlternativeProductsListService
+  );
+
+  protected list = computed(() => {
+    return this.$product()?.sku
+      ? this.alternativeProductsListService.get(this.$product()!.sku!)
```
tobi-or-not-tobi: what about:

```ts
const sku = this.$product()?.sku;
return sku ? this.alternativeProductsListService.get(sku) : of(undefined);
``

File: libs/domain/product/references/src/references.component.ts [hunk 3]
```diff
@@ -0,0 +1,35 @@
...
+import { unsafeHTML } from 'lit/directives/unsafe-html.js';
+import { of } from 'rxjs';
+
+@hydratable()
+export class ReferencesComponent extends ProductMixin(LayoutMixin(LitElement)) {
+  protected alternativeProductsListService = resolve(
+    AlternativeProductsListService
+  );
+
+  protected list = computed(() => {
```
tobi-or-not-tobi:
```suggestion
  protected $list = computed(() => {
```

we just agreed on a new convention to prefix both computed and signals with a dollar.

libs/domain/product/src/services/adapter/alternative-products-list.adapter.ts
```diff
@@ -0,0 +1,15 @@
+import { Product } from '@spryker-oryx/product';
+import { Observable } from 'rxjs';
+
+export interface AlternativeProductsListAdapter {
+  get(sku: string): Observable<Product[]>;
```
tobi-or-not-tobi: Can we use a qualifier instead? we like to avoid many arguments and rather pass qualifier objects through all layers. With that, we can extend qualifiers and do not need to change all application layers to bring in new things. 

This will allow to bring in component options going forward to query alternative alternatives - say up sell vs cross sell vs spair parts, etc.

libs/domain/product/src/services/adapter/default-alternative-products-list.adapter.ts
```diff
@@ -0,0 +1,32 @@
...
+  get(sku: string): Observable<Product[]> {
+    const include = [ApiProductModel.Includes.ConcreteProductPrices];
+
+    return this.http
+      .get<ApiProductModel.Response>(
+        `${this.SCOS_BASE_URL}/${this.getQueryEndpoint(
+          sku
+        )}?include=${include?.join(',')}`
+      )
+      .pipe(this.transformer.do(AlternativeProductsListNormalizer));
```
tobi-or-not-tobi, line 30: can we not use the default product (list) normalizer?

tobi-or-not-tobi (changes requested):

libs/domain/product/src/services/adapter/default-alternative-products-list.adapter.ts
tarashaliuk: The `ProductListNormalizer` is incompletable with the alternative products list data structure. The `AlternativeProductsListNormalizer` uses the `ProductNormalizer` inside

tobi-or-not-tobi (approved):

PR #464, opened 2023-06-16
test: use different customers in tests to avoid race conditions

AndreyYevtukhov:


apps/storefront-e2e/src/types/user.type.ts
```diff
@@ -1,5 +1,5 @@
-export type TestUserData = {
-  id: string;
+export type TestCustomerData = {
```
tolerants, line 1: I think it's worth to rename the file to `customer.type.ts`.

apps/storefront-e2e/src/support/page_fragments/login.fragment.ts
```diff
@@ -9,7 +9,7 @@ export class LoginFragment {
   getLoginButton = () => this.getWrapper().contains('button', 'Login');
   getBEValidationError = () => this.getWrapper().find('oryx-notification');
 
-  login = (user: TestUserData) => {
+  login = (user: TestCustomerData) => {
```
tolerants, line 12: What is the actual difference in tests between user and customer?

tolerants (changes requested):

apps/storefront-e2e/src/support/page_fragments/login.fragment.ts
AndreyYevtukhov: there is no difference now, but in the future, we might have "admin logged in as a customer" or something, so we should not use "user" in the code

I forget to change the variable name here, but I'll do it later, cause it might require more changes (this PR should be merge as soon as possible to unblock other branches)

apps/storefront-e2e/src/types/user.type.ts
AndreyYevtukhov: yep yep, sure thing!) I'm 100% sure that there are more places with `user` in the code(

AndreyYevtukhov: I'll fix all of these in the follow-up pr

tolerants (approved):

tobi-or-not-tobi (approved):

PR #471, opened 2023-06-16
feat: e2e tests for product references component

tarashaliuk:
closes: [HRZ-2839](https://spryker.atlassian.net/browse/HRZ-2839)

[HRZ-2839]: https://spryker.atlassian.net/browse/HRZ-2839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront-e2e/src/support/page_objects/product-details.page.ts
+++ b/apps/storefront-e2e/src/support/page_objects/product-details.page.ts
@@ -1,6 +1,9 @@
 import { TestProductData } from '../../types/product.type';
+import { ProductReferencesFragment } from '../page_fragments/product-references.fragment';
 import { QuantityInputFragment } from '../page_fragments/quantity-input.fragment';
 import { AbstractSFPage } from './abstract.page';
+
+const productReferencesFragment = new ProductReferencesFragment();
 
 export class ProductDetailsPage extends AbstractSFPage {
   url = '/product/';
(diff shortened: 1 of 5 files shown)


apps/storefront-e2e/src/support/page_objects/product-details.page.ts
```diff
@@ -1,7 +1,10 @@
 import { TestProductData } from '../../types/product.type';
+import { ProductReferencesFragment } from '../page_fragments/product-references.fragment';
 import { QuantityInputFragment } from '../page_fragments/quantity-input.fragment';
 import { AbstractSFPage } from './abstract.page';
 
+const productReferencesFragment = new ProductReferencesFragment();
```
AndreyYevtukhov: Let's have `ProductReferencesFragment` as part of the PDP page, like

`getReferences = () => new ProductReferencesFragment()`

and then in tests

`pdp.getReferences().getWrapper()`

it will make test code more readable because you will always know on which page and with which page elements you do actions + you won't have duplicated methods (now we have 2 getters in Fragment and 2 methods in PDP class. If in the future we will add more elements in Fragment - we will have to "syncronize" them with PDP, which is not good)

AndreyYevtukhov (changes requested):
All good, only 1 small change is needed

AndreyYevtukhov (dismissed):

AndreyYevtukhov (approved):

PR #476, opened 2023-06-19
chore: rename references component and extract it from src folder

tarashaliuk:
closes: [HRZ-3082](https://spryker.atlassian.net/browse/HRZ-3082)

[HRZ-3082]: https://spryker.atlassian.net/browse/HRZ-3082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/product/references/src/references.def.ts
tobi-or-not-tobi, line 1: name should be updated

tobi-or-not-tobi (changes requested):

libs/domain/product/references/src/references.def.ts
tarashaliuk: fixed

tobi-or-not-tobi (approved):

PR #478, opened 2023-06-19
feat: add product relations story, renamed service and adapter

tarashaliuk:
closes: [HRZ-3072](https://spryker.atlassian.net/browse/HRZ-3072)
Also, renamed service and adapter

[HRZ-3072]: https://spryker.atlassian.net/browse/HRZ-3072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/product/src/mocks/src/mock-product.providers.ts
```diff
@@ -45,4 +47,8 @@ export const mockProductProviders: Provider[] = [
     provide: ProductMediaConfig,
     useValue: productMediaConfig,
   },
+  {
+    provide: AlternativeProductsListService,
```
tobi-or-not-tobi: I had not captured this before, what about renaming the service to `ProductRelationService`?

tobi-or-not-tobi (dismissed):

libs/domain/product/src/mocks/src/mock-product.providers.ts
tarashaliuk: I guess `ProductRelationsListService` will be a better name

tobi-or-not-tobi (approved):

PR #480, opened 2023-06-20
fix: switch fulfillment icons to material designs

aleespryker:
Closes [HRZ-3085](https://spryker.atlassian.net/browse/HRZ-3085)

[HRZ-3085]: https://spryker.atlassian.net/browse/HRZ-3085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/themes/src/fulfillment.theme.ts
```diff
@@ -16,6 +16,15 @@ export const fulfillmentTheme: Theme = {
         resource: fulfillmentIcons,
         types: Object.keys(fulfillmentIcons.mapping ?? {}),
```
gund, line 17: Do we still need this old svg icons in resourses?

aleespryker: Unfortunately, some of the icons in the design don't match the ones in figma (eg, the shopping cart icon is shaped differently, loading indicator also looks different) so we have to keep some of the svg icons. Storefront is doing this too apparently.

tobi-or-not-tobi: correct. The shopping cart was explicitly not chosen from material design, which is understandable.

AndreyYevtukhov (approved):

PR #481, opened 2023-06-20
fix: back button does not return user to Picking Lists page

tarashaliuk:
closes: [HRZ-3091](https://spryker.atlassian.net/browse/HRZ-3091)

[HRZ-3091]: https://spryker.atlassian.net/browse/HRZ-3091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/src/components/navigate-back/navigate-back.component.ts
```diff
@@ -22,8 +22,14 @@ export class NavigateBackComponent
...
 
+  @asyncState()
+  protected currentLink = valueType(this.routerService.currentRoute());
+
   protected override render(): TemplateResult {
-    const link = this.link || this.fallbackUrl;
+    const link =
+      this.link && this.link !== this.currentLink
+        ? this.link
+        : this.fallbackUrl;
```
gund: Shouldn't we always navigate to the picking lists page instead? What's the case for navigating to a previous page?

tarashaliuk: I believe this component was built with the thought of using it also in other places (in the future). That's why it doesn’t navigate always to `/`

gund: Oh true, but then it is accepting a prop for that so it can be customised easily where it's used.
So still my concern remains - can we remove all that complex router logic and just have a link prop?

tarashaliuk: I guess we can do it and delegate the generating url logic to the parent component and provide it through the prop.

gund (dismissed):

gund (approved):

PR #487, opened 2023-06-21
refactor: checkout comand & query refactor

dunqan:
Closes [HRZ-3073](https://spryker.atlassian.net/browse/HRZ-3073)

[HRZ-3073]: https://spryker.atlassian.net/browse/HRZ-3073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/checkout/src/mixins/checkout.mixin.ts
tobi-or-not-tobi, line 1: let's use new dollar prefix conventions for signals and computed. Also, we could change some signals to get ride of rxjs logic in this layer.

tobi-or-not-tobi (dismissed):

tobi-or-not-tobi (approved):

PR #488, opened 2023-06-21
feat: add stories for fulfillment user profile component

aleespryker:
Closes [HRZ-3058](https://spryker.atlassian.net/browse/HRZ-3058)

[HRZ-3058]: https://spryker.atlassian.net/browse/HRZ-3058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/offline/data-plugin.ts
```diff
@@ -20,37 +20,41 @@ export class OfflineDataPlugin extends ExecPlugin {
   protected subscription?: Subscription;
 
   constructor() {
-    super((app) => {
-      const injector = app!.requirePlugin(InjectionPlugin).getInjector();
+    super(() => undefined);
+  }
+
+  apply(app: App): void | Promise<void> {
```
gund, line 26: This plugin does not need to have logic in `apply()` method as it's a special `ExecPlugin` which accepts function in constructor and executes it automatically.
Otherwise you do not need to extend `ExecPlugin` and just implement `AppPlugin` interface.

gund (changes requested):

libs/domain/picking/offline/data-plugin.ts
aleespryker: I wanted to leave it in constructor, but it wouldn't let me call it without also calling `super`, which is what I'm trying to avoid so we don't pull in all those other services we do not need.

gund: Then you do not need to extend the `ExecPlugin` - just implement `AppPlugin` interface.

libs/domain/picking/offline/data-plugin.ts
```diff
@@ -16,16 +16,14 @@ import {
...
 import { PickingListOnlineAdapter } from './services';
 
-export class OfflineDataPlugin extends ExecPlugin {
+export class OfflineDataPlugin implements AppPlugin {
   protected subscription?: Subscription;
 
-  constructor() {
-    super(() => undefined);
+  getName(): string {
+    return 'oryx$pickingOfflineData';
```
gund: We've dropped `$` sign to use `.` instead:
```suggestion
    return 'oryx.pickingOfflineData';
```

gund (approved):

gund (changes requested):
Wrong resolution, please see above comment =)

gund (approved):

tolerants:
It looks like commit type and the message are controversial - we have a `feat` but at the same time description says it's just adding some stories.

PR #490, opened 2023-06-21
feat: add pluggable search suggestions

supproduction:
adds pluggable search suggestions

closes [HRZ-3086](https://spryker.atlassian.net/browse/HRZ-3086)

[HRZ-3086]: https://spryker.atlassian.net/browse/HRZ-3086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/.env.template
```diff
@@ -3,7 +3,7 @@ SCOS_BASE_URL=
 
 # The STORE env variable is optional and can be used to initialise the application to use a
 # specific store. When this variable is not provided, the first store from the backend would be used.
-# STORE=DE
+STORE=DE
```
tobi-or-not-tobi: why did we move this back?

apps/storefront/.env.template
```diff
@@ -17,3 +17,9 @@ ORYX_CLOUDINARY_ID=
 # and might even be broken as they're not thoroughly tested
 # and not covered with automated tests.
 ORYX_LABS=
+
+# The Contentful Access Token
+ORYX_CONTENTFUL_TOKEN=
+# All content and assets in Contentful belong to a space.
+# You will generally have at least one space for a project.
+ORYX_CONTENTFUL_SPACE=
```
tobi-or-not-tobi, line 25: Can we add a warning in the integration when the contentful integration is enabled, but those are not provided. We did the same in the coudinary integration. It's convenient. Probably only in dev mode.

libs/domain/content/article/article.component.ts
```diff
@@ -0,0 +1,38 @@
...
+export class ArticleComponent extends ContentMixin<ContentArticleOptions>(
+  LitElement
+) {
+  protected contentService = resolve(ContentService);
+  protected contextController = new ContextController(this);
+
+  protected $articleContext = signal(
+    this.contextController.get<string>(ArticleContext.Id)
+  );
+  protected $article = computed(() => {
```
tobi-or-not-tobi, line 35:
```suggestion

  protected $article = computed(() => {
```

libs/domain/content/article/article.component.ts
```diff
@@ -0,0 +1,38 @@
...
+    const article = this.$articleContext();
+
+    return article
+      ? this.contentService.get({ article, entries: this.$options().entries })
+      : of(null);
+  });
+
+  protected override render(): TemplateResult {
+    return html`${this.$article()?.article}`;
```
tobi-or-not-tobi: Let's use `<oryx-text .content=${}></oryx-text>` now that we have wysiwyg text.

libs/domain/content/article/article.model.ts
```diff
@@ -0,0 +1,3 @@
+export interface ContentArticleOptions {
+  entries?: string[];
```
tobi-or-not-tobi: I don't understand entries. Can we document it so it becomes more clearer?

libs/domain/search/box/box.component.ts
```diff
@@ -29,28 +26,24 @@ import {
...
-
 @defaultOptions({
   minChars: 2,
   completionsCount: 5,
   productsCount: 5,
   categoriesCount: 5,
-  cmsCount: 0,
+  cmsCount: 5,
+  entries: [
+    SuggestionField.Completion,
```
tobi-or-not-tobi:
```suggestion
    SuggestionField.Suggestions,
```

libs/domain/search/src/services/suggestion/renderer/suggestion-renderer.service.ts
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Maximum amount of CMS links to show
+   * @default 5
+   */
+  cmsCount?: number;
+
+  /**
+   * List of entries which should be shown
+   */
+  entries?: SuggestionEntries;
```
tobi-or-not-tobi: ok, got it. The entries are the "entities" that you can search for. Wonder why they are coupled with suggestions, as I was imagine this is not only for related to suggestions, but also to the results of that entity.

libs/platform/experience/src/routes.ts
```diff
@@ -30,6 +30,13 @@ export const defaultExperienceRoutes: RouteConfig[] = [
     ></oryx-composition>`,
     name: 'Order',
   },
+  {
+    path: '/article/:id',
```
tobi-or-not-tobi, line 34: We do want to end up with various entities connected to various cms providers, i.e.:
- `/faq/:id` - could be resolved by cms 1
- `/jobs/:id` - could be resolved by cms 2
- `/:id` - fallback content, could be resolved by cms 3

Totally fine to work on this going forward, but want to make sure we're aligned on the vision so you can make the right choices in the design.

libs/template/labs/src/contentful/client/contentful-client.service.ts
```diff
@@ -0,0 +1,39 @@
...
+    };
+    fields: Record<string, string>;
+  }[];
+}
+
+export type ContentfulResult = { contentful: Record<'name', string>[] };
+
+export interface ContentfulClientService {
+  getEntries(search: ContentfulSearch): Observable<ContentfulResponse>;
+}
```
tobi-or-not-tobi: wondering why we have a contentful service. I thought we'd have a ContentService, with various ContentAdapters. 

So, 

Searchbox 
  -> Suggestions
    -> ContentService
       -> SAPI
       -> Contentful
       -> Stapi
       -> ...

libs/template/presets/storefront/experience/pages/article-page.ts
```diff
@@ -0,0 +1,25 @@
+import { StaticComponent } from '@spryker-oryx/experience';
+
+export const articlePage: StaticComponent = {
+  type: 'Page',
+  meta: {
+    title: 'Article Page',
+    route: '/article/:id',
+    description: 'Default Article Page Description',
```
tobi-or-not-tobi, line 8: we should – going forward – have those being generated from the content

tobi-or-not-tobi (changes requested):

apps/storefront/.env.template
supproduction: I just unified. Cause all examples are not commented out

libs/domain/search/src/services/suggestion/renderer/suggestion-renderer.service.ts
supproduction: I don't fully get it. I guess we have to have a sync regarding it

libs/template/labs/src/contentful/client/contentful-client.service.ts
supproduction: But it's client service. I just created it to reuse it inside  Suggestions and ContentService

libs/template/presets/storefront/experience/pages/article-page.ts
supproduction: agree

libs/domain/content/src/services/default-content.service.ts
```diff
@@ -0,0 +1,34 @@
...
+export class DefaultContentService implements ContentService {
+  constructor(protected adapters = inject(ContentAdapter, [])) {}
+
+  protected contentQuery = createQuery<Content | null, ContentQualifier>({
+    loader: (q: ContentQualifier) =>
+      combineLatest(this.adapters.map((adapter) => adapter.get(q))).pipe(
+        map((contents) =>
+          contents.reduce(
+            (acc, curr) => ({ ...acc, ...curr }),
+            Object.create(null)
```
dunqan: Why not simply `{}` ? 🤔

libs/domain/search/src/services/adapter/default-suggestion.adapter.ts
```diff
@@ -1,14 +1,14 @@
...
   ApiSuggestionModel,
-  Suggestion,
   SuggestionQualifier,
+  SuggestionResponse,
 } from '../../models';
 import { SuggestionNormalizer } from './normalizers';
-import { SuggestionAdapter } from './suggestion.adapter';
+import { SuggestionAdapter, SuggestionField } from './suggestion.adapter';
 
 export class DefaultSuggestionAdapter implements SuggestionAdapter {
```
dunqan, line 13: Should it be named default if it's a posible one of many?

libs/domain/search/src/services/adapter/default-suggestion.adapter.ts
```diff
@@ -23,20 +23,33 @@ export class DefaultSuggestionAdapter implements SuggestionAdapter {
...
+      entries.some((entry) =>
+        [
+          SuggestionField.Categories,
+          SuggestionField.Completion,
+          SuggestionField.Products,
+        ].includes(entry as SuggestionField)
       )
-      .pipe(this.transformer.do(SuggestionNormalizer));
+    ) {
+      const include = [
```
dunqan: can we optimize this one, if no products are needed, we don't have to include them?

libs/domain/search/src/services/suggestion/renderer/suggestion.renderer.ts
```diff
@@ -0,0 +1,102 @@
...
+              type: SemanticLinkType.ProductList,
+              params: { q: query },
+            }}
+            .content=${{ text: i18n('search.box.view-all-products') }}
+          ></oryx-content-link>
+        </oryx-button>
+      </section>
+    `;
+  };
+  const renderProduct = (product: Product): TemplateResult => {
```
dunqan, line 37: Wondering about separating them out, and move product renderer to product lib?

libs/template/labs/src/contentful/client/contentful-client.service.ts
dunqan: Maybe renaming to something like ContentfulApiService would make it more clear

libs/template/labs/src/contentful/contentful.providers.ts
```diff
@@ -0,0 +1,62 @@
...
+  },
+  {
+    provide: FeatureOptions,
+    useValue: {
+      'oryx-search-box': {
+        entries: [
+          SuggestionField.Completion,
+          SuggestionField.Categories,
+          SuggestionField.Products,
+          'contentful',
```
dunqan: I thought it will have 'articles' here, not 'contentful'?

libs/domain/content/src/services/default-content.service.ts
supproduction: Just because of avoiding TS typecasting

dunqan: I'd prefer typecasting instead of this, as it has 0 runtime overhead.

libs/domain/content/article/article.model.ts
supproduction: changed it to entities

libs/template/presets/storefront/experience/pages/article-page.ts
supproduction: But I guess this is should be part of article component when we implement it

libs/domain/search/src/services/adapter/default-suggestion.adapter.ts
supproduction: But it's default spryker adapter. Not sure if I can use Spryker prefix in it

libs/domain/search/src/services/suggestion/renderer/suggestion.renderer.ts
supproduction: Basically this renderer already changed @tobi-or-not-tobi in another PR

apps/storefront/.env.template
tobi-or-not-tobi: ok, fair. maybe we should also drop the value then.

apps/storefront/.env.template
tobi-or-not-tobi: nice!

libs/domain/search/box/box.component.ts
```diff
@@ -29,21 +32,17 @@ import { LitElement, TemplateResult } from 'lit';
...
-  entries: [
-    SuggestionField.Completion,
-    SuggestionField.Categories,
-    SuggestionField.Products,
-  ],
+  [`${SuggestionField.Suggestions}Count`]: 5,
+  [`${SuggestionField.Products}Count`]: 5,
+  [`${SuggestionField.Categories}Count`]: 5,
+  [`${SuggestionField.Articles}Count`]: 5,
+  entities: defaultEntities,
```
tobi-or-not-tobi: what about using the count as an indicator that we need to take the entity into account? Then we can drop the list of entities. Or we introduce a more complex object:

```
@defaultOptions({
  minChars: 2,
  product: {
    max: 5
  }
});
```

With the later, we can also come up with more complex strategies, i.e. to not search on mobile for some content as it's often too busy on a small device.

libs/domain/search/src/services/adapter/default-suggestion.adapter.ts
tobi-or-not-tobi: `Default` prefix is indeed the convention we took for all Spryker integrations. We have considered `Glue` prefix, but i'm glad we didn't as Glue is more a framework then a specific API.

libs/domain/search/src/services/suggestion/renderer/suggestion.renderer.ts
tobi-or-not-tobi: I'm still wondering if we can improve this area going forward, but happy to first let this land and unblock us with the (urgently needed) new UI. we have a cleaner picture then and can further refine.

libs/template/labs/src/contentful/contentful.providers.ts
```diff
@@ -1,22 +1,33 @@
...
-import {
-  ContentfulSuggestionRevealer,
-  DefaultContentfulSuggestionAdapter,
-} from './suggestion';
+
+let logged = 0;
+const logMissingEnv = (): void => {
+  if (logged > 0) return;
+  console.warn(
+    'Missing ORYX_CONTENTFUL_TOKEN or/and ORYX_CONTENTFUL_SPACE environment variables'
```
tobi-or-not-tobi:
```suggestion
    'Missing ORYX_CONTENTFUL_TOKEN and/or ORYX_CONTENTFUL_SPACE environment variables'
```

nitpick

tobi-or-not-tobi (changes requested):

libs/domain/search/box/box.component.ts
tobi-or-not-tobi: The idea with this one was to avoid `entities: defaultEntities`, as the existence of the object already indicates that the entity is required.

tobi-or-not-tobi (approved):

libs/domain/search/src/services/adapter/default-suggestion.adapter.ts
dunqan: 👍

PR #492, opened 2023-06-22
feat(i18n): add support for links in translations

gund:
- [x] Add transformation of links to a default `i18nInjectable`
- [x] Add support for HTML in `i18n` directive
- [x] Implement unit tests

Closes [HRZ-1998](https://spryker.atlassian.net/browse/HRZ-1998)

[HRZ-1998]: https://spryker.atlassian.net/browse/HRZ-1998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/base/utilities/src/directives/i18n/i18n.directive.ts
+++ b/libs/base/utilities/src/directives/i18n/i18n.directive.ts
@@ -39,7 +39,7 @@ export function i18n<T extends string | readonly string[]>(
   const result = i18nInjectable.get().translate<T>(token, context);
 
   if (typeof result === 'string') {
-    return asUnsafeHtml(result);
+    return result;
   }
 
   if ('text' in result) {
--- a/libs/base/utilities/src/directives/i18n/i18n.directive.ts
+++ b/libs/base/utilities/src/directives/i18n/i18n.directive.ts
@@ -44,15 +44,11 @@ export function i18n<T extends string | readonly string[]>(
 
   if ('text' in result) {
     if (result.hasHtml && typeof result.text === 'string') {
-      return asUnsafeHtml(result.text);
+      return html`${unsafeHTML(result.text)}`;
     } else {
       return result.text;
     }
   }
 
   return result;
-}
-
-function asUnsafeHtml(text: string) {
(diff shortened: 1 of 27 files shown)


libs/base/utilities/src/directives/i18n/i18n.directive.ts
```diff
@@ -33,6 +35,24 @@ import { i18nInjectable, InferI18nContext } from '../../injectables';
...
+    } else {
+      return result.text;
+    }
+  }
+
+  return result;
+}
+
+function asUnsafeHtml(text: string) {
+  return html`${unsafeHTML(text)}`;
```
supproduction: Do we need separate function for it?

libs/platform/i18n/project.json
```diff
@@ -7,7 +7,12 @@
       "executor": "./tools/executors/components-library:build"
     },
     "test": {
-      "executor": "./tools/executors/oryx-vite:vitest"
+      "executor": "./tools/executors/oryx-vite:vitest",
+      "configurations": {
+        "watch": {
+          "watch": true
+        }
+      }
```
supproduction, line 15: why do we need it? 
we can just use `--watch`

supproduction (changes requested):

libs/base/utilities/src/directives/i18n/i18n.directive.ts
gund: Yeah I previously had 2 places where I was using it but after refactoring it's not needed so will remove it 👌

libs/platform/i18n/project.json
gund: Well I'm using NX Console extension to launch nx commands and it does not support writing extra flags so it's nice to have a shortcut to be able to launch watch mode in one click 😊

supproduction (dismissed):

dunqan (dismissed):

dunqan (dismissed):

dunqan (approved):

PR #496, opened 2023-06-22
feat: fullscreen modal size on small screen

tobi-or-not-tobi:
Modal width is aligned to avoid layout shifts from modal to modal. Also, modals will default to fullscreen on small screens, unless the `minimal` property is used.

closes [HRZ-3106](https://spryker.atlassian.net/browse/HRZ-3106)

BREAKING CHANGE: fullscreen property is removed; fullscreen is now the default on small screens.

[HRZ-3106]: https://spryker.atlassian.net/browse/HRZ-3106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sinucid (changes requested):
Looks good, but seems to be stories are broken, breakpoint for small screens is not applied correctly

Sinucid (approved):

PR #507, opened 2023-06-26
fix: cart loading issues 

dunqan:
Create cart on first add entry for registered user

### DIFF
--- a/libs/domain/cart/src/services/adapter/default-cart.adapter.ts
+++ b/libs/domain/cart/src/services/adapter/default-cart.adapter.ts
@@ -37,6 +37,7 @@ export class DefaultCartAdapter implements CartAdapter {
     return this.identity.get().pipe(
       take(1),
       switchMap((identity) => {
+        console.log('identity', identity);
         const url = this.generateUrl(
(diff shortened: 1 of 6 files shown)


libs/domain/cart/src/services/adapter/default-cart.adapter.ts
```diff
@@ -99,6 +111,34 @@ export class DefaultCartAdapter implements CartAdapter {
     );
   }
 
+  protected createCartIfNeeded(
+    identity: AuthIdentity,
+    cartId: string | undefined
+  ): Observable<[AuthIdentity, string | undefined]> {
+    if (!identity.isAuthenticated || cartId) return of([identity, cartId]);
+
+    // if we ara registered user and we do not have a cartId, we need to create a cart first
```
tobi-or-not-tobi:
```suggestion
    // if we are a registered user and we do not have a cartId, we need to create a cart first
```

tobi-or-not-tobi (dismissed):

tobi-or-not-tobi (approved):

PR #510, opened 2023-06-27
fix: sorting form is not confirming by "Enter" key

tarashaliuk:
closes: [HRZ-3087](https://spryker.atlassian.net/browse/HRZ-3087)

[HRZ-3087]: https://spryker.atlassian.net/browse/HRZ-3087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/picking/src/components/filters/filters.component.ts
+++ b/libs/domain/picking/src/components/filters/filters.component.ts
@@ -24,11 +24,6 @@ export class FiltersComponent extends LitElement {
   protected fieldRenderer = resolve(FormRenderer);
   protected pickingListService = resolve(PickingListService);
 
-  disconnectedCallback(): void {
-    super.disconnectedCallback();
-    this.form?.removeEventListener('keydown', this.onFormKeyDown);
-  }
-
   protected $selectedSortingValue = signal(
     this.pickingListService.getSortingQualifier().pipe(map(this.formatValue)),
     { initialValue: this.formatValue(defaultSortingQualifier) }
@@ -67,7 +62,10 @@ export class FiltersComponent extends LitElement {
     this.form?.reset();
   }
 
-  protected onApply(): void {
+  protected onApply(event: Event): void {
+    if (event instanceof KeyboardEvent && event.key !== 'Enter') return;
+    event.preventDefault();
+
     //For safari 15- and other old browsers
     if (!this.form?.requestSubmit) {
       this.form?.submit();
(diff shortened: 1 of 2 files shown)


libs/domain/picking/src/components/filters/filters.component.ts
```diff
@@ -90,14 +90,14 @@ export class FiltersComponent extends LitElement {
           <button>${i18n('picking.filter.reset')}</button>
         </oryx-button>
 
-        <form @submit=${this.onSubmit}>
+        <form @submit=${this.onSubmit} id="filter-form">
```
supproduction: Why do we need form id?

supproduction (changes requested):

libs/domain/picking/src/components/filters/filters.component.ts
tarashaliuk: To connect form with button which is outside of the form. To make it we need 'id' and 'form' attributes

supproduction: Can we use kepress listener on this button and don't pollute html with unnecessary id's?

tarashaliuk: But the current approach is more HTML. It generates the additional `id` on the form but reduces using JS code.
@tobi-or-not-tobi what do you think?

tobi-or-not-tobi: we do indeed not like IDs, even generated. Lets try an alternative solution, let me know if you need help.

libs/domain/picking/src/components/filters/filters.component.ts
```diff
@@ -72,6 +77,13 @@ export class FiltersComponent extends LitElement {
     this.form.requestSubmit();
   }
 
+  protected onFormKeyDown(event: KeyboardEvent): void {
+    if (event.key === 'Enter') {
+      event.preventDefault();
+      this.onApply();
```
supproduction: Please move `if (event.key === 'Enter') {event.preventDefault()} ` to apply method

libs/domain/picking/src/components/filters/filters.component.ts
```diff
@@ -24,6 +24,11 @@ export class FiltersComponent extends LitElement {
   protected fieldRenderer = resolve(FormRenderer);
   protected pickingListService = resolve(PickingListService);
 
+  disconnectedCallback(): void {
+    super.disconnectedCallback();
+    this.form?.removeEventListener('keydown', this.onFormKeyDown);
+  }
```
supproduction: We don't need it anymore

supproduction (changes requested):

supproduction (approved):

PR #518, opened 2023-06-28
chore: lighthouse plugin should be used only for SF

AndreyYevtukhov:
Fixes Lighthouse plugin configuration so that it is being run only for SF


AndreyYevtukhov:
It does not fixe the issue with the run - so we have to make a small refactoring here (later)

PR #523, opened 2023-06-29
fix: add escape string to meta service 

supproduction:
escapes string in the meta service

closes [HRZ-3137](https://spryker.atlassian.net/browse/HRZ-3137)

[HRZ-3137]: https://spryker.atlassian.net/browse/HRZ-3137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/core/src/services/page-meta/default-page-meta.service.ts
+++ b/libs/platform/core/src/services/page-meta/default-page-meta.service.ts
@@ -102,6 +102,13 @@ export class DefaultPageMetaService implements PageMetaService {
   }
 
   protected escapeValue(value: string): string {
-    return value.replace(/"/g, '&quot;');
+    return value
+      .replace(/"/g, '&quot;')
+      .replace(/'/g, '&apos;')
+      .replace(/'/g, '&amp;')
+      .replace(/</g, '&lt;')
+      .replace(/>/g, '&gt;')
+      .replace(/—/g, '&mdash;')
+      .replace(/–/g, '&ndash;');
   }
 }
--- a/libs/platform/core/src/services/page-meta/default-page-meta.service.ts
+++ b/libs/platform/core/src/services/page-meta/default-page-meta.service.ts
@@ -102,6 +102,6 @@ export class DefaultPageMetaService implements PageMetaService {
   }
 
   protected escapeValue(value: string): string {
-    return value.replace(/"/g, '\\$&');
+    return value.replace(/"/g, '&quot;');
   }
 }
--- a/libs/platform/core/src/services/page-meta/default-page-meta.service.ts
+++ b/libs/platform/core/src/services/page-meta/default-page-meta.service.ts
@@ -69,7 +69,7 @@ export class DefaultPageMetaService implements PageMetaService {
       }
 
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      element.setAttribute(key, value!);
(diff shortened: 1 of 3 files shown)


libs/platform/core/src/services/page-meta/default-page-meta.service.ts
```diff
@@ -93,10 +93,15 @@ export class DefaultPageMetaService implements PageMetaService {
...
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      attrs += `[${key}="${this.escapeValue(value!)}"]`;
     }
 
     const name = this.getTagName(definition.name);
     return document.head.querySelector(`${name}${attrs}`);
   }
+
+  protected escapeValue(value: string): string {
+    return value.replace(/"/g, '\\$&');
```
tobi-or-not-tobi: Can we cover this with a test? 
And perhaps we test other characters that could break it?

tobi-or-not-tobi: I believe we should replace it with `&quot;` as other wise search engines will break, see [https://support.sitebeat.com/help/what-is-a-page-meta-title-description and https://stackoverflow.com/questions/1137045/meta-description-quotes-escape

tobi-or-not-tobi (changes requested):

libs/platform/core/src/services/page-meta/default-page-meta.service.ts
supproduction: Only quotes might break it, cause we create selector `key="${value}"`

libs/platform/core/src/services/page-meta/default-page-meta.service.ts
```diff
@@ -93,10 +93,15 @@ export class DefaultPageMetaService implements PageMetaService {
...
+      attrs += `[${key}="${this.escapeValue(value!)}"]`;
     }
 
     const name = this.getTagName(definition.name);
     return document.head.querySelector(`${name}${attrs}`);
   }
+
+  protected escapeValue(value: string): string {
+    return value.replace(/"/g, '&quot;');
+  }
```
tobi-or-not-tobi: > Here are some characters that you should consider > encoding:
> 
> Double quotation marks: As you mentioned, using &quot; is > the correct way to encode double quotation marks within > HTML attributes.
> 
> Single quotation marks: Similar to double quotation > marks, you can encode single quotation marks as &apos; or > use the numeric character reference &#39;.
> 
> Ampersand: The ampersand character itself is special in > HTML and is used to denote character entities. To > represent an ampersand, use &amp;.
> 
> Less than and greater than signs: These characters are > used for HTML tags and can cause parsing issues if not > properly encoded. Use &lt; for < and &gt; for >.
> 
> Non-breaking space: In some cases, you may want to > include non-breaking spaces in your meta description. > Encode them as &nbsp; to ensure they are preserved.
> 
> Em dashes and en dashes: Depending on the context, em > dashes (—) and en dashes (–) may cause issues. It's > generally safe to use their respective HTML entities: &> mdash; and &ndash;.

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #527, opened 2023-06-30
fix: improved search box result UI

tobi-or-not-tobi:
The search box result panel is redesigned. The implementation now uses the product card component to render the list items. The search box comes with a backdrop (30%) as soon as the user selects the box.

closes [HRZ-3144](https://spryker.atlassian.net/browse/HRZ-3144)
fixes [HRZ-3111](https://spryker.atlassian.net/browse/HRZ-3111)

Leftovers:
- the icon-only button comes with too much padding; we need a button refactor to do this properly
- the search input has not enough height (should be `52px`)
- the margin between the box and the popover is to large; we like to do this as part of the refactor of search input control.
- there is no loading state design yet

[HRZ-3144]: https://spryker.atlassian.net/browse/HRZ-3144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-3111]: https://spryker.atlassian.net/browse/HRZ-3111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tolerants:
As long as there are ui changes it's not a pure refactoring then. So suggest to use `feat` or `fix` here.

libs/base/ui/action/link/src/styles/storefront.styles.ts
```diff
@@ -54,9 +58,9 @@ export const storefrontLinkStyles = css`
     color: var(--oryx-color-primary-9);
   }
 
-  :host([color='light']) {
+  /* :host([color='light']) {
     color: var(--oryx-color-neutral-0);
-  }
+  } */
```
supproduction: Do we need commented code?

libs/domain/search/src/services/suggestion/renderer/suggestion-renderer.service.ts
```diff
@@ -21,6 +23,7 @@ export type SuggestionRendererOptions = {
 export interface SuggestionRendererParams extends SuggestionFieldOptions {
   query: string;
   title: string;
+  icon?: string;
```
supproduction: We don't need duplicate options here as we extend it `SuggestionRendererParams extends SuggestionFieldOptions`

supproduction (changes requested):

supproduction (dismissed):

PR #535, opened 2023-06-30
fix: avoid increasing when min value = 0

tobi-or-not-tobi:
fixes: [HRZ-3146](https://spryker.atlassian.net/browse/HRZ-3146)

[HRZ-3146]: https://spryker.atlassian.net/browse/HRZ-3146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/cart/quantity-input/src/quantity-input.component.ts
+++ b/libs/domain/cart/quantity-input/src/quantity-input.component.ts
@@ -123,6 +123,7 @@ export class QuantityInputComponent
   protected isMaxDisabled(): boolean {
     return (
       this.disabled ||
+      this.min === 0 ||
       (this.value !== undefined &&
         this.max !== undefined &&
         this.value >= this.max)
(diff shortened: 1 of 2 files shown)


supproduction (dismissed):

libs/domain/cart/quantity-input/src/quantity-input.component.ts
```diff
@@ -123,6 +123,7 @@ export class QuantityInputComponent
   protected isMaxDisabled(): boolean {
     return (
       this.disabled ||
+      this.min === 0 ||
```
tarashaliuk: Why `+` button should be disabled if the `min` value is 0?

tarashaliuk (changes requested):

tarashaliuk (approved):

PR #539, opened 2023-06-30
test: fix some sf e2e test flakiness

AndreyYevtukhov:
Resolves some SF E2E tests flakiness related to login and checkout failures.
GLUE API interceptors were added + 2 different Cypress Actions for checkout, making code cleaner and more reliable.

Also, there is a workaround for issue 3157 (with Checkout Page not being rendered sometimes) - that should make tests more stable. We will remove it when the issue is fixed.

### DIFF
--- a/apps/storefront-e2e/src/support/page_fragments/header.fragment.ts
+++ b/apps/storefront-e2e/src/support/page_fragments/header.fragment.ts
@@ -56,6 +56,8 @@ export class HeaderFragment {
     // eslint-disable-next-line cypress/no-unnecessary-waiting
     cy.wait(500);
     this.getCurrencyButton().click();
-    this.getCurrencySelector().find(`oryx-option[value="${currency}"]`).click();
+    this.getCurrencySelector()
+      .find(`oryx-option[value="${currency}"]`)
+      .click({ force: true });
   };
 }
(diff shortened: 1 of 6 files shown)


apps/storefront-e2e/src/support/page_fragments/header.fragment.ts
```diff
@@ -56,6 +56,8 @@ export class HeaderFragment {
     // eslint-disable-next-line cypress/no-unnecessary-waiting
     cy.wait(500);
     this.getCurrencyButton().click();
-    this.getCurrencySelector().find(`oryx-option[value="${currency}"]`).click();
+    this.getCurrencySelector()
+      .find(`oryx-option[value="${currency}"]`)
+      .click({ force: true });
```
tolerants, line 61: Is it safe? Is I understand we are ignoring every possible accessibility check, so this element can be invisible, disabled and so on.

tolerants (changes requested):

apps/storefront-e2e/src/support/page_fragments/header.fragment.ts
AndreyYevtukhov: It is not safe for sure, but it is temporary
I'll remove this check when https://spryker.atlassian.net/browse/HRZ-3157 is fixed (added a note is a bug)

tolerants: I see.

tolerants (approved):

PR #540, opened 2023-07-03
refactor!: Refactored auth and site packages

tolerants:
Auth
- [x] Move oauth-handler component to a separate folder
- [x] Refactor login component to get rid of rx-js operators
- [x] Refactor login-link component to get rid of rx-js operators
- [x] Move components to separate sub-packages

Site
- [x] Remove usages of deprecated methods
- [x] Move components to separate sub-packages

BREAKING CHANGE: all components in auth and site packages has been moved to a separate sub-packages, so now there are not available to import from the main package.

closes: https://spryker.atlassian.net/browse/HRZ-3025, https://spryker.atlassian.net/browse/HRZ-2912, https://spryker.atlassian.net/browse/HRZ-2911, https://spryker.atlassian.net/browse/HRZ-3024


libs/platform/auth/login/login.component.ts
```diff
@@ -28,8 +29,6 @@ export class AuthLoginComponent extends ContentMixin<LoginOptions>(LitElement) {
   @property() passwordName = 'password';
   @property() rememberMeName = 'rememberme';
 
-  @state() protected isDisabled = false;
-
   protected emailInputRef = createRef<HTMLInputElement>();
   protected passwordInputRef = createRef<HTMLInputElement>();
   protected rememberInputRef = createRef<HTMLInputElement>();
```
tobi-or-not-tobi: Let's drop these `createRef` and use `query` instead:

```ts
@query(`input[name='email']`) email?: HTMLInputElement;
```

libs/platform/auth/login/login.component.ts
```diff
@@ -28,8 +29,6 @@ export class AuthLoginComponent extends ContentMixin<LoginOptions>(LitElement) {
   @property() passwordName = 'password';
   @property() rememberMeName = 'rememberme';
```
tobi-or-not-tobi: Do we need `emailName`, `passwordName`, and `rememberMeName`? If not, let's drop them.

libs/platform/auth/login/login.component.ts
```diff
@@ -41,37 +40,44 @@ export class AuthLoginComponent extends ContentMixin<LoginOptions>(LitElement) {
...
-          this.hasError = true;
-          return EMPTY;
-        })
-      )
-    ),
-    tap(() => (this.isLoading = false)),
-    switchMap(() => {
-      if (this.componentOptions?.disableRedirect) {
-        return EMPTY;
+  protected async doLogin(loginState: LoginRequest): Promise<void> {
```
tobi-or-not-tobi, line 39: I wonder if we cannot do this in a simplified way:
- Store the state in a signal; update the state on login
- Use an ElementEffect to observe the state, and act accordingly
- Use a signal to observe whether the user is authenticated (`protected isAuthenticated = signal(this.authService.isAuthenticated());`)
- Use an ElementEffect to observe the authenticated state (redirect)

We've done something similar at the account component in checkout.

libs/platform/auth/login/login.component.ts
```diff
@@ -41,37 +40,44 @@ export class AuthLoginComponent extends ContentMixin<LoginOptions>(LitElement) {
...
+
+      const previousRoute = await firstValueFrom(
+        this.routerService.previousRoute()
+      );
+      const redirectRoute = previousRoute ? previousRoute : '/';
+      this.routerService.navigate(redirectRoute);
+    }
+  }
+
+  protected login(event: Event): void {
```
tobi-or-not-tobi:
```suggestion
  protected onSubmit(event: Event): void {
```

tobi-or-not-tobi (changes requested):

libs/platform/auth/login/login.component.ts
tolerants: > Store the state in a signal; update the state on login
But here state is a state of the form, we don't need to observe it actually.

> update the state on login
Here is the main issue, as long as we have a loginStrategy where we can have it's own redirect login.

tobi-or-not-tobi (approved):

PR #550, opened 2023-07-04
fix: avoid closing the search popover unexpectedly

tobi-or-not-tobi:
The popover no longer closes unexpectedly. When the user clicks inside the search popover, it must only close when a link is clicked. When the user clicks on the layout or interactive elements (i.e. add to cart), the popover must remain open. 

fixes: [HRZ-3163](https://spryker.atlassian.net/browse/HRZ-3163)

[HRZ-3163]: https://spryker.atlassian.net/browse/HRZ-3163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/base/ui/overlays/popover/src/controllers/toggle.controller.ts
+++ b/libs/base/ui/overlays/popover/src/controllers/toggle.controller.ts
@@ -143,10 +143,10 @@ export class ToggleController implements ReactiveController {
   protected async handleMouseup(e: MouseEvent): Promise<void> {
     this.focusShouldBeFocusedMaybe();
 
-    const outside =
+    const isEmitterOutside =
       timePassed(this.timeStarted) && !this.emitterIsInsidePopover(e);
 
-    if ((await this.shouldClosePopover(e)) || outside) {
+    if ((await this.shouldClosePopover(e)) || isEmitterOutside) {
       this.toggle(false);
     }
   }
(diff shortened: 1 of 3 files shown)


libs/base/ui/overlays/popover/src/controllers/toggle.controller.ts
```diff
@@ -143,10 +143,10 @@ export class ToggleController implements ReactiveController {
   protected async handleMouseup(e: MouseEvent): Promise<void> {
     this.focusShouldBeFocusedMaybe();
 
-    if (
-      (await this.shouldClosePopover(e)) ||
-      (timePassed(this.timeStarted) && !this.emitterIsInsidePopover(e))
-    ) {
+    const outside =
```
tolerants: First of all I'd rename it to `isEmitterOutside` or `isTargetOutside`. And also first part of this condition doesn't look logical and self-explainable to be assigned to the `outside` variable. Like why if time is passed then something is outside?

tolerants (changes requested):

libs/base/ui/overlays/popover/src/controllers/toggle.controller.ts
tobi-or-not-tobi: Renamed. The timeStarted comes with already with a small docs that explains it. Maybe still a bit of mystery, we could add more docs to the method. Here's what's going on:
- The user clicks on the search box
- The search box should not instantly close on mouseup event - however, if the user clicks again on the search box it must close.

tolerants (approved):

PR #551, opened 2023-07-04
feat: add storyblok integration

supproduction:
adds storyblok integration
adds dynamic cmd header\description

closes [HRZ-3142](https://spryker.atlassian.net/browse/HRZ-3142) [HRZ-3150](https://spryker.atlassian.net/browse/HRZ-3150)

[HRZ-3142]: https://spryker.atlassian.net/browse/HRZ-3142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HRZ-3150]: https://spryker.atlassian.net/browse/HRZ-3150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/.env.template
```diff
@@ -23,3 +23,6 @@ ORYX_CONTENTFUL_TOKEN=
 # All content and assets in Contentful belong to a space.
 # You will generally have at least one space for a project.
 ORYX_CONTENTFUL_SPACE=
+
+# The Contentful Access Token
```
tobi-or-not-tobi:
```suggestion
# The Storyblok Access Token
```

libs/domain/content/article/article.component.ts
```diff
@@ -26,27 +26,32 @@ export class ArticleComponent extends ContentMixin<ContentArticleOptions>(
...
     const data = this.$data();
 
-    if (!data?.article) {
+    if (!data?.content) {
       return;
     }
 
     return html`<oryx-text
-      .content=${marked.parse(data.article)}
+      .content=${marked.parse(data.content)}
```
tobi-or-not-tobi, line 54: wondering if we should not move markdown conversion to the adapter layer, since we do not know which element comes with which type of content.

libs/domain/content/src/models/content.model.ts
```diff
@@ -1,10 +1,13 @@
...
-  article?: string;
+  type?: string;
+  id?: string;
   entities?: ContentEntities;
 }
 
 export interface Content {
-  article?: string;
+  heading: string;
+  description: string;
```
tobi-or-not-tobi, line 11: should we introduce a `meta` object here, and store description in there?

libs/domain/search/src/services/suggestion/default-suggestion.service.ts
```diff
@@ -16,9 +16,19 @@ export class DefaultSuggestionService implements SuggestionService {
       combineLatest(
         this.adapters.map((adapter) => adapter.get(qualifier))
       ).pipe(
-        map((suggestions) =>
-          suggestions.reduce((acc, curr) => ({ ...acc, ...curr }), {})
-        )
+        map((suggestions) => {
+          console.log(suggestions);
```
tobi-or-not-tobi: left over

libs/template/labs/src/articles/storyblok/api/default-storyblok-api.service.ts
```diff
@@ -0,0 +1,45 @@
...
+    return this.locale
+      .get()
+      .pipe(
+        switchMap((locale) =>
+          this.http.get<StoryblokEntryResponse>(
+            `${this.endpoint}/${search.slug}?token=${this.storyblokToken}&language=${locale}`
+          )
+        )
+      );
+  }
```
tobi-or-not-tobi: what about

```ts
  getEntries(search: StoryblokSearch): Observable<StoryblokEntriesResponse> {
    const endpoint = `${this.endpoint}?search_term=${search.query}&token=${this.storyblokToken}`;
    return this.search<StoryblokEntriesResponse>(endpoint);
  }

  getEntry(search: StoryblokSearch): Observable<StoryblokEntryResponse> {
    const endpoint = `${this.endpoint}/${search.slug}?token=${this.storyblokToken}`;
    return this.search<StoryblokEntryResponse>(endpoint);
  }

  protected search<T>(endpoint: string): Observable<T> {
    return this.locale
      .get()
      .pipe(
        switchMap((locale) =>
          this.http.get<T>(`${endpoint}&language=${locale}`)
        )
      );
  }
```

libs/template/labs/src/articles/storyblok/api/storyblok-api.service.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export interface StoryblokApiService {
+  getEntries(search: StoryblokSearch): Observable<StoryblokEntriesResponse>;
+  getEntry(search: StoryblokSearch): Observable<StoryblokEntryResponse>;
+}
+
+declare global {
+  interface InjectionTokensContractMap {
+    [StoryblokApiService]: StoryblokApiService;
```
tobi-or-not-tobi: I actually think client was a better name, I just got my self confused initially. This stuff is not an api, so we should not name it an api.

libs/template/presets/storefront/experience/footer.ts
```diff
@@ -23,10 +23,10 @@ const legalLinks = {
   name: 'legal links',
   components: [
     link('©️ 2023 Spryker'),
-    link('Imprint'),
+    link('Imprint', '/faq/imprint'),
     link('Terms & conditions', '/article/terms-and-conditions'),
     link('Privacy Notice', '/article/privacy'),
-    link('Data preference'),
+    link('Data preference', '/faq/data-preference'),
```
tobi-or-not-tobi: There's no content for this link

libs/template/presets/storefront/experience/pages/article-page.ts
```diff
@@ -2,11 +2,27 @@ import { StaticComponent } from '@spryker-oryx/experience';
 
 export const articlePage: StaticComponent = {
   type: 'Page',
-  meta: {
-    title: 'Article Page',
-    route: '/article/:id',
-    description: 'Default Article Page Description',
+  meta: { route: '/article/:id' },
```
tobi-or-not-tobi, line 5: we should actually move those pages to labs package for the time being, as it's not yet part of final release.

tobi-or-not-tobi (changes requested):

libs/domain/content/article/article.component.ts
supproduction: As it's temporary implementation I thought it's ok if component handle it. OFC in real world we should do it in normalizers

libs/domain/content/src/models/content.model.ts
supproduction: not sure, IMO plain object simplier for understanding. But I can introduce if we need it

libs/domain/content/article/article.component.ts
supproduction: What do you think if we move this component into labs domain too?

tobi-or-not-tobi (approved):

PR #573, opened 2023-07-07
ci: fixes smoke tests grouping error in netlify

AndreyYevtukhov:
Fixes an issue with smoke tests run after adding grouping https://github.com/spryker/oryx/pull/563 
Cypress can't pick up build id on Netlify automatically, so we have to set that variable

closes: -

---

### Checklist before requesting a code review

- [x] PR title is aligned with [PR titles and conventional commits](https://spryker.atlassian.net/wiki/spaces/FA/pages/3821175291/PR+titles+and+conventional+commits)
- [x] I have performed a self-review of my code and have fixed all jobs failures

### DIFF
--- a/tools/netlify-plugins/cypress-tests-runner/index.js
+++ b/tools/netlify-plugins/cypress-tests-runner/index.js
@@ -14,6 +14,8 @@ export const onSuccess = ({ inputs }) => {
   process.env.CYPRESS_BASE_URL = deployURL;
   process.env.CYPRESS_BUILD_ID = process.env.BUILD_ID;
 
+  console.log(process.env)
+
   // sync call is needed here, because if async is used -> netlify will kill child process
   // when the main process is over
   const childProcess = spawnSync('npx', ['nx', 'run', nxRunCommand], { stdio: 'inherit' });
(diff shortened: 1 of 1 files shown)


tools/netlify-plugins/cypress-tests-runner/index.js
```diff
@@ -1,26 +1,58 @@
...
+ * modify nx config a bit to add/edit missing values
+ *
+ * @param {string} buildId
+ * @param {string} baseUrl
+ */
+function updateCypressConfig(buildId, baseUrl) {
+  console.log(`Deployed site URL: ${baseUrl}`);
+  console.log(`Deployed build id: ${buildId}`);
+
+  const pathToProjectConfig = '../../apps/storefront-e2e/project.json';
```
tolerants, line 27: We are hardcodeing path to the storefront `project.json`, but at the same time it looks like this script is a general one, how we will run it for PA or testfront?

AndreyYevtukhov: For testfront it will be similar because both storefront and testfront use the same e2es and the same configs

As for the fulfillment app - you are right, we will have to change it. But there are no smoke tests yet, and I'm not even sure when we will be ready to add those. And when we will have B2C and B2B envs to run tests against them we will have to extend this plugin even further (I'm pretty sure there will be some additional configs and env variables)

tolerants (approved):

PR #579, opened 2023-07-08
feat(experience): configurable typography for compositions

tobi-or-not-tobi:
The footer links must be of size `small`. In order to simplify this, we introduced typography through composition styles.

closes: [HRZ-3187](https://spryker.atlassian.net/browse/HRZ-3187)

BREAKING CHANGE: we've moved the global font size of the html the root of the document (using `ThemeMetaInitializer`, so that we can move our typography font-size tokens to `rem` rather than `em`.  Previously, all typography was based on `em`. 

[HRZ-3187]: https://spryker.atlassian.net/browse/HRZ-3187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/src/models/composition.model.ts
```diff
@@ -208,6 +209,13 @@ export interface StyleProperties extends LayoutStylesProperties {
    */
   scale?: number;
 
+  /**
+   * Specifies the typography tag, which is one of the available heading tags.
+   * The typography is applied to the component and all it's sub components, including those components
+   * with a shadow dom, since typography is inherited in shadow dom.
+   */
+  typo?: HeadingTag;
```
tolerants: Suggest to use the full word here to exclude incorrect interpretations. Maybe it's even better to use rename it for better understanding - `defaultTypographyTag` for example.

tolerants (approved):

PR #590, opened 2023-07-11
fix: wrong locale is set after choosing a locale 

tolerants:
closes: [HRZ-3223](https://spryker.atlassian.net/browse/HRZ-3223)

[HRZ-3223]: https://spryker.atlassian.net/browse/HRZ-3223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/i18n/src/lib/locale/default-locale.service.ts
```diff
@@ -28,18 +31,11 @@ export class DefaultLocaleService implements LocaleService {
     protected queryService = inject(QueryService)
   ) {
     this.active$ = concat(
-      this.adapter.getDefault().pipe(takeUntil(this.setActive$)),
+      this.adapter.getDefault().pipe(takeUntil(this.setActive$.pipe(skip(1)))),
```
dunqan, line 34: I think we could keep replay subject and then we should not need changes in this stream (skip(1) and filter)

tolerants: But in this case we will not be able to simply get a value in `set` method. So it will a be a tradeoff in any case.

libs/platform/i18n/src/lib/locale/default-locale.service.spec.ts
```diff
@@ -58,25 +58,6 @@ describe('DefaultLocaleService', () => {
 
       expect(cb).toHaveBeenCalledWith('mock-locale');
     });
-
-    it('should not emit LocaleChanged query with locale initially', () => {
```
dunqan, line 62: I'd expect some more unit tests after a bug fix, not only removing existing ones ;)

tolerants: But it's already covered in `set` section.

tobi-or-not-tobi (approved):

PR #596, opened 2023-07-12
test: adjust locales, currencies and add to cart tests a bit

AndreyYevtukhov:
Improve some actions related to locales, currencies and add to cart tests that should make e2es more stable 

closes: -

### DIFF
--- a/apps/storefront-e2e/src/support/page_fragments/header.fragment.ts
+++ b/apps/storefront-e2e/src/support/page_fragments/header.fragment.ts
@@ -36,7 +36,7 @@ export class HeaderFragment {
   changeLocale = (locale: string) => {
     this.initProductsUpdateInterceptor();
 
-    this.getLocaleButton().trigger('mouseenter').wait(75).click();
+    this.getLocaleButton().trigger('mouseenter').wait(150).click();
 
     this.getLocaleSelector()
       .find(`oryx-option[value="${locale}"]`)
@@ -49,7 +49,7 @@ export class HeaderFragment {
   changeCurrency = (currency: string) => {
     this.initProductsUpdateInterceptor();
 
-    this.getCurrencyButton().trigger('mouseenter').wait(75).click();
+    this.getCurrencyButton().trigger('mouseenter').wait(150).click();
 
     this.getCurrencySelector()
       .find(`oryx-option[value="${currency}"]`)
--- a/apps/storefront-e2e/src/support/page_fragments/header.fragment.ts
+++ b/apps/storefront-e2e/src/support/page_fragments/header.fragment.ts
@@ -34,30 +34,39 @@ export class HeaderFragment {
   };
 
   changeLocale = (locale: string) => {
-    cy.intercept({
-      method: 'GET',
-      url: /(\/catalog-search.*|\/concrete-products\/.*)/,
-    }).as('productOrSearchRequests');
-    // hydrate
-    this.getLocaleButton().click();
-    // eslint-disable-next-line cypress/no-unnecessary-waiting
-    cy.wait(500);
-    this.getLocaleButton().click();
-    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    this.initProductsUpdateInterceptor();
+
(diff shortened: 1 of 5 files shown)


apps/storefront-e2e/src/support/page_fragments/header.fragment.ts
```diff
@@ -34,30 +34,39 @@ export class HeaderFragment {
...
-    }).as('productOrSearchRequests');
-    // hydrate
-    this.getLocaleButton().click();
-    // eslint-disable-next-line cypress/no-unnecessary-waiting
-    cy.wait(500);
-    this.getLocaleButton().click();
-    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    this.initProductsUpdateInterceptor();
+
+    this.getLocaleButton().trigger('mouseenter').wait(75).click();
```
tolerants: I'd probably move this `75` somewhere to utility layer, so to have a constant for ui delay.

tolerants: What is the trick - why we should trigger mouse enter first and wait a bit and then click on it?

tolerants (dismissed):

apps/storefront-e2e/src/support/page_fragments/header.fragment.ts
AndreyYevtukhov: The `mouseenter` event triggers component hydration 

The wait is needed to postpone a click on the element while `js` file is being loaded and the component is re-rendered

I've tried to solve it in a better way (without waiting) by checking the `defer-hydration` attribute existence on the element - but it didn't work. Let's have it as it is now, and when Stan is back maybe we will try to improve it further

AndreyYevtukhov: I'm not sure that this number is consistent and will be used somewhere else(
It's more like a temporary solution for hydration, not a final one...

apps/storefront-e2e/src/support/page_fragments/search.fragment.ts
```diff
@@ -22,25 +22,25 @@ export class SearchFragment {
...
-      )}*`
-    ).as('searchQuery');
+    const encoded = encodeURIComponent(text).replace(/%20/g, '+');
+
+    cy.intercept(`/catalog-search-suggestions?q=${encoded}*`).as('searchQuery');
 
     // trigger hydration
-    this.getTypeahead().click();
-    // wait for hydration
+    cy.intercept('/assets/box.component-*.js').as('searchHydationFiles');
```
tolerants, line 30: But we shouldn't rely on details of how assets are bundled.... and why this component has been chosen? :)

apps/storefront-e2e/src/support/page_fragments/search.fragment.ts
```diff
@@ -22,25 +22,25 @@ export class SearchFragment {
...
+    this.getInput().click();
+    cy.wait('@searchHydationFiles');
+
+    // wait till search elements are re-rendered
     // eslint-disable-next-line cypress/no-unnecessary-waiting
-    cy.wait(500);
-    this.getInput().type(text, { delay: 10, force: true });
+    cy.wait(250);
```
tolerants, line 36: All these magic 250, 150, 125, 75 delays are killing me :)

tolerants (changes requested):

apps/storefront-e2e/src/support/page_fragments/search.fragment.ts
AndreyYevtukhov: Yep, that looks weird, but that's the only thing which proves that hydration was triggered 

I've just checked which one is being loaded all the time on search hydration + this one is the last one

apps/storefront-e2e/src/support/page_fragments/search.fragment.ts
AndreyYevtukhov: I hope we will be able to get rid of all of these someday

tolerants (approved):

PR #597, opened 2023-07-12
feat(experience): components's visibility rules

Sinucid:
Dynamic visibility rules that allows to control visibility state of the components based on provided configuration

closes to [HRZ-3071](https://spryker.atlassian.net/browse/HRZ-3071)

[HRZ-3071]: https://spryker.atlassian.net/browse/HRZ-3071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/composition/src/composition.component.ts
```diff
@@ -35,6 +36,26 @@ export class CompositionComponent extends LayoutMixin(
   protected registryService = resolve(ComponentsRegistryService);
   protected layoutBuilder = resolve(LayoutBuilder);
 
+  protected $dynamicVisibilityState = signal(
+    this.contentController.dynamicVisibilityState()
+  );
+
+  @elementEffect()
+  protected toggleVisibilityState = (): void => {
+    if (!this.uid) return;
```
tobi-or-not-tobi: should we count for uid for route as well?

libs/platform/experience/composition/src/composition.component.ts
```diff
@@ -35,6 +36,26 @@ export class CompositionComponent extends LayoutMixin(
   protected registryService = resolve(ComponentsRegistryService);
   protected layoutBuilder = resolve(LayoutBuilder);
 
+  protected $dynamicVisibilityState = signal(
+    this.contentController.dynamicVisibilityState()
+  );
+
+  @elementEffect()
+  protected toggleVisibilityState = (): void => {
```
tobi-or-not-tobi: Does this mean that we only support visibility on compositions?

libs/platform/experience/composition/src/composition.component.ts
```diff
@@ -35,6 +36,26 @@ export class CompositionComponent extends LayoutMixin(
   protected registryService = resolve(ComponentsRegistryService);
   protected layoutBuilder = resolve(LayoutBuilder);
 
+  protected $dynamicVisibilityState = signal(
+    this.contentController.dynamicVisibilityState()
+  );
```
tobi-or-not-tobi: What about `$visible` and `this.contentController.isVisible()` ?

tobi-or-not-tobi (changes requested):
- can we get visibility on components?
- can we rename visibilityRules just to `visible` in most cases? Have already put a few notes on this in the comments, but it's general feedback.

libs/template/presets/storefront/experience/header.ts
```diff
@@ -93,19 +93,61 @@ export const HeaderTemplate: StaticComponent = {
...
+                  components: [{ type: 'oryx-auth-login-link' }],
+                },
+              ],
+            },
+            {
+              type: 'oryx-composition',
+              visibility: {
+                hide: false,
+                token: 'CART.!EMPTY',
+              },
```
tobi-or-not-tobi: We like visibility to be part of the options model, as the idea has always been to not have too many data fields in the component model.

tobi-or-not-tobi (changes requested):

libs/platform/experience/composition/src/composition.component.ts
Sinucid: For now - yes. I tried co combine all related logic inside content mixin, but @elementEffect in mixin caused strange glitches on ssr. I asked Stan for some help. I agree, would be nice to have more generic solution, that can be applied for any component in the future, but for now composition only

libs/platform/experience/composition/src/composition.component.ts
Sinucid: I don't think so. Route is needed to get components id by route, and then the id is stored to the uid, in result uid only important

libs/platform/experience/composition/src/composition.component.ts
Sinucid: It is not a boolean, it returns 4 types of states: none, defer, visible, hidden

none - when component does not have visibility rules
defer - when it has, but the value is not resolved yet
visible, hidden - when value is resolved

libs/template/presets/storefront/experience/header.ts
Sinucid: We discussed it with Stan, and considered that visibility config should have even separate tab in EB components configuration, and more logical to keep them separate. Agree, this is debatable and we can discuss it again

tobi-or-not-tobi: yes it might have a separate tab, similar to layout and styling, but I would still keep it in options (similar to layout and styling btw)

libs/platform/experience/composition/src/composition.component.ts
tobi-or-not-tobi: either way, i'd consider dropping `dynamic` and `state` as to me those words can be added to so much things. All is dynamic, a lot of stuff is state.

libs/template/presets/storefront/experience/header.ts
Sinucid: I'm not against this, but i'm a little worried that we are polluting options with reserved fields. We already have rules for styling, and plan to add visibility there, this can cause potential issues, if some component have its own options with same names

libs/platform/core/src/services/hydration/default-hydration.service.ts
```diff
@@ -13,6 +13,8 @@ import { ComponentsPlugin } from '../../orchestration/components';
 import { HydrationService, HydrationTrigger } from './hydration.service';
 import { ContextService } from '../context';
 
+export const visibilityAttribute = 'dynamic-visibility';
```
dunqan: Wondering, if we could reuse existing hydratable directive, with some additional new value. Take a look also at changes here: https://github.com/spryker/oryx/pull/602

libs/platform/core/src/services/hydration/default-hydration.service.ts
```diff
@@ -57,6 +59,11 @@ export class DefaultHydrationService implements HydrationService, OnDestroy {
         return;
       }
 
+      //need to hydrate elements with dynamic visibility rule on load
```
dunqan: actually, we could come up with an idea, that composition that knows, that has dynamic visibility, can just setup hydration for window:onload. No need to pollute hydratable durective then.

libs/platform/core/src/services/token-resolver/default-token-resolver.service.ts
```diff
@@ -1,21 +1,49 @@
...
+import { map, of } from 'rxjs';
 import {
   ResolvedToken,
   TokenResolver,
   TokenResourceResolver,
   TokenResourceResolvers,
 } from './token-resolver.service';
 
-const tokenRE = /^[A-Z_-]+\.[A-Z_-]+$/;
+// const tokenRE = /^[A-Z_-]+\.((NOT\.)?)[A-Z_-]+(?<!NOT)$/;
```
dunqan: leftover?

libs/platform/experience/src/models/composition.model.ts
```diff
@@ -4,6 +4,12 @@ import { LayoutAttributes } from '../../layout/src';
 
 export interface CompositionProperties {
   rules?: StyleRuleSet[];
+  visibility?: ComponentVisibility;
```
tobi-or-not-tobi, line 7: Is this model flexible enough to support:
- multiple conditional rules, e.g.: `hide when x,  show when y`; `hide when x AND y`; `hide when x OR y`.
- parameterized rules, e.g.: `hide when route = checkout` (semantic link ref); `show when cart entries > 5`; `hide when screen size = sm`

I'm not asking to build anything link this, just to ensure that we can extend the data model going forward without breaking changes.

libs/platform/experience/src/models/composition.model.ts
```diff
@@ -4,6 +4,12 @@ import { LayoutAttributes } from '../../layout/src';
...
+}
+
+export interface ComponentVisibility {
+  hide?: boolean;
+  hideRule?: string;
```
tobi-or-not-tobi: what about dropping `hide` and use a hideRule for that? `hideAlways` or smth

libs/platform/experience/src/services/experience/default-experience.service.ts
```diff
@@ -195,6 +195,27 @@ export class DefaultExperienceService implements ExperienceService {
...
+  // }
+
+  // protected reloadVisibilityState(uid: string): void {
+  //   this.getComponent({ uid })
+  //     .pipe(take(1))
+  //     .subscribe((component) => {
+  //       this.visibilityHiddenState[uid].next(component?.visibility ?? null);
+  //     });
+  // }
+
```
tobi-or-not-tobi: what is this about?

libs/platform/experience/src/services/experience/models.ts
```diff
@@ -1,5 +1,11 @@
 import { CompositionProperties } from '../../models';
 
+export enum Visibility {
```
tobi-or-not-tobi:
```suggestion
export const enum Visibility {
```

libs/template/presets/storefront/experience/header.ts
```diff
@@ -105,7 +121,17 @@ export const HeaderTemplate: StaticComponent = {
             },
             {
               type: 'oryx-site-navigation-item',
-              id: 'mini-cart',
+              options: {
+                data: {
+                  visibility: {
+                    hideRule: 'CART.!EMPTY',
```
tobi-or-not-tobi: Alternatively this could be a parameterized rule with cart entries 0.

libs/platform/experience/src/mixins/content.mixin.ts
```diff
@@ -48,6 +51,29 @@ export const ContentMixin = <
...
+
+    //   if (visibility === DynamicVisibilityStates.None) {
+    //     this._dynamicVisibility = null;
+    //   } else {
+    //     this._dynamicVisibility =
+    //       !isServer && visibility === DynamicVisibilityStates.Visible
+    //         ? DynamicVisibilityStates.Visible
+    //         : DynamicVisibilityStates.Hidden;
+    //   }
+    // };
```
tobi-or-not-tobi: left overs I asssume

tobi-or-not-tobi (changes requested):

libs/template/presets/storefront/experience/header.ts
tobi-or-not-tobi: Yes, that's a valid concern. I had a conversation with Stan on this topic and we came out with the following conclusions:

- we like to make "visibility" part of the option rules. The option rules have a `query` property that allows to put conditions for a rule.  With this mechanism we can currently apply different style rules per screen size (aka breakpoint), or styles for the child components. 
- the visibility is one of the outcomes of the conditions that you're build. The visibility logic must remain in the composition component to ensure that the component is not rendered inside the DOM
- going forward, we probably want to move the query up, so that options can also become conditional. We will do this in follow up work
- We do not expect to support screen size driven conditions yet, but that's also one of the things next up. If we combine this with the previous point, than we can probably move the `ResizeObserver` from layout to composition, so that all options (including layout) are filtered out up front. This would simplify the stuff nicely, and make it more efficient as well. 

So, for layout we currently have this model:

```ts
const options = {
  data: {
    rules: [
      {
        layout: "carousel",
        padding: "30px 0 5px",
        align: "stretch",
      },
      {
        query: {
          breakpoint: "sm",
        },
        padding: "20px",
      },
    ],
    category: "10",
    sort: "rating",
  },
};
```

We like to move visibility (conditions) into this model:

```ts
const options = {
  data: {
    rules: [
      {
        layout: "carousel",
        padding: "30px 0 5px",
        align: "stretch",
      },
      {
        query: {
          condition: 'USER.AUTHENTICATED',
        },
        hidden: true
      },
    ],
    category: "10",
    sort: "rating",
  },
};
```

libs/platform/experience/src/services/experience/default-experience.service.ts
Sinucid: dropped

libs/platform/experience/src/mixins/content.mixin.ts
Sinucid: dropped

libs/template/presets/storefront/experience/header.ts
Sinucid: Created a follow up ticked based on your suggestion
https://spryker.atlassian.net/browse/HRZ-3268

libs/platform/experience/src/models/composition.model.ts
Sinucid: Currently multi-conditions or passing properties to the token is not supported. I believe that all of this can be achieved by extending the token resolver service, multiple conditions can be like: `CART.EMPTY&USER.AUTHANTIFICATED`, `CART.!EMPTY|USER.!AUTHANTIFICATED`. And passing params like this: `CART.ENTRIES.==.5`, `CART.ENTRIES.<=.5`. Routes are more specific I would say, each one can have its own resolver: `ROUTE.CHECKOUT`

libs/platform/experience/src/models/composition.model.ts
Sinucid: The best that I came up with: `hide` and `hideByRule`

libs/template/presets/storefront/experience/header.ts
Sinucid: For sure, when token resolvers support it

libs/platform/experience/composition/src/composition-components.controller.ts
```diff
@@ -0,0 +1,102 @@
...
+  protected componentsVisibility(): Observable<
+    Observable<Visibility>[] | null
+  > {
+    return this.components().pipe(
+      map((components) =>
+        components
+          ? components.map((c) => {
+              const visibility = c.options?.data?.visibility;
+
+              if (!visibility) {
```
dunqan, line 52: couldn't we just filter out those components? (just wondering)

libs/platform/experience/composition/src/composition-components.controller.ts
```diff
@@ -0,0 +1,102 @@
...
+              return of(Visibility.Visible);
+            })
+          : null
+      )
+    );
+  }
+
+  getComponents(): Observable<Component[] | null> {
+    return this.components().pipe(
+      combineLatestWith(this.componentsVisibility()),
```
dunqan: just a nitpick, I'd consider calling mapping function inside next switchMap, instead of using combineLatestWith, as you do combineLatest in a switch anyway.

I guess you could also drop this visibility enum, as you can filter in one go (if not visible, return null as component, and filter out all nulls for example)

libs/platform/experience/composition/src/composition.component.ts
```diff
@@ -50,15 +57,23 @@ export class CompositionComponent extends LayoutMixin(
...
   });
 
-  protected $components = computed(() => {
-    if (!this.uid) {
-      return [];
-    }
+  protected $components = signal(this.componentsController.getComponents());
+
+  protected $hasDynamicallyVisibleSuccessor = signal(
+    this.componentsController.hasDynamicallyVisibleSuccessor()
```
dunqan: is it a successor or child/subcomponent?

tobi-or-not-tobi (approved):

PR #603, opened 2023-07-13
feat: adjust header and user modal

tarashaliuk:
closes: https://spryker.atlassian.net/browse/HRZ-3115


libs/domain/picking/header/header.component.ts
aleespryker, line 1: Can we also use this in the picking lists header or is it too different there?

tarashaliuk: `header` component is used in `picking-header`

aleespryker (approved):

PR #605, opened 2023-07-13
fix(cart): support cart entry volume price

tobi-or-not-tobi:
We've ignored the volume price on cart entries so far. With this fix we show the cart unit price next to the (strikethrough) original item price. The prices are now formatted using the `SitePriceComponent`.

closes: [HRZ-3182](https://spryker.atlassian.net/browse/HRZ-3182)

[HRZ-3182]: https://spryker.atlassian.net/browse/HRZ-3182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/cart/entry/src/entry.model.ts
```diff
@@ -21,6 +21,11 @@ export interface CartEntryAttributes {
    */
   price?: number;
 
+  /**
+   * The item sales price that override the item sales price.
```
tolerants: Can you double check this description?

libs/domain/cart/entry/src/entry.model.ts
```diff
@@ -21,6 +21,11 @@ export interface CartEntryAttributes {
    */
   price?: number;
```
tolerants, line 25: It looks like we are not using it anymore.

libs/domain/cart/entry/src/entry.component.ts
```diff
@@ -60,7 +61,8 @@ export class CartEntryComponent
 
   @signalProperty({ type: Number }) quantity?: number;
   @property() key?: string;
-  @property({ type: Number }) price?: number;
+  @property({ type: Number }) itemPrice?: number;
+  @property({ type: Number }) entryPrice?: number;
```
tolerants: Snouldn't be `entryPrice` a part of entry model?

libs/domain/product/price/src/price.component.ts
```diff
@@ -35,10 +28,44 @@ export class ProductPriceComponent extends ProductMixin(
...
 
-  protected $prices = computed(() => {
-    return this.formatPrices(this.$product()?.price);
+  protected $isDiscounted = computed(() => {
+    return (
+      (this.sales &&
+        typeof this.$salesPrice() === 'number' &&
+        this.sales < Number(this.$salesPrice())) ||
+      (typeof this.$salesPrice() === 'number' &&
+        typeof this.$salesPrice() === 'number' &&
```
tolerants: Why we need to identical conditions?

tolerants (changes requested):

libs/domain/product/price/src/price.component.ts
tobi-or-not-tobi: Thanks, that was wrong. I've actually updated the logic and aligned e2e.

tobi-or-not-tobi: Also renamed the `SitePriceComponent`, as it had a flaw.

tolerants (dismissed):

tolerants (dismissed):

tolerants (dismissed):

PR #612, opened 2023-07-17
refactor: move semantic link types into router config

supproduction:
Moves semantic link types into router config.

closes: [HRZ-3205](https://spryker.atlassian.net/browse/HRZ-3205)

[HRZ-3205]: https://spryker.atlassian.net/browse/HRZ-3205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/cart/entry/src/entry.component.ts
+++ b/libs/domain/cart/entry/src/entry.component.ts
@@ -10,7 +10,6 @@ import {
   NotificationService,
   PricingService,
   SemanticLinkService,
-  SemanticLinkType,
 } from '@spryker-oryx/site';
 import { AlertType } from '@spryker-oryx/ui';
 import { ButtonType } from '@spryker-oryx/ui/button';
@@ -38,6 +37,8 @@ import {
   RemoveByQuantity,
 } from './entry.model';
 import { cartEntryStyles } from './styles';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { RouteLinkType } from '@spryker-oryx/router/lit';
 
 /**
  * Supports updating the quantity as well as removing the entry entirely.
@@ -88,7 +89,7 @@ export class CartEntryComponent
 
   protected $productLink = computed(() => {
     return this.semanticLinkService.get({
-      type: SemanticLinkType.Product,
+      type: RouteLinkType.Product,
       id: this.$product()?.sku,
     });
   });
@@ -105,7 +106,7 @@ export class CartEntryComponent
     if (!this.$options()?.enableItemImage) return;
 
     return html`
-      <a href=${this.$productLink()}>
+      <a href=${ifDefined(this.$productLink())}>
         <oryx-product-media
           .options=${{ containerSize: ProductMediaContainerSize.Thumbnail }}
         ></oryx-product-media>
--- a/libs/domain/cart/entry/src/entry.component.ts
+++ b/libs/domain/cart/entry/src/entry.component.ts
@@ -37,7 +37,6 @@ import {
   RemoveByQuantity,
 } from './entry.model';
 import { cartEntryStyles } from './styles';
-import { ifDefined } from 'lit/directives/if-defined.js';
 import { RouteType } from '@spryker-oryx/router';
 
 /**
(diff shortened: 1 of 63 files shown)


libs/domain/cart/entry/src/entry.component.ts
```diff
@@ -105,7 +106,7 @@ export class CartEntryComponent
     if (!this.$options()?.enableItemImage) return;
 
     return html`
-      <a href=${this.$productLink()}>
+      <a href=${ifDefined(this.$productLink())}>
```
tobi-or-not-tobi:
```suggestion
      <a .href=${this.$productLink()}>
```

This seems to work fine, where the `ifDefined` comes for free. I'm not sure how this works, if there are some properties that will be reflected as attributes.

libs/domain/checkout/src/services/default-checkout.service.ts
```diff
@@ -26,6 +26,7 @@ import {
   PlaceOrderStart,
   PlaceOrderSuccess,
 } from './state';
+import { RouteLinkType } from '@spryker-oryx/router/lit';
```
tobi-or-not-tobi: can we get the type from the router entry point? Is there anything lit specific here?

libs/domain/search/src/models/suggestion/model.ts
```diff
@@ -1,12 +1,12 @@
...
-import { SemanticLinkType } from '@spryker-oryx/site';
 import { DirectiveResult } from 'lit/async-directive';
 import { SuggestionField } from '../../services';
+import { RouteLinkType } from '@spryker-oryx/router/lit';
 
 export interface SuggestionLinks {
   title?: DirectiveResult;
   options: SuggestionResource[];
-  type: SemanticLinkType;
+  type: RouteLinkType | string;
```
tobi-or-not-tobi: should we rename `RouteLinkType` to just `LinkType`?

libs/platform/experience/src/routes.ts
```diff
@@ -15,27 +15,31 @@ export const defaultExperienceRoutes: RouteConfig[] = [
       route="/product/:sku"
     ></oryx-composition>`,
     name: 'Product test',
+    type: RouteLinkType.Product,
```
tobi-or-not-tobi: Can we drop the name on these routes? Where do we use it?

tobi-or-not-tobi: Stan I also discussed to drop the `<oryx-composition>` by default. Have you considered this?

libs/template/labs/src/articles/article-routes.ts
```diff
@@ -1,32 +1,22 @@
...
     ></oryx-composition>`,
     name: 'FAQ',
-  },
-  {
-    path: '/article',
-    render: (): TemplateResult => html`<oryx-composition
-      route="/article"
-    ></oryx-composition>`,
-    name: 'Articles',
+    type: StoryblokContentFields.Faq,
```
tobi-or-not-tobi, line 8: This route config is an example where we'd hope that the render is not needed, as we can have a fallback using the `path`. Optionally, however, one could bring in a custom render method.

tobi-or-not-tobi (changes requested):

libs/domain/cart/entry/src/entry.component.ts
supproduction: I hope we want to keep at as attribute not as property in order of ssr. But the main point of ifDefined it's that attribute will be not rendered if value is null or undefined

libs/template/labs/src/articles/article-routes.ts
supproduction: But without router definition we don't know if we have this route at all

libs/domain/cart/entry/src/entry.component.ts
supproduction: But I can drop ofc

libs/domain/search/src/models/suggestion/model.ts
tobi-or-not-tobi: mmm, is this interface even in use?

tobi-or-not-tobi (approved):

PR #616, opened 2023-07-17
feat(core): add font preloading

supproduction:
Adds font preloading

closes: [HRZ-3237](https://spryker.atlassian.net/browse/HRZ-3237)

[HRZ-3237]: https://spryker.atlassian.net/browse/HRZ-3237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/content/src/services/default-font.service.ts
```diff
@@ -7,13 +7,23 @@ export class DefaultFontService implements FontService {
...
-      });
+        {
+          name: 'link',
+          attrs: {
+            rel: 'stylesheet',
+            href: `https://fonts.googleapis.com/css2?family=${font}:wght@400;500;600;700&display=swap`,
+            media: 'all',
+          },
+        },
+      ]);
```
tobi-or-not-tobi: Is there a reasonable design thinkable to do the heavy lifting of adding preload (and potentially other things) lower level? I'd prefer to have an api that is doing this stuff for you, as this is quite specific knowledge.

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #618, opened 2023-07-18
refactor(ui): merge button and icon-button component

tobi-or-not-tobi:
The button component has gone through some changes. The button now supports an icon type, so that the icon-button component could be dropped. 

closes: [HRZ-3283](https://spryker.atlassian.net/browse/HRZ-3283)

BREAKING CHANGE: 
- The `oryx-icon-button` is removed, use `<oryx-button type="icon">` going forward.
- Previously, it was mandatory to _slot in_ a `<button>` or `<a>` element. While this is still supported, using the `slot="custom"`, buttons will now create the button or link based on properties.
- The `type` property was previously used to indicate both the color and style. This has been separated, so that you can define the `type` using the `ButtonType` enum (`solid`, `outline`, `icon` or `text`) and the color using the `ButtonColor` enum (`primary`, `neutral` or `error`) separately. 
- Previously the `outline` was set as a separate property, this property is removed and moved into the `ButtonType` enum.
- The `size` property is now limited to the values of the `ButtonType`, so that other sizes (`xs` and `xl`) cannot be used on the button component.

[HRZ-3283]: https://spryker.atlassian.net/browse/HRZ-3283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/action/button/button.component.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { fixture, html } from '@open-wc/testing-helpers';
+import { useComponent } from '@spryker-oryx/core/utilities';
+import { ButtonComponent } from './button.component';
+import { buttonComponent } from './button.def';
+
+describe('ButtonComponent', () => {
+  let element: ButtonComponent;
```
Sinucid, line 7: Will the rest functionality be covered by tests?

libs/base/ui/action/button/button.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  @property() text?: string;
+  @property() label?: string;
+  @property() icon?: Icons | string;
+  @property() href?: string;
+
+  @property({ reflect: true }) type?: ButtonType;
+  @property({ reflect: true }) size?: ButtonSize;
+  @property({ type: Boolean, reflect: true }) block?: boolean;
+  @property({ reflect: true })
```
Sinucid: Do we need to reflect prop that does not set attribute in setter?

libs/base/ui/action/button/button.component.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.style.setProperty('--_c8', `var(--oryx-color-${color}-8)`);
+    this.style.setProperty('--_c9', `var(--oryx-color-${color}-9)`);
+    this.style.setProperty('--_c10', `var(--oryx-color-${color}-10)`);
+  }
+
+  @property({ type: Boolean, reflect: true }) loading?: boolean;
+  @property({ type: Boolean, reflect: true }) confirmed?: boolean;
+  @property({ type: Boolean, reflect: true }) disabled?: boolean;
+
+  @property({ type: Boolean }) custom?: boolean;
```
Sinucid: What about render slot anyway and try to use `queryAssignedElements` instead property? The template will look like this:
```

@queryAssignedElements()
protected customContent!: Array<HTMLElement>;

protected render(): TemplateResult {
    return html`
      <slot></slot>
      ${when(link && !this.customContent.length, this.renderLink)}
      ${when(!link && !this.customContent.length, this.renderButton)}
    `;
}
```

libs/base/ui/action/button/button.component.ts
```diff
@@ -0,0 +1,137 @@
...
+            ${this.renderIcon()}${this.text}
+          </slot>
+          ${preHydrate(hydrateSlotChange, this.tagName.toLowerCase())}`
+      );
+      if (this.custom) return html`${template}`;
+    }
+
+    if (this.href) {
+      return html`<a
+        part="link"
```
Sinucid: due to each type of trigger renders separately - the can have the same part name, ie part='button'

libs/base/ui/action/button/button.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  protected onSlotChange(): void {
+    this.setAttributes();
+  }
+
+  /**
+   * There's no point in setting has-text when has-icon is false.
+   */
+  protected setAttributes(): void {
+    const hasIcon = !!(this.icon || this.querySelector?.('oryx-icon'));
```
Sinucid, line 139: Are you sure that this is able to query slotted oryx-icon?

libs/base/ui/action/button/button.component.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.toggleAttribute('has-icon', hasIcon);
+    if (hasIcon) {
+      this.toggleAttribute(
+        'has-text',
+        !!(this.text || this.textContent?.trim())
```
Sinucid, line 144: The same, not sure that it can check textContent inside slot

libs/base/ui/action/button/button.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  @property({ type: Boolean }) custom?: boolean;
+
+  protected willUpdate(props: PropertyValues): void {
+    super.willUpdate(props);
+    this.setAttributes();
+  }
+
+  protected override render(): TemplateResult {
+    const template = [this.renderLoader(), this.renderConfirmed()];
```
Sinucid: What happened if loading and confirmed both true?

libs/base/ui/action/button/button.def.ts
```diff
@@ -0,0 +1,15 @@
+import { componentDef } from '@spryker-oryx/core';
+
+export const buttonComponent = componentDef({
+  name: 'oryx-button',
+  impl: () => import('./button.component').then((m) => m.ButtonComponent),
+  //   stylesheets: [
```
Sinucid: leftover?

libs/base/ui/action/button/button.model.ts
```diff
@@ -0,0 +1,117 @@
...
+  loading?: boolean;
+
+  /**
+   * Indicates whether the button is in the confirmed state.
+   * When `confirmed` is set to `true`, an effect can be rendered on the button temporarily,
+   * typically after a user clicked on the button, for example, when an item is added to the cart.
+   *
+   * @default false
+   */
+  confirmed?: boolean;
```
Sinucid, line 81: disabled is missed

libs/base/ui/action/button/button.model.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * A button with an outlined border.
+   */
+  Outline = 'outline',
+
+  /**
+   * A button which is rendered as icon.
+   */
+  Icon = 'icon',
```
Sinucid, line 120: also we already have designs for icon-button's types

libs/base/ui/action/button/button.styles.ts
```diff
@@ -0,0 +1,289 @@
...
+const baseStyles = css`
+  :host {
+    ${headingUtil(HeadingTag.Bold)};
+
+    color: var(--_text-color);
+    isolation: isolate;
+  }
+
+  :is(a, button),
+  ::slotted(:is(a, button)) {
```
Sinucid, line 10: why not just :slotted(a, button) ?

libs/base/ui/action/button/button.styles.ts
```diff
@@ -0,0 +1,289 @@
...
+    transition: var(--oryx-transition-time);
+    cursor: pointer;
+  }
+
+  :host(:is([type='solid'], [type='outline'])) :is(a, button),
+  :host(:is([type='solid'], [type='outline'])) ::slotted(:is(a, button)) {
+    border: solid 2px var(--_border-color);
+  }
+
+  :host([block]) > *,
```
Sinucid:
```suggestion
  :host([block]) > :is(a, button),
```

libs/base/ui/action/button/button.styles.ts
```diff
@@ -0,0 +1,289 @@
...
+
+  :is(a, button),
+  ::slotted(:is(a, button)),
+  :host([custom]) {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+  }
+
+  /* :host([custom]:not([icon-only]):not([loading]):not([confirmed])), */
```
Sinucid: leftover?

libs/base/ui/action/button/button.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    return html`
+      <button
+        part="button"
+        type="button"
+        ?disabled=${this.disabled}
+        ?empty=${!this.text}
+        .ariaLabel=${this.label}
+      >
+        ${template}
```
Sinucid: what happened if type="icon" and it has text?

libs/base/ui/action/button/stories/demo.stories.ts
```diff
@@ -0,0 +1,128 @@
...
+  ButtonComponentAttributes,
+  ButtonType,
+} from '../button.model';
+
+export default {
+  title: `${storybookPrefix}/Actions/Action`,
+  args: {
+    text: 'Action',
+    color: ButtonColor.Primary,
+    // type: ButtonType.Outline,
```
Sinucid: ?

libs/base/ui/action/button/stories/demo.stories.ts
```diff
@@ -0,0 +1,128 @@
...
+      options: ['link', 'button', 'content'],
+      control: { type: 'select' },
+      table: { category: 'demo' },
+    },
+    href: {
+      control: { type: 'text' },
+      table: { category: 'demo' },
+    },
+
+    //   color: {
```
Sinucid: .

libs/base/ui/action/button/stories/demo.stories.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+const Template: Story<ButtonComponentAttributes & Props> = (
+  props
+): TemplateResult => {
+  let template;
+
+  if (props.slotted === 'content') {
+    template = html`
+      ${when(
```
Sinucid: lots of duplications of icon and text for each template type, can we extract them to the const?

libs/base/ui/action/button/stories/static.stories.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const WarningTemplate: Story = (): TemplateResult => {
+  return staticButtons(ButtonColor.Warning);
+};
+
+const ErrorTemplate: Story = (): TemplateResult => {
+  return staticButtons(ButtonColor.Error);
+};
+
+export const Primary = PrimaryTemplate.bind({});
```
Sinucid: Can we not to create a bunch of separate static stories and combine variants in single one?

libs/base/ui/action/button/stories/util.ts
```diff
@@ -0,0 +1,192 @@
...
+  icon?: string;
+  color?: string;
+  disabled?: boolean;
+  loading?: boolean;
+  confirmed?: boolean;
+  slotted?: 'content' | 'link' | 'button';
+  href?: string;
+}
+
+const renderButton = (props: ButtonProps) => {
```
Sinucid: can we reuse this for the demo story?

libs/domain/cart/add/src/add.component.ts
```diff
@@ -54,21 +52,20 @@ export class CartAddComponent extends ProductMixin(
...
-        @mouseup=${this.onMouseUp}
-        @click=${this.onSubmit}
-      >
-        <oryx-icon .type=${IconTypes.CartAdd} size=${Size.Lg}></oryx-icon>
-        ${when(enableLabel, () => html`${this.i18n('cart.add-to-cart')}`)}
-      </button>
-    </oryx-button>`;
+      .size=${Size.Md}
+      .type=${type}
+      .text=${enableLabel ? text : undefined}
```
Sinucid: ifDefined?

libs/domain/cart/add/src/add.component.ts
```diff
@@ -113,6 +110,7 @@ export class CartAddComponent extends ProductMixin(
   }
 
   protected onSubmit(e: Event | CustomEvent<QuantityEventDetail>): void {
+    console.log('e', e);
```
Sinucid: console.log

libs/domain/order/entries/entries.component.ts
```diff
@@ -73,14 +74,15 @@ export class OrderEntriesComponent
...
-      <button @click=${this.toggle}>
-        ${this.expanded ? '-' : '+'}${restItemsCount}
-        ${this.i18n(
-          this.expanded ? 'order.less-products' : 'order.more-products'
-        )}
-      </button>
-    </oryx-button>`;
+    let text = this.expanded ? '-' : '+';
+    text += restItemsCount;
+    text += this.expanded ? 'order.less-products' : 'order.more-products';
```
Sinucid: you missed i18n here. Probably will be better to use parametrised token here:
```
i18n(
  this.expanded ? 'order.<value>-less-products' : '<value>-order.more-products',
  {value: `${this.expanded ? '-' : '+'} ${restItemsCount}`}
)
```

libs/domain/site/navigation-item/navigation-item.component.ts
```diff
@@ -116,16 +116,12 @@ export class SiteNavigationItemComponent extends ContentMixin<SiteNavigationItem
     return html`
```
Sinucid, line 89: there is still oryx-icon-button trigger type inside

libs/domain/user/address-add-button/address-add-button.component.ts
```diff
@@ -24,26 +25,19 @@ export class UserAddressAddButtonComponent extends AddressMixin(
...
-    }
+    const href =
+      this.$options().target === Target.Link ? this.$createLink() : undefined;
+
+    return html`
+      <oryx-button
+        .type=${ButtonType.Outline}
+        .icon=${IconTypes.Add}
+        .text=${this.i18n(['add', 'user.address.add'])}
+        .href=${href}
```
Sinucid, line 39: ifDefined?

libs/domain/user/address-edit-button/address-edit-button.component.ts
```diff
@@ -26,25 +27,17 @@ export class UserAddressEditButtonComponent extends AddressMixin(
...
-      `;
-    }
+    const href =
+      this.$options().target === Target.Link ? this.editLink() : undefined;
+    return html`
+      <oryx-button
+        .type=${ButtonType.Outline}
+        .icon=${IconTypes.Edit}
+        .text=${this.i18n(['edit', 'user.address.edit-address'])}
+        .href=${href}
```
Sinucid, line 37: ifDefined?

Sinucid (changes requested):

libs/base/ui/action/button/button.component.spec.ts
tobi-or-not-tobi: yup, it's only draft for now

libs/base/ui/action/button/button.component.ts
tobi-or-not-tobi: no, that's wrong.

libs/base/ui/action/button/button.component.ts
tobi-or-not-tobi: One of the reasons to try a diff approach with a slot-less component is that i realised that using web components is quit heavy on the number of DOM nodes. I've not seen direct impact on performance in the client, but Lighthouse/core web vitals do complain about this. And honestly, when I wrote performance reviews for frontend in the past I also put this as a concern.

But it doesn't make the implentation cleaner indeed.

libs/base/ui/action/button/button.component.ts
tobi-or-not-tobi: Yeah, but it's a bit confusing as it's not a button. I can also imagine we drop this, and leave it to the slotted button/link in case fine-grained css is needed.

libs/domain/user/address-edit-button/address-edit-button.component.ts
tobi-or-not-tobi: Why? The dot-syntax does this already.

libs/domain/user/address-add-button/address-add-button.component.ts
tobi-or-not-tobi: Why? The dot-syntax does this already.

libs/domain/site/navigation-item/navigation-item.component.ts
tobi-or-not-tobi: I have not integrated `oryx-button` for all `oryx-icon-button` yet, and therefor kept `oryx-icon-button` in the code base for now.

libs/domain/order/entries/entries.component.ts
tobi-or-not-tobi: Decided to clean up the code a bit. We do not like the tokens to be too opinionated. We can still add a localisation in the labs package if we want to.

libs/domain/cart/add/src/add.component.ts
tobi-or-not-tobi: not sure how that would help

libs/base/ui/action/button/stories/util.ts
tobi-or-not-tobi: Nice one

libs/base/ui/action/button/stories/static.stories.ts
tobi-or-not-tobi: dropped because of reuse of `renderButton`

libs/base/ui/action/button/stories/demo.stories.ts
tobi-or-not-tobi: dropped because of reuse of `renderButton`

libs/base/ui/action/button/button.component.ts
tobi-or-not-tobi: This is actually a feature, see https://www.figma.com/file/KsCTJtf3xenIL91rGs8FCa/FES-Storefront-Components?type=design&node-id=2665%3A12504&mode=design&t=PEuSzw9roaO59Waq-1

libs/base/ui/action/button/button.styles.ts
tobi-or-not-tobi: because that won't work

libs/base/ui/action/button/button.model.ts
tobi-or-not-tobi: Not sure what you mean here. We have followed designs, that have both version of icon-only buttons, as well as icons combined in solid, outline and text variants.

libs/base/ui/action/button/button.component.ts
tobi-or-not-tobi: Not a supported case. To ensure that cannot happen, we could introduce a `state` property that can be an enum value of `disabled`, `loading`, or `confirmed`.

tobi-or-not-tobi: But I like the separate boolean properties more, as they're easier to manage and use in our component/styles.

libs/base/ui/action/button/button.component.ts
tobi-or-not-tobi: yup. the querySelector goes against the light dom, where as if we we query against the shadowRoot, it would query inside the shadow dom.

libs/base/ui/action/button/button.component.ts
tobi-or-not-tobi: same

libs/base/ui/action/button/button.component.ts
Sinucid: I don't see any reason not to keep the slot for custom content constantly. Without content it will be empty

libs/base/ui/action/button/button.component.ts
Sinucid: I believe there should be some priorities like: disabled > confirmed > loading

libs/base/ui/action/button/button.styles.ts
Sinucid: yeah, my bad

libs/base/ui/action/button/button.component.ts
Sinucid: But is this not yet implemented?

libs/domain/cart/add/src/add.component.ts
Sinucid: undefined should be a valid value for property, and I'm wondering whether it triggers additional render when set such values or not? Do not set the property at all is a more stable way, from my perspective, but I can be wrong

libs/domain/order/entries/entries.component.ts
```diff
@@ -71,15 +72,18 @@ export class OrderEntriesComponent
...
 
-    return html`<oryx-button type="text">
-      <button @click=${this.toggle}>
-        ${this.expanded ? '-' : '+'}${restItemsCount}
-        ${this.i18n(
-          this.expanded ? 'order.less-products' : 'order.more-products'
-        )}
-      </button>
+    const i18nToken = this.expanded
+      ? 'order.less-products-<count>'
```
Sinucid: count vs value, choose one of them

Sinucid (changes requested):

libs/base/ui/action/button/button.component.ts
tobi-or-not-tobi: fair. i've done in component logic and small css change .

libs/base/ui/action/button/button.component.ts
tobi-or-not-tobi: done a new attempt.

libs/base/ui/action/button/button.component.spec.ts
Sinucid: Let it be like a reminder

libs/base/ui/action/button/button.component.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  protected willUpdate(props: PropertyValues): void {
+    super.willUpdate(props);
+    this.setAttributes();
+  }
+
+  protected override render(): TemplateResult {
+    const templates = [this.renderLoader(), this.renderConfirmed()];
+
+    const x = this.href ? this.renderLink() : this.renderButton();
```
Sinucid: `const button` ?

libs/base/ui/action/button/button.def.ts
```diff
@@ -0,0 +1,25 @@
...
+export const buttonComponent = componentDef({
+  name: 'oryx-button',
+  impl: () => import('./button.component').then((m) => m.ButtonComponent),
+  stylesheets: [
+    {
+      rules: [
+        {
+          media: { screen: Size.Sm },
+          css: css`
+            /* :host {
```
Sinucid: leftover?

libs/base/ui/action/button/prehydrate.ts
```diff
@@ -0,0 +1,13 @@
+import { LitElement } from 'lit';
+
+export const hydrateSlotChange = async (host: LitElement): Promise<void> => {
+  const slot = host.shadowRoot?.querySelector('slot') as HTMLElement;
```
Sinucid:
```suggestion
  const slot = host.shadowRoot?.querySelector<HTMLSlotElement>('slot');
```

libs/base/ui/overlays/notification/src/notification.component.spec.ts
```diff
@@ -44,10 +43,10 @@ describe('NotificationComponent', () => {
         ></oryx-notification>
       `);
     });
-    it('should have correct translated aria-labels', async () => {
-      const closeBtn = getButton('button');
 
-      expect(closeBtn?.getAttribute('aria-label')).toBe(closeButtonAriaLabel);
+    it('should have correct translated aria-labels', async () => {
```
Sinucid: async without await

libs/base/ui/structure/collapsible-text/collapsible-text.component.spec.ts
```diff
@@ -100,21 +101,27 @@ describe('CollapsibleTextComponent', () => {
...
-        beforeEach(() => {
-          element.renderRoot.querySelector('button')?.click();
-        });
-
-        it('should have Read less aria label', () => {
-          const button = element.renderRoot.querySelector('button');
-          expect(button?.textContent).toContain('Read less');
-        });
-      });
+      // TODO: fix this test
```
Sinucid: as reminder

libs/base/ui/overlays/dropdown/src/dropdown.component.spec.ts
```diff
@@ -53,39 +52,38 @@ describe('DropdownComponent', () => {
     });
   });
 
-  describe('when "oryx.close" event dispatched', () => {
-    let input: HTMLInputElement | null | undefined;
+  // describe('when "oryx.close" event dispatched', () => {
```
Sinucid, line 55: need to fix (reminder)

libs/domain/picking/picking-header/picking-header.component.spec.ts
```diff
@@ -80,66 +79,72 @@ describe('PickingHeaderComponent', () => {
...
-    expect(getDiscardModal()).not.toBeNull();
-  });
+    it('should render discard modal', () => {
+      expect(element).toContainElement('oryx-discard-picking');
+    });
 
-  it('should not show discard modal', () => {
-    expect(getDiscardModal()?.hasAttribute('open')).toBe(false);
+    it('should not show discard modal', () => {
+      expect(
```
Sinucid: `expect(element).toContainElement('oryx-discard-picking:not([open])');`

libs/domain/picking/picking-header/picking-header.component.spec.ts
```diff
@@ -155,17 +160,23 @@ describe('PickingHeaderComponent', () => {
...
+      element.renderRoot
+        .querySelectorAll<HTMLElement>('oryx-button')
+        .forEach((el) =>
+          expect(el).not.toHaveProperty('text', 'Customer note')
+        );
     });
 
     it('should not render customer note modal', () => {
-      expect(getCustomerNoteModal()).toBeNull();
+      // expect(getCustomerNoteModal()).toBeNull();
```
Sinucid: need to fix

libs/domain/picking/picking-header/picking-header.component.spec.ts
```diff
@@ -177,17 +188,30 @@ describe('PickingHeaderComponent', () => {
     beforeEach(async () => {
       showDialogTrigger.next(true);
     });
+
     it('should open discard modal', () => {
-      expect(getDiscardModal()?.hasAttribute('open')).toBe(true);
```
Sinucid, line 181: `expect(element).toContainElement('oryx-discard-picking[open]');`

libs/domain/picking/picking-header/picking-header.component.spec.ts
```diff
@@ -177,17 +188,30 @@ describe('PickingHeaderComponent', () => {
...
-        getDiscardModal()?.renderRoot.querySelector('button')?.click();
+        element.renderRoot
+          .querySelector('oryx-discard-picking')
+          ?.dispatchEvent(
+            new CustomEvent(CLOSE_EVENT, { bubbles: true, composed: true })
+          );
       });
 
       it('should close discard modal', () => {
-        expect(getDiscardModal()?.hasAttribute('open')).toBe(false);
```
Sinucid, line 190: `expect(element).toContainElement('oryx-discard-picking:not([open])');`

libs/domain/picking/picking-list-item/picking-list-item.component.spec.ts
```diff
@@ -111,20 +111,15 @@ describe('PickingListItemComponent', () => {
...
+      element.renderRoot.querySelector<HTMLElement>('oryx-button')?.click();
       expect(event).toHaveBeenCalled();
     });
 
     describe('and picking is proceed', () => {
       beforeEach(() => {
         element.renderRoot
-          .querySelector('oryx-button button')
-          ?.dispatchEvent(new MouseEvent('click'));
+          .querySelectorAll<HTMLElement>('oryx-button')[1]
```
Sinucid: can we use more specific selector here?

libs/domain/picking/picking-list-item/picking-list-item.styles.ts
```diff
@@ -39,11 +39,11 @@ export const pickingListItemComponentStyles = css`
     width: 100%;
   }
 
-  oryx-icon-button {
+  oryx-button.show-customer {
```
Sinucid:
```suggestion
  .show-customer {
```

libs/domain/picking/picking-list-item/picking-list-item.styles.ts
```diff
@@ -39,11 +39,11 @@ export const pickingListItemComponentStyles = css`
...
     margin-inline-start: auto;
   }
 
-  oryx-button {
+  oryx-button.start-picking {
```
Sinucid:
```suggestion
  .start-picking {
```

libs/domain/picking/picking-product-card/picking-product-card.component.spec.ts
```diff
@@ -90,7 +90,11 @@ describe('PickingProductCardComponent', () => {
         });
 
         it('should disable the button', () => {
-          expect(element).toContainElement('oryx-button button[disabled]');
+          expect(
```
Sinucid: `expect(element).toContainElement('oryx-button[disabled]');`

libs/domain/picking/user-profile/user-profile.component.spec.ts
```diff
@@ -204,15 +209,19 @@ describe('UserProfileComponent', () => {
 
     describe('and the receive data button is clicked', () => {
       beforeEach(() => {
-        getReceiveDataButton().click();
+        element.renderRoot
+          .querySelectorAll<HTMLElement>('oryx-button')[1]
```
Sinucid: more specific selector?

libs/domain/picking/user-profile/user-profile.component.spec.ts
```diff
@@ -204,15 +209,19 @@ describe('UserProfileComponent', () => {
...
+
       it('should call offline data plugin', () => {
         expect(mockOfflineDataPlugin.refreshData).toHaveBeenCalled();
       });
+
       it('should render loading indicator', async () => {
-        expect(
-          element.renderRoot.querySelector('oryx-button[loading]')
-        ).not.toBeNull();
+        const button = element.renderRoot.querySelectorAll('oryx-button')[1];
```
Sinucid: same

libs/domain/picking/user-profile/user-profile.component.spec.ts
```diff
@@ -180,11 +180,16 @@ describe('UserProfileComponent', () => {
...
-          element.renderRoot.querySelector('button')?.click();
+
+          element.renderRoot
+            .querySelectorAll<HTMLElement>('oryx-button')[0]
+            ?.click();
         });
 
         it('should not show loading indicator', () => {
-          expect(element).not.toContainElement('oryx-button[loading]');
+          const button = element.renderRoot.querySelectorAll('oryx-button')[0];
```
Sinucid: more specific selector?

libs/domain/picking/user-profile/user-profile.component.spec.ts
```diff
@@ -222,15 +231,20 @@ describe('UserProfileComponent', () => {
           element = await fixture(
             `<oryx-picking-user-profile></oryx-picking-user-profile>`
           );
-          getReceiveDataButton().click();
+          element.renderRoot
+            .querySelectorAll<HTMLElement>('oryx-button')[1]
```
Sinucid: this and the rest

libs/domain/picking/warehouse-assignment/warehouse-assignment.component.spec.ts
```diff
@@ -97,10 +98,11 @@ describe('WarehouseAssignmentComponent', () => {
   });
 
   it('should render button of each location', () => {
-    const buttons = el.renderRoot.querySelectorAll('button');
+    const buttons =
+      el.renderRoot.querySelectorAll<ButtonComponent>('oryx-button');
     expect(buttons.length).toBe(mockWarehouseUserAssignments.length);
     mockWarehouseUserAssignments.forEach((item, index) => {
-      expect(buttons[index].textContent).toContain(i18n('picking.select'));
+      expect(buttons[index]).toHaveProperty('text', i18n('picking.select'));
```
Sinucid, line 105: If some of the buttons is failed the check will be hard to understand which one exact. I believe each case should be wrapped on its own it('...')

libs/domain/user/address-list-item/address-list-item.component.ts
```diff
@@ -102,23 +103,23 @@ export class UserAddressListItemComponent extends AddressMixin(
...
-        <oryx-icon-button outline>
-          <a href=${this.editLink()}>
-            <oryx-icon .type=${IconTypes.Edit}></oryx-icon>
-          </a>
-        </oryx-icon-button>
+        <oryx-button
+          .type=${ButtonType.Icon}
+          .size=${ButtonSize.Md}
+          .icon=${IconTypes.Edit}
+          .href=${this.editLink()}
```
Sinucid, line 111: Does link not contain any text? It is wrong at least from accessibility perspective

libs/platform/experience/src/color/colors/colored/red.ts
```diff
@@ -2,6 +2,7 @@ import { Color } from '../../color.model';
 
 export const red: Color = {
   light: {
+    0: 'white',
```
Sinucid, line 5: Is light red color white?

libs/template/themes/design-tokens/src/backoffice/button.token.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const buttonTokens: DesignToken[] = [
+  { 'button-border-radius': '10px' },
+  {
+    media: { screen: Size.Sm },
+    'button-size-factor': '1',
+  },
+];
+
+// 'solid-button-icon-size': '16px',
```
Sinucid: leftovers?

libs/template/themes/design-tokens/src/backoffice/icon.token.ts
```diff
@@ -1,9 +1,9 @@
 import { ThemeToken } from '@spryker-oryx/experience';
 
 export const iconTokens: ThemeToken = {
-  icon: {
-    size: {
-      default: 'var(--oryx-icon-size-lg)',
-    },
-  },
+  // icon: {
```
Sinucid, line 4: ?

libs/template/themes/design-tokens/src/storefront/button.token.ts
```diff
@@ -0,0 +1,14 @@
+import { DesignToken } from '@spryker-oryx/experience';
+
+export const buttonTokens: DesignToken[] = [
+  {
+    button: {
+      // 'font-weight': '700',
```
Sinucid: ?

Sinucid (changes requested):

libs/platform/experience/src/color/colors/colored/red.ts
tobi-or-not-tobi: position 0 is white for all non-bright colors.

Sinucid (dismissed):

libs/domain/picking/user-profile/user-profile.component.spec.ts
```diff
@@ -151,24 +143,28 @@ describe('UserProfileComponent', () => {
...
+      const button = element.renderRoot.querySelectorAll('oryx-button')[0];
+      expect(button).toHaveProperty('text', 'Log out');
+      expect(button?.hasAttribute('disabled')).toBe(true);
     });
   });
 
   describe('when sync is not pending and picking is not in progress', () => {
     it('should not disable log out button', () => {
-      expect(getDisabledButton()).toBeNull();
+      const button = element.renderRoot.querySelectorAll('oryx-button')[0];
```
Sinucid, line 154: you've added specific classes for the buttons, use the as selectors here

libs/domain/picking/user-profile/user-profile.component.spec.ts
```diff
@@ -151,24 +143,28 @@ describe('UserProfileComponent', () => {
...
 
     describe('and when log out button is clicked', () => {
       beforeEach(() => {
-        element.renderRoot.querySelector('button')?.click();
+        element.renderRoot.querySelector<HTMLElement>('oryx-button')?.click();
       });
 
       it('should show loading indicator', () => {
-        expect(element).toContainElement('oryx-button[loading]');
+        const button = element.renderRoot.querySelectorAll('oryx-button')[0];
```
Sinucid, line 165: and here

libs/domain/picking/user-profile/user-profile.component.spec.ts
```diff
@@ -151,24 +143,28 @@ describe('UserProfileComponent', () => {
     });
 
     it('should disable log out button', () => {
-      const button = getDisabledButton();
-      expect(button).not.toBeNull();
-      expect(button?.textContent).toContain('Log Out');
+      const button = element.renderRoot.querySelectorAll('oryx-button')[0];
```
Sinucid, line 146: and here

Sinucid (changes requested):

Sinucid (approved):

PR #621, opened 2023-07-18
chore: fix layer hierarchy

supproduction:
Moves base/form package to the platform domain.
Moves components plugin/useComponent/LazyLoad resolver from core into utilities.
Moves Type interface from core into utilities.
Fixes unit tests missing prototypes errors.

closes: [HRZ-2132](https://spryker.atlassian.net/browse/HRZ-2132)

[HRZ-2132]: https://spryker.atlassian.net/browse/HRZ-2132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.eslintrc.json
```diff
@@ -19,27 +19,20 @@
                   "layer:platform",
                   "layer:base",
                   "layer:domain",
-                  "layer:template",
-                  "lib:experience",
-                  "lib:i18n",
-                  "lib:ui",
-                  "lib:form"
+                  "layer:template"
```
tolerants: But it's a violation, `domain` shouldn't depend on `template` layer. So basically everything is allowed for this layer.

.eslintrc.json
```diff
@@ -19,27 +19,20 @@
...
               {
                 "sourceTag": "layer:platform",
                 "onlyDependOnLibsWithTags": [
                   "layer:platform",
                   "layer:base",
-                  "lib:experience",
-                  "lib:i18n",
-                  "lib:ui",
-                  "lib:form"
+                  "layer:domain"
```
tolerants: The same.

libs/base/di/src/models/type.ts
```diff
@@ -1,6 +1,5 @@
+// TODO: remove when fix circular dependency utilities <=> di
```
tolerants, line 1: Let's create a ticket maybe to track it?

tolerants (changes requested):

.eslintrc.json
supproduction: we use import from core into domains `feature` definitions. So we can't omit platform, and template import we use for picking import. We can create picking exception but we can't omit platform until we move out orchestrator?

.eslintrc.json
supproduction: But all good here 
```
{
                "sourceTag": "layer:platform",
                "onlyDependOnLibsWithTags": [
                  "layer:platform",
                  "layer:base",
                  "layer:domain"
                ]
              },
```

.eslintrc.json
```diff
@@ -51,37 +39,12 @@
               },
               // Exception rules
               {
-                "sourceTag": "lib:experience",
+                "sourceTag": "lib:picking",
```
tolerants, line 42: As I understand the only issue is in BAPI feature from the `application` package. But I'm thinking if we can move them to the `auth` package.

tolerants (changes requested):

tolerants (approved):

PR #623, opened 2023-07-19
feat(cart): make "empty message" on the cart page editable

tobi-or-not-tobi:
We have moved the cart empty message into a content manageable component (`<oryx-content-text>`) that is hidden with general visibility rules (`CART.!EMPTY`). Because of this change we've also fixed a but that showed an empty cart temporarily. 

We did not hide the cart entries component with visibility rules as we probably add skeleton design for cart entries going forward. 

closes: [HRZ-3243](https://spryker.atlassian.net/browse/HRZ-3243)
fixes: [HRZ-3168](https://spryker.atlassian.net/browse/HRZ-3168)

[HRZ-3243]: https://spryker.atlassian.net/browse/HRZ-3243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-3168]: https://spryker.atlassian.net/browse/HRZ-3168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/cart/src/services/default-cart.service.ts
```diff
@@ -24,6 +24,7 @@ import {
   startWith,
   switchMap,
   take,
+  tap,
```
supproduction: Do we need it?

supproduction (changes requested):

supproduction (dismissed):

libs/template/presets/storefront/experience/pages/cart-page.ts
```diff
@@ -10,15 +10,39 @@ export const cartPage: StaticComponent = {
...
+        data: {
+          text: `
+          <oryx-icon type="shopping_cart" style="--oryx-icon-size: 40px;"></oryx-icon>
+          <p>Your shopping cart is empty</p><oryx-button>
+          <a href="/search">Shop now</a></oryx-button>`,
+        },
+      },
+      options: {
+        data: {
+          visibility: { hideByRule: 'CART.!EMPTY' },
```
Sinucid: Should then cart-entries be hidden when `CART.EMPTY` ?

tobi-or-not-tobi: no, we have that logic inside. I'm thinking to keep it there, as we likely move into skeleton design for the cart page. We'd ideally do the skeleton with existing (empty) ui components, just iterating over an empty ("start with") list of entries.

supproduction (dismissed):

libs/domain/cart/entry/src/entry.component.ts
```diff
@@ -6,23 +6,24 @@ import {
   ProductMediaContainerSize,
   ProductMixin,
 } from '@spryker-oryx/product';
+import { RouteType } from '@spryker-oryx/router';
```
supproduction, line 9: Do we need it?

supproduction (changes requested):

libs/domain/cart/entry/src/entry.component.ts
tobi-or-not-tobi: yeah, we use it in:
```ts
protected $productLink = computed(() => {
    return this.semanticLinkService.get({
      type: RouteType.Product,
      id: this.$product()?.sku,
    });
  });
```

not sure why the PR changes do not show it.

supproduction (approved):

PR #629, opened 2023-07-20
fix: product title ellipses

tobi-or-not-tobi:
We've refactored the heading and underlying text component. 
- html text (e.g. loaded from CMS) is transformed to Oryx Design System components. For now, only headings, links and buttons are converted. Consequently, we have cleaned up the text component styles since they no longer need to bring Oryx styles. 
- Heading component styles have been optimised to not bloat the CSS with all heading styles. Instead, internal CSS variables are added dynamically (e.g. `--_s: var(--oryx-typography-h1-size)`)

fixes: [HRZ-3300](https://spryker.atlassian.net/browse/HRZ-3300) [HRZ-3299](https://spryker.atlassian.net/browse/HRZ-3299)

[HRZ-3300]: https://spryker.atlassian.net/browse/HRZ-3300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-3299]: https://spryker.atlassian.net/browse/HRZ-3299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/presets/storefront/experience/footer.ts
```diff
@@ -1,17 +1,29 @@
...
+  content: {
+    data: {
+      text: `${text(token)}`,
+    },
+  },
   options: { rules: [rule] },
 });
 
+const text = (token: string): string => {
+  const text = i18n(token);
```
dunqan: I don't think this will work besides auto converting tokens to string repersentation. It's not reactive either, so only first/default language will land in sample data. Thus not much point for wiring i18n here IMO.

libs/domain/product/title/src/title.component.ts
```diff
@@ -33,28 +31,64 @@ export class ProductTitleComponent extends ProductMixin(
...
-      .as=${as}
-      .asLg=${asLg}
-      .asMd=${asMd}
-      .asSm=${asSm}
-    >
-      ${this.$link() ? this.renderLink() : html`${this.$product()?.name}`}
-    </oryx-heading>`;
+    const name = this.$product()?.name;
+    if (!name) return;
+    if (featureVersion <= '1.4') {
```
dunqan: Shou;dn't it be `< '1.4'`? 
Also we decided to use approach with `>= '1.4'` instead.

libs/template/application/oryx-app/oryx-app.styles.ts
```diff
@@ -11,7 +12,9 @@ export const styles = css`
     background-color: var(--oryx-color-neutral-1);
     min-height: 100vh;
     box-sizing: border-box;
-    text-wrap: balance;
+    ${featureVersion <= '1.4'
```
dunqan: as before

libs/base/ui/structure/text/text.component.ts
```diff
@@ -11,6 +11,76 @@ export class TextComponent extends LitElement implements TextProperties {
...
+
+  /**
+   * Modifies an HTML string by adding custom inline styles to specific HTML elements
+   * based on their tags, classes, and screen sizes. The styles align roughly with the
+   * Oryx design system without reusing the design system components directly.
+   *
+   * @param htmlString - The HTML string to be modified.
+   * @returns The modified HTML string with custom inline styles.
+   */
+  protected appendStyleToHeaders(htmlString: string): string {
```
dunqan: I'm just wondering about efficiency of such a solution. 

The component is basically generating custom (non lit-template) html on it's own, now with additional preprocessing. If it would be only for the (fairly static) content, then all good. But we intend to use it as a low level component for components like product title. In a consequence, we are generating html in a custom way for dynamic data also there, and this approach is kind of encouraged 🤔

dunqan (changes requested):

libs/base/ui/structure/text/text.component.ts
```diff
@@ -11,6 +11,76 @@ export class TextComponent extends LitElement implements TextProperties {
...
+    const headerRegex = /<(\w+)([^>]*)>(.*?)<\/\1>|<(\w+)([^>]*)\/>/gis;
+
+    const modifiedHtmlString = htmlString.replace(
+      headerRegex,
+      (match, tag, attributes, content) => {
+        if (!tag) return match;
+        const classes = attributes?.match(/class="(.*?)"/i)?.[1].trim();
```
Sinucid: I would extract all these regexps to the separate consts and put the at the top. At least this huge amount of processing become a bit more readable

Sinucid: And should all these regexps be case-insenitive?

libs/base/ui/structure/text/text.def.ts
```diff
@@ -1,6 +1,32 @@
-import { componentDef } from '@spryker-oryx/utilities';
+import { componentDef, screenCss } from '@spryker-oryx/utilities';
+import { css } from 'lit';
+
+const textScreenStyles = screenCss({
```
Sinucid: Should these be in def file?

libs/base/ui/structure/text/text.def.ts
```diff
@@ -1,6 +1,32 @@
...
+  lg: css`
+    *[class*='lg-'] {
```
Sinucid:
```suggestion
    [class*='lg-'] {
```

libs/base/ui/structure/text/text.def.ts
```diff
@@ -1,6 +1,32 @@
...
+      font-size: var(--_fs-lg, var(--_fs));
+      font-weight: var(--_fw-lg, var(--_fw));
+      line-height: var(--_lh-lg, var(--_lh));
+    }
+  `,
+  md: css`
+    *[class*='md-'] {
```
Sinucid:
```suggestion
    [class*='md-'] {
```

libs/base/ui/structure/text/text.def.ts
```diff
@@ -1,6 +1,32 @@
...
+      font-size: var(--_fs-md, var(--_fs));
+      font-weight: var(--_fw-md, var(--_fw));
+      line-height: var(--_lh-md, var(--_lh));
+    }
+  `,
+  sm: css`
+    *[class*='sm-'] {
```
Sinucid:
```suggestion
    [class*='sm-'] {
```

libs/base/ui/structure/text/text.component.ts
Sinucid: Maybe better to extract this logic to the controller? It will be much more flexible, If we need this logic in some another component, just add the controller instead of `<oryx-text>` inside render, in `oryx-content-text` for example

Sinucid (changes requested):

libs/base/ui/structure/text/text.component.ts
tobi-or-not-tobi: I'd say yes, they need to be case insensitive, as we do not know if third party CMSes or other (manua) content providers are strict on case.

tobi-or-not-tobi: Good suggestion with externalizing the regex.

libs/base/ui/structure/text/text.def.ts
tobi-or-not-tobi: It can. but I've moved it to the styles file. It's indeed better, as we do not load those styles while bootstrapping the app.

libs/base/ui/structure/heading/src/heading.util.ts
```diff
@@ -0,0 +1,72 @@
...
+      tagName = 'span';
+      break;
+    case HeadingTypography.Bold:
+      tagName = 'b';
+      break;
+    default:
+      tagName = tag ?? 'span';
+  }
+
+  return `<${tagName} style="${style}" ${
```
dunqan: we are building string here, for the whole TAG, and passing it into `unsafeStatic` level above, which makes it less efficient and more memory efficient, especially if header will change often at runtime.

https://lit.dev/docs/templates/expressions/#static-expressions
> Changing the value of static expressions is expensive. Expressions using literal values should not change frequently, as they cause a new template to be re-parsed and each variation is held in memory.

maybe we should consider use unsafeStatic only for the parts we really need, and return template instead? Then change of content will not cause full heading structure regeneration?
(but then directive is less usefull for rich text component)

libs/template/presets/storefront/experience/footer.ts
tobi-or-not-tobi: agreed. Also we want those footer navigations to be loaded from CMS anyway.

libs/base/ui/structure/text/text.component.ts
tobi-or-not-tobi: refactored heavily so closing this.

libs/base/ui/structure/text/text.def.ts
tobi-or-not-tobi: This has also dropped from 1.4 onwards with the latest refactor of this PR.

libs/base/ui/structure/text/text.component.ts
tobi-or-not-tobi: We've tried a lot of variations. We ended up with a reusable text mixin, but it's only concerned with converting html to Oryx design system and no longer comes with opinions on the styling of the text, since this is part of the design system.

libs/base/ui/structure/heading/src/heading.util.ts
tobi-or-not-tobi: Refactored.

libs/base/ui/overlays/modal/src/modal.component.ts
```diff
@@ -191,6 +188,17 @@ export class ModalComponent
     `;
   }
 
+  // temporary implementation for backwards compatibility
+  private __renderHeading(): TemplateResult {
```
Sinucid, line 192: Are we starting using **private** descriptor?

libs/base/ui/structure/card/src/card.component.ts
```diff
@@ -12,13 +14,22 @@ export class CardComponent extends LitElement {
 
   protected override render(): TemplateResult {
     return html`
-      <slot name="heading">
-        <oryx-heading>
-          <h5>${this.heading}</h5>
-        </oryx-heading>
+      <slot name="heading"
+        >${this.__renderHeading()} ${this.__renderHeading()}
```
Sinucid: Double heading?

libs/base/ui/structure/card/src/stories/card-demo.stories.ts
```diff
@@ -33,9 +33,10 @@ const Template: Story<CardProperties> = (
       )}
       ${when(
         props.heading,
-        () => html` <oryx-heading slot="heading">
-          <h5>${props.heading}</h5>
-        </oryx-heading>`
+        () =>
```
Sinucid, line 36: Do we still need this? We can just pass **heading** attribute to the component

libs/base/ui/structure/heading/src/typography.controller.ts
```diff
@@ -20,33 +24,57 @@ export class TypographyController implements ReactiveController {
...
+    name: string,
+    size?: Size,
+    value?: string | number
+  ): void {
+    const screen = size ? `-${size}` : '';
+
+    if (value !== undefined) {
+      this.host.style.setProperty(`${name}${screen}`, String(value));
+    } else {
+      this.host.style.removeProperty?.(`${name}${screen}`);
```
Sinucid:
```suggestion
    const propName = `${name}${size ? `-${size}` : ''}`;

    if (value !== undefined) {
      this.host.style.setProperty(propName, String(value));
    } else {
      this.host.style.removeProperty(propName);
```

libs/base/ui/structure/heading/src/heading.component.ts
```diff
@@ -1,47 +1,55 @@
...
 
-  protected override render(): TemplateResult {
-    return html`${this.renderTag(html`<slot></slot>`)}`;
+  protected willUpdate(
+    properties: PropertyValueMap<any> | Map<PropertyKey, unknown>
+  ): void {
+    super.willUpdate(properties);
+    // as long as the controller does not support willUpdate on SSR,
+    // we need to call this manually.
+    this.typographyController.hostUpdate();
```
dunqan: What about wrapping it with `isServer`? as willUpdate will be called multiple times on the client (and calling controller should not be needed in those cases)

libs/domain/content/text/text.component.spec.ts
```diff
@@ -61,16 +61,36 @@ describe('ContentTextComponent', () => {
...
   describe('when content is provided', () => {
-    beforeEach(async () => {
-      element = await fixture(
-        html`<oryx-content-text
-          .content=${{ text: '<p>content</p>' } as ContentTextContent}
-        ></oryx-content-text>`
-      );
+    describe('featureVersion = 1.0', () => {
+      beforeEach(async () => {
+        mockFeatureVersion('1.0');
```
Sinucid: I believe this is not required, the default feature version already 1.0

libs/domain/picking/warehouse-assignment/warehouse-assignment.component.ts
```diff
@@ -60,31 +66,38 @@ export class PickingWarehouseAssignmentComponent extends LitElement {
...
 
         <oryx-image resource="no-orders"></oryx-image>
       </div>
     `;
   }
 
+  // temporary implementation for backwards compatibility
+  private __renderFallbackHeading(): TemplateResult {
+    if (featureVersion >= '1.4') {
+      return html`<oryx-heading .tag=${HeadingTag.H1}>
```
Sinucid: Is new one correct? The original has **as="h2"** attr

Sinucid (changes requested):

libs/base/ui/structure/heading/src/heading.component.ts
tobi-or-not-tobi: yes we had that and I forgot to bring it back

libs/base/ui/overlays/modal/src/modal.component.ts
tobi-or-not-tobi: yes, we do this in case of an utility that is not part of our API. It rarely happens. In this case, we want to avoid BC and alternate the changes based on featureVersion during the 1.x releases. We do not want to bloat the template logic, as it's not so easy to read. This should not pollute the formal API as well as the informal API (hence the double underscore).

libs/base/ui/structure/card/src/stories/card-demo.stories.ts
tobi-or-not-tobi: yeah we have this in the story as we also can change the icon. If an icon is used in the demo, we need to provide a heading.

Sinucid (dismissed):

dunqan (dismissed):

libs/base/ui/structure/text/text.util.ts
```diff
@@ -0,0 +1,55 @@
...
+      if (tag === 'b') tag = 'strong';
+      return `<oryx-heading tag="${tag}" ${attributes}>${content}</oryx-heading>`;
+    })
+    .replace(
+      subtitleRegex,
+      (_, el, attrBefore, cls, tag, attrAfter, content) => {
+        const attributes =
+          attrBefore || attrAfter
+            ? ` ${attrBefore ?? ''} ${attrAfter ?? ''}`
+            : '';
```
Sinucid:
```suggestion
        const attributes = ` ${attrBefore ?? ''} ${attrAfter ?? ''}`.trim();
```

libs/base/ui/structure/text/text.util.ts
```diff
@@ -0,0 +1,55 @@
...
+        const classNames = cls.replace(tag, '').trim().length
```
Sinucid: Do we need this? What about cases like **class=""**? I would say that we can pass class as is, as it was provided, without processing and clean up

libs/domain/content/text/text.component.ts
```diff
@@ -1,31 +1,50 @@
...
   protected fontService = resolve(FontService);
 
+  /**
+   * Installs font used inside the text if the autoInstallFont option is enabled.
+   * Fonts are only installed once.
+   */
   @elementEffect()
   protected installFonts = (): void => {
     const text = this.$content()?.text;
+    console.log(text);
```
Sinucid: console.log

Sinucid (changes requested):

Sinucid (approved):

PR #633, opened 2023-07-20
refactor: move fulfillment sw to application package

aleespryker:
closes: [HRZ-3311](https://spryker.atlassian.net/browse/HRZ-3311)

[HRZ-3311]: https://spryker.atlassian.net/browse/HRZ-3311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/fulfillment/sw/main.ts
```diff
@@ -1,27 +1,6 @@
...
 import { app } from './app';
 
-declare let self: ServiceWorkerGlobalScope;
-
-cleanupOutdatedCaches();
-precacheAndRoute(self.__WB_MANIFEST ?? []);
-skipWaiting();
-clientsClaim();
-
 app
```
tolerants: Suggest to move this code to the `app.ts`, so we have just one boilerplate file instead of two.

libs/template/application/sw/index.ts
```diff
@@ -0,0 +1,22 @@
+/// <reference lib="WebWorker" />
+
+// Stub window because lit it's not tree-shakeable and not compatible with service workers
+import '@spryker-oryx/utilities/window-stub';
+
+import { clientsClaim } from 'workbox-core';
```
tolerants: Please add all used dependencies to the application package.json.

tsconfig.base.json
```diff
@@ -15,6 +15,7 @@
       "@spryker-oryx/application/server": [
         "libs/template/application/server/index.ts"
       ],
+      "@spryker-oryx/application/sw": ["libs/template/application/sw/index.ts"],
```
tolerants: Suggest to rename `sw` into `service-worker`.

libs/template/application/sw/index.ts
```diff
@@ -0,0 +1,22 @@
+/// <reference lib="WebWorker" />
```
tolerants: I'd love to have it as a separate file. So we can easily add any service-worker related functionality later. But it make sense only in case we can move this code inside exported function.

tolerants (changes requested):

AndreyYevtukhov (approved):

apps/fulfillment/sw/app.ts
```diff
@@ -1,8 +1,13 @@
...
 import { fallbackEnv } from '../src/fallback-env';
 
-export const app = appBuilder()
+appBuilder()
   .withEnvironment({ ...fallbackEnv, ...process.env })
   .withFeature(offlineServiceWorkerFulfillmentFeatures())
-  .create();
+  .create()
+  .then(() => console.debug('Service worker app started!'))
+  .catch(console.error);
```
tobi-or-not-tobi, line 11: do we need this?

libs/template/application/package.json
```diff
@@ -18,7 +18,9 @@
     "@webcomponents/template-shadowroot": "^0.1.0",
     "abort-controller": "^3.0.0",
     "express": "^4.17.1",
-    "url-join": "^5.0.0"
+    "url-join": "^5.0.0",
+    "workbox-core": "^6.5.4",
+    "workbox-precaching": "^6.5.4"
```
tobi-or-not-tobi, line 23: I'd love to avoid this. Workbox is only used for apps that need sw, but storefront currently does not need this, and customers might not need this going forward. Can we not move/keep this at a lower level package instead?

tobi-or-not-tobi (changes requested):

apps/fulfillment/sw/app.ts
aleespryker: It was there in the original code and I just thought to keep it. But it doesn't show up in the logs anyways despite the service worker registering successfully.

libs/template/application/package.json
aleespryker: What package would you suggest? The ticket description wants the logic in the `application` package.

tolerants: I'd leave it as it is for now. `offline` package doesn't seem right for it, new package is also not so good option, so in case we want to reduce our boilerplate we need this code here.

tolerants (approved):

tobi-or-not-tobi (approved):

PR #634, opened 2023-07-21
[HOLD] feat: add merge strategies for experience data 

supproduction:
Add merge strategies for experience data.

closes: [HRZ-3249](https://spryker.atlassian.net/browse/HRZ-3249)

[HRZ-3249]: https://spryker.atlassian.net/browse/HRZ-3249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/src/services/experience-data/experience-data.service.ts
```diff
@@ -0,0 +1,70 @@
+import { Provider } from '@spryker-oryx/di';
+import { Component } from '@spryker-oryx/experience';
+
+export const ExperienceData = 'oryx.ExperienceData*';
+export const ExperienceDataService = 'oryx.ExperienceDataService';
+
+export const enum ExperienceDataMergeType {
+  Before = 'before',
+  After = 'after',
```
tobi-or-not-tobi, line 11: We discussed append and prepend and the dynamic behaviour depending on the target (component with child components vs component without child components. I wonder if we should not allow for also adding before/after a composition. 
What about: before, after, append and prepend?

Append and prepend will only work with components that have child components though.

libs/template/presets/storefront/experience/pages/home-page.ts
```diff
@@ -1,19 +1,23 @@
...
+export const homePage: ExperienceComponent = {
+  id: 'home',
   type: 'Page',
   meta: {
     title: 'Home Page',
     route: '/',
     description: 'Home Page Description',
   },
+  ref: {
+    header: 'prepend',
```
tobi-or-not-tobi: A ref is always added at a specific location and should nit need a merge strategy

tobi-or-not-tobi (changes requested):

libs/template/application/oryx-app/oryx-app.component.ts
```diff
@@ -10,10 +10,6 @@ export class OryxAppComponent extends LitElement {
   protected router = new LitRouter(this, []);
 
   protected override render(): TemplateResult {
-    return html`
-      <oryx-composition uid="header"></oryx-composition>
-      ${this.router.outlet()}
-      <oryx-composition uid="footer"></oryx-composition>
-    `;
+    return html` ${this.router.outlet()} `;
```
tobi-or-not-tobi:
```suggestion
    return this.router.outlet();
```

libs/template/presets/storefront/experience/header.ts
```diff
@@ -81,6 +84,7 @@ export const HeaderTemplate: StaticComponent = {
         },
         {
           type: 'oryx-composition',
+          id: 'headerNavigation',
```
tobi-or-not-tobi:
```suggestion
          id: 'headerActions',
```

libs/template/presets/storefront/experience/header.ts
```diff
@@ -48,14 +48,17 @@ export const HeaderTemplate: StaticComponent = {
...
+      id: 'headerBody',
       name: 'Composition',
       components: [
         {
           type: 'oryx-content-image',
           content: {
-            graphic: 'logo',
-            link: '/',
-            label: 'Composable Storefront based on Oryx',
+            data: {
```
tobi-or-not-tobi, line 59: Let's add an id to `logo`, so it's easy to replace.

`id: 'headerLogo'`

libs/platform/experience/src/services/resolvers/content-page-meta.resolver.ts
```diff
@@ -2,13 +2,13 @@ import { ElementResolver, PageMetaResolver } from '@spryker-oryx/core';
...
 import { Observable, combineLatest, map } from 'rxjs';
-import { ExperienceStaticData, StaticComponent } from '../experience';
+import { ExperienceComponent, ExperienceData } from '../experience';
 
 export class ContentPageMetaResolver implements PageMetaResolver {
   constructor(
     protected router = inject(RouterService),
     // Temporary: TODO use real mock data
-    protected staticData = inject(ExperienceStaticData, []).flat()
+    protected staticData = inject(ExperienceData, []).flat()
```
tobi-or-not-tobi:
```suggestion
    protected experienceData = inject(ExperienceData, []).flat()
```

libs/platform/experience/src/services/experience/preview-experience.service.ts
```diff
@@ -155,7 +155,7 @@ export class PreviewExperienceService extends DefaultExperienceService {
   protected sendStaticData(): void {
     postMessage({
       type: MessageType.Static,
-      data: this.staticData,
+      data: this.experienceData,
```
tobi-or-not-tobi, line 158: What about renaming this method to `sendData` or `sendExperienceData`

libs/template/presets/storefront/experience/footer.ts
```diff
@@ -43,19 +43,22 @@ const legalLinks = {
...
+        <a href="/">Returns & Exchanges</a><br/>
+        <a href="/">Withdrawal rights</a>`,
+    },
   },
 };
 
 const promises = {
   type: 'oryx-composition',
-  options: { rules: [{ layout: 'list', gap: 0 }] },
+  id: 'promises',
```
tobi-or-not-tobi, line 61: can we add an id for customer support and about us as well?

tobi-or-not-tobi (changes requested):

PR #638, opened 2023-07-24
chore: use vite to build fulfillment sw instead of rollup

aleespryker:
closes: [HRZ-3312](https://spryker.atlassian.net/browse/HRZ-3312)

[HRZ-3312]: https://spryker.atlassian.net/browse/HRZ-3312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/fulfillment/project.json
```diff
@@ -20,7 +20,7 @@
       "options": {
         "commands": [
           "rm -rf dev-dist/sw",
-          "npx rollup --config rollup.sw.config.mjs"
```
tolerants, line 23: Can we remove `rollup` dependency from the monorepo and from the FA `package.json` now?

tolerants (changes requested):

apps/fulfillment/vite.config.sw.ts
```diff
@@ -0,0 +1,27 @@
+import tsconfigPaths from 'vite-tsconfig-paths';
+import { defineConfig } from 'vite';
+import { Plugin } from 'vite';
```
tolerants: Hm, it looks like prettier is not working for some reason... can you please check why?
```suggestion
import { defineConfig, Plugin } from 'vite';
```

tolerants (changes requested):

apps/fulfillment/vite.config.sw.ts
aleespryker: Fixed with prettier plugin update.

tolerants (approved):

PR #643, opened 2023-07-25
feat(experience): dynamic visibility based on screen size 

Sinucid:
- dropped src folders of experience components
- components visibility rules moved to experience options rules
- added visibility rules based on screen size sm | md | lg

closes: [HRZ-3268](https://spryker.atlassian.net/browse/HRZ-3268)

[HRZ-3268]: https://spryker.atlassian.net/browse/HRZ-3268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/composition/composition-components.controller.spec.ts
```diff
@@ -1,12 +1,12 @@
...
 import { createInjector, destroyInjector, getInjector } from '@spryker-oryx/di';
-import * as litRxjs from '@spryker-oryx/utilities';
+import * as utils from '@spryker-oryx/utilities';
 import { LitElement } from 'lit';
 import { of } from 'rxjs';
 import { SpyInstance } from 'vitest';
-import { ExperienceService } from '../../src/services';
-import { Component } from '../../src/services/experience/models';
 import { CompositionComponentsController } from './composition-components.controller';
+import { ExperienceService, LayoutService } from '../src/services';
```
supproduction:
```suggestion
import { ExperienceService, LayoutService } from '"@spryker-oryx/experience"';
```

libs/platform/experience/composition/composition-components.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  protected filterHiddenComponents(
+    components: Component[],
+    breakpoint?: Size
+  ): Observable<Component[]> {
+    return combineLatest(
+      components.map((component) => {
+        //short check for the SSR
+        if (!breakpoint) {
```
supproduction:
```suggestion
        //short check for the SSR
        if (isSsr) {
```

libs/platform/experience/composition/composition-components.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+    ]).pipe(
+      switchMap(([components, breakpoint]) =>
+        components
+          ? this.filterHiddenComponents(components, breakpoint)
+          : of([])
+      )
+    );
+  }
+
+  protected hasVisibilityRules(component: Component): boolean {
```
supproduction: Can we group protected/public methods?
Can we reuse `getCurrentRules` instead, we have two methods which are doing almost the same

libs/platform/experience/composition/composition-components.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+      ({ query, hide, hideByRule }) =>
+        (hide || hideByRule) &&
+        (!query?.breakpoint || query.breakpoint === breakpoint)
+    );
+  }
+
+  getComponents(): Observable<Component[] | null> {
+    return combineLatest([
+      this.components(),
+      this.layoutService.getBreakpoint().pipe(startWith(undefined)),
```
supproduction: Why do we trigger first time with undefined?

libs/platform/experience/src/models/composition.model.ts
```diff
@@ -1,18 +1,20 @@
 import { HeadingTag } from '@spryker-oryx/ui/heading';
 import { Breakpoint } from '@spryker-oryx/utilities';
-import { LayoutAttributes } from '../../layout/src';
+import { LayoutAttributes } from '../../layout';
```
supproduction:
```suggestion
import { LayoutAttributes } from '@spryker-oryx/experience/layout';
```

libs/platform/experience/src/services/layout/default-layout.service.ts
```diff
@@ -1,15 +1,59 @@
...
 
 export class DefaultLayoutService implements LayoutService {
-  constructor(protected screenService = inject(ScreenService)) {}
+  constructor(protected screenService = inject(ScreenService)) {
+    this.setupBreakpointsObserver(document.body);
+  }
+
+  protected screenWidth$ = new ReplaySubject<number>();
+
+  protected setupBreakpointsObserver(target: HTMLElement): void {
```
supproduction:
```suggestion
  protected setupBreakpointsObserver(target?: HTMLElement): void {
```

libs/platform/experience/src/services/layout/default-layout.service.ts
```diff
@@ -1,15 +1,59 @@
...
+    const observer = new ResizeObserver(
+      throttle(
+        () =>
+          window.requestAnimationFrame(() => {
+            this.screenWidth$.next(window.innerWidth);
+          }),
+        50
+      )
+    );
+    observer.observe(target);
```
supproduction: We should unobserve whe service has been destroyed

libs/platform/experience/src/services/layout/default-layout.service.ts
```diff
@@ -1,15 +1,59 @@
...
 import { LayoutService } from './layout.service';
 import { ScreenService } from './screen.service';
 
 export class DefaultLayoutService implements LayoutService {
-  constructor(protected screenService = inject(ScreenService)) {}
+  constructor(protected screenService = inject(ScreenService)) {
+    this.setupBreakpointsObserver(document.body);
+  }
+
+  protected screenWidth$ = new ReplaySubject<number>();
```
supproduction:
```suggestion
  protected screenWidth$ = new ReplaySubject<number>(1);
```

libs/platform/experience/src/services/layout/default-layout.service.ts
```diff
@@ -1,15 +1,59 @@
...
+      }
+    )![0] as Size;
+  }
+
+  getBreakpoint(): Observable<Size> {
+    return this.screenWidth$?.pipe(
+      map((width) => this.evaluateBreakpoint(width)),
+      distinctUntilChanged()
+    );
+  }
```
supproduction: Looks like these functionality closer to `ScreenService` then to layout

supproduction (changes requested):

libs/platform/experience/composition/composition-components.controller.ts
Sinucid: dropped

libs/platform/experience/composition/composition-components.controller.ts
Sinucid: Moved this initial value to the layout service and refactored it a bit. The idea was not to block emitting of `getComponents()` observer in case when breakpoint part was not emitted due to wrong selector for observing or ssr.  For such cases breakpoint will be undefined

libs/platform/experience/src/services/layout/default-layout.service.ts
Sinucid: Not sure. Screen service is like a mediator between app and theme. I agree that method's name could be not very convenient, renamed it to `getActiveBreakpoint`

libs/base/ui/structure/collapsible-text/collapsible-text.component.spec.ts
```diff
@@ -5,7 +5,7 @@ import { ButtonComponent, ButtonType } from '../../action/button/src';
 import { CollapsibleTextComponent } from './collapsible-text.component';
 import { collapsibleTextComponent } from './collapsible-text.def';
 import { CollapsibleTextToggle } from './collapsible-text.model';
-import './ponyfill';
+import '@/tools/testing';
```
tobi-or-not-tobi: could we drop `// organize-imports-ignore` because of this?

libs/platform/experience/composition/composition-components.controller.ts
```diff
@@ -1,22 +1,45 @@
...
+    ]).pipe(
+      switchMap(([components, breakpoint]) =>
+        components
+          ? this.filterHiddenComponents(components, breakpoint)
+          : of([])
+      )
+    );
+  }
+
+  hasDynamicallyVisibleChild(): Observable<boolean> {
```
tobi-or-not-tobi: I wander if we can improve the name, but while going through code I did not really get what we're doing here. I was expecting this code to check if the child would have a `visible` rule, but did not see any of those.

libs/platform/experience/composition/composition-components.controller.ts
```diff
@@ -32,47 +55,38 @@ export class CompositionComponentsController implements ReactiveController {
   }
 
   protected filterHiddenComponents(
-    components: Component[]
+    components: Component[],
+    breakpoint?: Size
```
tobi-or-not-tobi:
```suggestion
    screenSize?: Size
```

tobi-or-not-tobi: or just size

libs/platform/experience/composition/composition-components.controller.ts
```diff
@@ -32,47 +55,38 @@ export class CompositionComponentsController implements ReactiveController {
...
-
-  hasDynamicallyVisibleChild(): Observable<boolean> {
-    return this.components().pipe(
-      map(
-        (components) =>
-          !!components?.some((component) => !!component.options?.visibility)
-      )
+  protected getRules(
+    component: Component,
+    breakpoint?: Size
```
tobi-or-not-tobi:
```suggestion
    screenSize?: Size
```

tobi-or-not-tobi (changes requested):

libs/platform/experience/composition/composition-components.controller.ts
Sinucid: It is doing exactly that you described - try to find at least 1 component inside with visibility rules (has hide === true or hideByRule). I changed its name to `hasDynamicallyVisibleComponent`, to make it a bit more clear, until we have better ideas

libs/platform/experience/src/services/layout/default-layout.service.ts
Sinucid: Discussed with @tobi-or-not-tobi and moved related logic to the screen service

libs/platform/experience/src/services/layout/default-screen.service.spec.ts
```diff
@@ -9,10 +9,11 @@ const mockTheme: Theme = {
   name: 'name',
   breakpoints: {
     [Size.Sm]: {
-      min: 0,
+      max: 767,
     },
     [Size.Md]: {
       min: 768,
+      max: 1023,
```
tobi-or-not-tobi, line 16: we didn't need those initially as we'd get them from the next size, but I can imagine that there's no logical way neccessarily to figure what's the "next" screen size. No big deal to change this config, but like to understand why we did this.

tobi-or-not-tobi (approved):

supproduction (approved):

PR #644, opened 2023-07-25
test: update cypress to the latest version

AndreyYevtukhov:
Updates Cypress to the latest version 

closes: https://spryker.atlassian.net/browse/HRZ-3330


supproduction (approved):

package.json
```diff
@@ -102,7 +101,7 @@
     "chromatic": "^6.5.1",
     "concurrently": "^7.6.0",
     "copyfiles": "^2.4.1",
-    "cypress": "^12.7.0",
```
tolerants, line 105: It better to keep it without patches at least and it so precise version only if we really need it (there is a bugfix or improvement we are relying on).

tolerants (approved):

package.json
AndreyYevtukhov: But we don't use `npm i` in our workflows, so the version from `package-lock.json` will be picked up in any case, and it should not cause problems (as far as I see it)

PR #651, opened 2023-07-27
fix: update modal bleed for fulfillment

aleespryker:
A minimum modal must have a little bleed to not stick to the window. 

closes: [HRZ-3319](https://spryker.atlassian.net/browse/HRZ-3319)

[HRZ-3319]: https://spryker.atlassian.net/browse/HRZ-3319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/themes/design-tokens/src/backoffice-ng/index.ts
```diff
@@ -45,5 +45,8 @@ export const backofficeNgTokens: DesignToken[] = [
     },
     ...layoutSmTokens,
     ...commonTokensSmall,
+    modal: {
+      bleed: '20px',
+    },
```
tobi-or-not-tobi: can you try the following:
- drop this specific `modal bleed`
- drop layout tokens from `backoffice-ng/layout.tokens.ts`
- reuse layout tokens from `design-tokens/src/layout-tokens.ts` in backoffice-ng

I do not expect any visual issues with this approach, and its valueable to use the same layout system cross all apps

tobi-or-not-tobi (changes requested):

libs/template/themes/design-tokens/src/backoffice-ng/index.ts
aleespryker: Unfortunately, the default layout-token does not have the 20px modal bleed we need to have in FA. I can put the 20px bleed in the default layout tokens, but then all the other apps using it will have it too.

tobi-or-not-tobi: Understand. 

We should only add the margin on the `minimal` modals. Agreed that this is not fixed with the suggestion above. What about the following:

```css
:host([minimal]) dialog {
  min(width: var(--oryx-modal-width, 510px), calc(100vw - 20px));
}
```

tobi-or-not-tobi (approved):

PR #657, opened 2023-07-27
test: improve empty cart checks

AndreyYevtukhov:
Improves cart emptiness check and prevents a regression with empty cart block shown above the cart entries

closes: -

### DIFF
--- a/apps/storefront-e2e/src/support/page-objects/cart.page.ts
+++ b/apps/storefront-e2e/src/support/page-objects/cart.page.ts
@@ -8,11 +8,11 @@ export class CartPage extends AbstractSFPage {
   private cartTotals = new CartTotalsFragment();
 
   waitForLoaded(): void {
-    this.getEmptyCartMessage().should('be.visible');
+    this.getCartEntriesWrapper().should('exist');
   }
 
   getCartEntriesWrapper = () => cy.get('oryx-cart-entries');
-  getEmptyCartMessage = () => cy.get('oryx-content-text');
+  getEmptyCartMessage = () => cy.contains('Your shopping cart is empty');
   getCartEntries = () =>
     this.getCartEntriesWrapper()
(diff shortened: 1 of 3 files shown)


apps/storefront-e2e/src/support/page-objects/cart.page.ts
```diff
@@ -8,11 +8,11 @@ export class CartPage extends AbstractSFPage {
   private cartTotals = new CartTotalsFragment();
 
   waitForLoaded(): void {
-    this.getEmptyCartMessage().should('be.visible');
+    this.getCartEntriesWrapper().should('exist');
   }
 
   getCartEntriesWrapper = () => cy.get('oryx-cart-entries');
-  getEmptyCartMessage = () => cy.get('oryx-content-text');
+  getEmptyCartMessage = () => cy.contains('Your shopping cart is empty');
```
tolerants, line 15: It's not get then.

tolerants (changes requested):

apps/storefront-e2e/src/support/page-objects/cart.page.ts
AndreyYevtukhov: `contains` is a getter - https://docs.cypress.io/api/commands/contains

AndreyYevtukhov: > Get the DOM element containing the text. DOM elements can contain more than the desired text and still match. Additionally, Cypress [prefers some DOM elements](https://docs.cypress.io/api/commands/contains#Notes) over the deepest element found.

tolerants: Ah, I see. But it's better to not rely on the text...

tolerants (approved):

PR #663, opened 2023-07-28
fix(picking): redirect to main page from `WarehouseAssignmentComponent` if there is only one warehouse

nsilva-spryker:
When loading the warehouses, if there is only one warehouse, redirect to main page.

closes: [HRZ-3322](https://spryker.atlassian.net/browse/HRZ-3322)

[HRZ-3322]: https://spryker.atlassian.net/browse/HRZ-3322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/picking/warehouse-assignment/warehouse-assignment.component.ts
```diff
@@ -100,6 +100,11 @@ export class WarehouseAssignmentComponent extends LitElement {
   }
 
   protected renderLoading(): TemplateResult {
+    const locations = this.$locations();
+    if (locations?.length === 1) {
+      this.onSelect(locations[0].id);
```
tolerants: I have some concern regarding calling a method with the name `onSelect`. And another question is it a good idea to have this redirect in render function? Maybe we can make it earlier as long as we are not depending on user interaction?
Also it looks like it's worth to cover with tests.

tolerants (changes requested):

tolerants:
Can you please add a scope `picking` for the fix. And also mention the fix was done in `WarehouseAssignmentComponent`.

libs/domain/picking/warehouse-assignment/warehouse-assignment.component.ts
nsilva-spryker: The render() method is dispatch every time the location property changes.
The firstupdate() lifecycle  could be a good approach, but when reached the locations$ property his still null.
Not sure there is any other alternative.
I will rename the method and update unit tests

tolerants (approved):

PR #670, opened 2023-07-28
feat(router): authenticated users will be redirected from login page when accidentally navigate to it 

Sinucid:
- separate route for login page was created with **enter** hook for it that perform check and redirect already authenticated users to the home page when they accidentally navigate to the login page
- link service was improved with proper handling of routes without dynamic parts
- dropped start value of resolved token in token resolver service

closes: [HRZ-3304](https://spryker.atlassian.net/browse/HRZ-3304)

[HRZ-3304]: https://spryker.atlassian.net/browse/HRZ-3304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/src/routes.ts
```diff
@@ -7,6 +7,11 @@ export const defaultExperienceRoutes: RouteConfig[] = [
   {
     pattern: new URLPattern({ pathname: '/{index.html}?' }),
   },
+  {
```
tobi-or-not-tobi, line 13: I'd rather keep route guards at the page setup. We will eventually manage the access of pages in eb. For example a b2b shop is often partially "closed", the business would configure which pages are "open"

tobi-or-not-tobi (changes requested):

libs/platform/core/src/services/token-resolver/default-token-resolver.service.ts
```diff
@@ -29,6 +29,7 @@ export class DefaultTokenService implements TokenResolver {
 
     return tokenResolver.resolve(resolver).pipe(
       startWith(null),
+      skip(1),
```
supproduction: Looks like this we can omit these parts

libs/platform/experience/src/routes.ts
```diff
@@ -7,6 +7,11 @@ export const defaultExperienceRoutes: RouteConfig[] = [
...
+    path: '/login',
+    type: RouteType.Login,
+    redirect: { when: 'USER.AUTHENTICATED' },
```
supproduction: Lets use enter guard instead

supproduction (changes requested):

libs/platform/experience/src/routes.ts
supproduction: Also lets move this route into authentication package

libs/platform/experience/src/routes.ts
Sinucid: Good idea, but I would prefer to keep both implementation (for the route directly, and for the page on EB side) EB will override or get the default value from route. But I can imagine that this is a lot of work, lets skip it for now, because it is a candidate for the big feature

libs/platform/experience/src/routes.ts
Sinucid: `enter` hook came from the source of lit-router implementation and its aim like `beforeMount` (execute some code before route fully processed). It does not make any redirection under the hood. I would suggest to keep it as is and implement separate `redirect` method.

supproduction: But I don't see any reason why we need this redirect method. We already have custom implementation of enter function let's adjust it for our needs.

supproduction (approved):

tobi-or-not-tobi (approved):

libs/domain/site/src/services/link/default-link.service.ts
```diff
@@ -1,8 +1,8 @@
 import { inject } from '@spryker-oryx/di';
```
tolerants, line 1: Can you please add some more details into PR description.

tolerants (changes requested):

tolerants (approved):

PR #673, opened 2023-07-31
ci: remove lighthouse plugin pemporarily

AndreyYevtukhov:
update lighthouse plugin version

closes: -


AndreyYevtukhov:
This PR was needed only to trigger Netlify plugin version update.
Now it is installed, so we can close it

PR #679, opened 2023-07-31
fix: netlify caching issue

Sinucid:
closes: [HRZ-3297](https://spryker.atlassian.net/browse/HRZ-3297)

[HRZ-3297]: https://spryker.atlassian.net/browse/HRZ-3297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/template/application/lambda/src/handler.lambda.ts
+++ b/libs/template/application/lambda/src/handler.lambda.ts
@@ -45,7 +45,7 @@ export const storefrontHandler = async (
         ...event.headers,
       },
       body,
-      ...(hasQueryParams
+      ...(!hasQueryParams
         ? { ttl: process.env.ORYX_TTL ? Number(process.env.ORYX_TTL) : ttl }
         : {}),
     };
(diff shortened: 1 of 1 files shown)


libs/template/application/lambda/src/handler.lambda.ts
```diff
@@ -44,7 +45,9 @@ export const storefrontHandler = async (
         ...event.headers,
       },
       body,
-      ttl: process.env.ORYX_TTL ? Number(process.env.ORYX_TTL) : ttl,
+      ...(!hasQueryParams
+        ? { ttl: process.env.ORYX_TTL ? Number(process.env.ORYX_TTL) : ttl }
+        : {}),
```
tobi-or-not-tobi, line 50: Let's not play with TTL for this. I'd rather turn params into the path as we want to have it anyways for better UX and SEO.

tobi-or-not-tobi (changes requested):

PR #687, opened 2023-08-02
fix(user): don't update UI on validity triggered by change event in `UserAddressFormComponent`

aleespryker:
closes: [HRZ-3293](https://spryker.atlassian.net/browse/HRZ-3293)

[HRZ-3293]: https://spryker.atlassian.net/browse/HRZ-3293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/user/address-form/address-form.component.ts
+++ b/libs/domain/user/address-form/address-form.component.ts
@@ -171,28 +171,24 @@ export class UserAddressFormComponent
     this.countryService.set(this.country);
   }
 
-  protected preventInvalidDisplay = (e: Event): void => {
-    e.stopImmediatePropagation();
-  };
-
   protected onChange(): void {
     if (this.form) {
       const address: Address = Object.fromEntries(
         new FormData(this.form).entries()
       );
       address.id = this.values?.id;
-      this.form.addEventListener('invalid', this.preventInvalidDisplay, true);
       this.dispatchEvent(
         new CustomEvent<AddressEventDetail>('change', {
-          detail: { address, valid: this.form.checkValidity() },
+          detail: {
+            address,
+            // We don't want to trigger valdity in the UI, just store details in storage.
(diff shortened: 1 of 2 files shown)


libs/domain/user/address-form/address-form.component.ts
```diff
@@ -171,19 +171,29 @@ export class UserAddressFormComponent
     this.countryService.set(this.country);
   }
 
+  protected preventInvalidDisplay = (e: Event): void => {
+    e.stopImmediatePropagation();
```
Sinucid: Probably better to remove event listener inside handler, after the first handling, to make the behavior stable, ie:

```
protected preventInvalidDisplay = (e: Event): void => {
    e.stopImmediatePropagation();

    this.form.removeEventListener(
        'invalid',
        this.preventInvalidDisplay,
        true
    );
};
```

Sinucid: And also leave a comment about the purpose of this hack, and cover new functionality with tests

Sinucid (changes requested):

libs/domain/user/address-form/address-form.component.ts
aleespryker: I wanted to do it like that, but `invalid` does not bubble and only fires on elements inside the form, not the form itself, so removing inside the handler will not remove all of the listeners, just the first one. Or it just gets removed too early.

At least that's what happened when I played with it.

Sinucid (approved):

tolerants:
@aleespryker It's impossible to understand what was affected by this issue and fix. Can you please specify a component?

tolerants (approved):

PR #689, opened 2023-08-03
fix(picking): fix z-index for picking lists header in `pickingListsComponent`

aleespryker:
closes: [HRZ-89546](https://spryker.atlassian.net/browse/HRZ-89546)

[HRZ-89546]: https://spryker.atlassian.net/browse/HRZ-89546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/picking/picking-lists-header/picking-lists-header.styles.ts
+++ b/libs/domain/picking/picking-lists-header/picking-lists-header.styles.ts
@@ -1,6 +1,10 @@
 import { css } from 'lit';
 
 export const pickingListsHeaderComponentStyles = css`
+  :host {
+    z-index: 1;
+  }
+
   oryx-header {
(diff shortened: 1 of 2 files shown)


libs/domain/picking/picking-lists-header/picking-lists-header.styles.ts
```diff
@@ -1,6 +1,10 @@
 import { css } from 'lit';
 
 export const pickingListsHeaderComponentStyles = css`
+  :host {
```
tolerants: As for me it should be done in picking lists component, where we are using it. Component itself shouldn't specify it's z-index.

tolerants (changes requested):

tolerants (approved):

PR #695, opened 2023-08-08
feat(ui): breadcrumbs component

Sinucid:
- breadcrumbs ui component
- demo and static story
- unit tests

closes: [HRZ-2263](https://spryker.atlassian.net/browse/HRZ-2263)

[HRZ-2263]: https://spryker.atlassian.net/browse/HRZ-2263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/navigation/breadcrumbs/breadcrumbs.model.ts
```diff
@@ -0,0 +1,11 @@
+import { IconTypes } from '../../graphical/icon/src';
+
+export interface Breadcrumb {
+  label: string;
+  url?: string;
+}
+
+export interface BreadcrumbsProperties {
+  breadcrumbs?: Breadcrumb[];
```
tolerants: Please add documentation, so customer can understand what is the structure and how to use it.

libs/base/ui/navigation/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,10 @@
+import { html, LitElement, TemplateResult } from 'lit';
+import { breadcrumbsStyles } from './breadcrumbs.styles';
+
+export class BreadcrumbsComponent extends LitElement {
+  static styles = breadcrumbsStyles;
+
+  protected override render(): TemplateResult {
+    return html`<slot></slot>`;
```
tolerants, line 40: I don't really like we have a component without any props. So from one perspective we provide some freedom for customers, but from another:
- API is unclear, what should be passed? How to use this component?
- styles are still bound to specific markup
- to use this component developer should manually prepare the whole template

Suggest to add a `breadcrumbs` and `divider` properties and add rendering logic here.

libs/base/ui/navigation/breadcrumbs/breadcrumbs.styles.ts
```diff
@@ -0,0 +1,44 @@
...
+export const breadcrumbsStyles = css`
+  :host {
+    display: flex;
+    align-items: center;
+    gap: 8px;
+    flex-wrap: wrap;
+  }
+
+  ::slotted(oryx-icon) {
+    --_size: 12px;
```
tolerants: Why don't you want to leverage `--oryx-icon-size`?

libs/base/ui/navigation/breadcrumbs/breadcrumbs.def.ts
```diff
@@ -0,0 +1,12 @@
+import { componentDef } from '@spryker-oryx/utilities';
+
+export const breadcrumbsComponent = componentDef({
+  name: 'oryx-breadcrumbs',
```
tolerants, line 4:
```suggestion
  name: 'oryx-ui-breadcrumbs',
```

libs/base/ui/navigation/breadcrumbs/stories/demo.stories.ts
```diff
@@ -0,0 +1,65 @@
...
+  {
+    label: 'Home',
+    url: '/',
+  },
+  {
+    label: 'Cameras & Camcorders',
+    url: '/category',
+  },
+  {
+    label: 'Digital Cameras',
```
tolerants: I believe last part of breadcrumbs should be just text. I mean we can support both link and text, but at least we should check we have a variation with text.

libs/base/ui/navigation/breadcrumbs/breadcrumbs.model.ts
```diff
@@ -0,0 +1,11 @@
+import { IconTypes } from '../../graphical/icon/src';
```
tolerants:
```suggestion
import { IconTypes } from '@spryker-oryx/ui/icon';
```

tolerants (changes requested):

libs/base/ui/navigation/breadcrumbs/breadcrumbs.model.ts
Sinucid: it is outdated, dropped

libs/base/ui/navigation/breadcrumbs/breadcrumbs.def.ts
Sinucid: Should our ui component have domain prefix? Currently no one of them have it.

libs/base/ui/navigation/breadcrumbs/breadcrumbs.component.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+      await expect(element).shadowDom.to.be.accessible(a11yConfig);
+    });
+
+    it('should render slot', async () => {
+      expect(slot).toBeDefined();
+    });
+
+    it('should render all slotted elements', async () => {
+      expect(slot.assignedElements().length).toBe(5);
+    });
```
tarashaliuk: those functions shouldn't be async

tarashaliuk (changes requested):

libs/base/ui/navigation/breadcrumbs/breadcrumbs.def.ts
tolerants: Yep you're right.

libs/base/ui/navigation/breadcrumbs/breadcrumbs.model.ts
Sinucid: dropped

libs/base/ui/navigation/breadcrumbs/breadcrumbs.component.ts
Sinucid: You are right. But I tried different approaches, with props and without and all of them have their own drawbacks or limitations. 
For example **dividers**, they are several between breadcrumbs links, and only one thing that we can do to customise them - specify the type of the icon, that has poor flexibility (customer can not provide its own divider or provide different types of them for single breadcrumbs, or if the design does not have dividers between links). Also populating of breadcrumbs like a property has many limitations, for example we are limited to use our own links components, only that hardcoded inside components, or if we need to render icon instead text for the main **Home** breadcrumb.

Slotting of the links the most flexible and simple way. Yes, in result we get just a styled component, but I believe its enough for the main purposes. Developers have enough flexibility to make some customisations, or if need something more complex - use their own solution. The main part - resolving the breadcrumbs data will be the responsibility of another component.   

I will provide a little doc to make component's purpose a bit more clear

tolerants: I understand drawbacks and limitations, but I think in the most of the cases the usage of this components will be really simple - just one divider of the same type or no divider and a few link/text nodes. That's it. And in rare cases when customer will need to customize rendering he can use the slot to replace the whole template or even extend this component.

But as for me it's really important to support an simple usage of our components for easy cases, so I'd put some effort and provide some rendering logic in this component to not force customers provide templates by themselfs, especially keeping in mind that we are still limiting their html with our default css.

In case you're still not convenced - let's have a call, or we can have it together with @dunqan to have a side view.

Sinucid:
obsolete

PR #697, opened 2023-08-08
fix(router): fix displaying not found page for simple/product:sku/category:id routes

supproduction:
closes [HRZ-3282](https://spryker.atlassian.net/browse/HRZ-3282)

[HRZ-3282]: https://spryker.atlassian.net/browse/HRZ-3282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AndreyYevtukhov:
Let's also fix https://spryker.atlassian.net/browse/HRZ-89565 in this PR (if possible)

supproduction:
> Let's also fix [spryker.atlassian.net/browse/HRZ-89565](https://spryker.atlassian.net/browse/HRZ-89565) in this PR (if possible)

done

libs/template/labs/src/articles/storyblok/storyblok.adapter.ts
```diff
@@ -15,21 +15,23 @@ export class StoryblokAdapter implements ContentAdapter {
   }
 
   getAll(qualifier: ContentQualifier): Observable<Content[] | null> {
+    console.log(qualifier, 'qualifier');
```
dunqan: 👀

libs/platform/experience/src/routes.ts
```diff
@@ -18,29 +18,16 @@ export const defaultExperienceRoutes: RouteConfig[] = [
...
-              resolve(RouterService).navigate('/');
-            }
-
-            return !state;
-          })
+          map((state) => (state ? '/' : !state))
         ),
   },
-  {
-    path: '/product/:sku',
```
dunqan: what about backyard's compatibility? 
I think we discussed we should keep it?

libs/platform/experience/src/feature/latest.ts
```diff
@@ -0,0 +1,36 @@
...
+import { defaultExperienceLatestRoutes } from '../routes';
+import {
+  experiencePreviewProviders,
+  experienceProviders,
+  layoutProviders,
+} from '../services';
+
+export const experienceLatestComponents = [components.compositionComponent];
+
+export const layoutLatestFeature: AppFeature = {
```
dunqan: if there is no difference, I'd go with simple
`layoutLatestFeature = layoutFeature` for such cases. Only if we will add something new, we can provide more customized sets.

libs/platform/experience/src/routes.ts
```diff
@@ -18,31 +18,30 @@ export const defaultExperienceRoutes: RouteConfig[] = [
...
   {
     path: '/*',
+    type: RouteType.NotFound,
     render: (): TemplateResult =>
       html`<oryx-heading><h1>Error 404</h1></oryx-heading>
         <p>Page not found</p>`,
   },
 ];
+
+export const defaultExperienceRoutes: RouteConfig[] = [
```
dunqan, line 40: change order of elements in const array could also be considered a breaking change

libs/platform/experience/src/routes.ts
```diff
@@ -1,12 +1,12 @@
...
 import { resolve } from '@spryker-oryx/di';
-import { RouterService, RouteType } from '@spryker-oryx/router';
+import { RouteType } from '@spryker-oryx/router';
 import { RouteConfig } from '@spryker-oryx/router/lit';
 import { html, TemplateResult } from 'lit';
 import { map, take } from 'rxjs';
 import 'urlpattern-polyfill';
 
-export const defaultExperienceRoutes: RouteConfig[] = [
+export const defaultExperienceLatestRoutes: RouteConfig[] = [
```
dunqan: I'm wondering about just naming it `experienceRoutes`, and use it in latest feature instead of the old one.

libs/template/presets/storefront/app/stable/experience/experience-data.ts
```diff
@@ -0,0 +1,39 @@
+import { AppFeature } from '@spryker-oryx/core';
```
dunqan: As before, I'd try to avoid this much of code duplication. 
If latest has no difference with current/stable, I'd just reuse stable. 

In case of static data, we may think about additional providers for latest, that could even just use static data merging strategies (on top of stable data).

libs/platform/experience/src/feature/feature.ts
```diff
@@ -1,13 +1,13 @@
 import { AppFeature } from '@spryker-oryx/core';
```
tolerants: I don't really like the naming. As I understand here we have an initial version of feature set. Maybe let's call it `initial`? Or `1_0`?

libs/platform/experience/src/feature/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './feature';
```
tolerants: I don't really like that customer is not able to lock to the `1_1` version. So suggest to add both: `1_1` and `latest` even they are identical in this case.

tolerants (changes requested):

libs/platform/experience/src/feature/feature.ts
tolerants: `1_0`.

libs/platform/experience/src/feature/index.ts
tolerants: `1_1` => `latest`.

libs/domain/search/src/services/index.ts
```diff
@@ -1,10 +1,11 @@
 // organize-imports-ignore
```
tolerants, line 1: Do we need it? And is it even working?

supproduction: IDK, but I remember that we had an issue in production without it, we should retest it

tolerants (dismissed):

apps/storefront/src/app.ts
```diff
@@ -1,12 +1,12 @@
...
 import { labsFeatures } from '@spryker-oryx/labs';
-import { storefrontFeatures } from '@spryker-oryx/presets/storefront';
+import { storefrontFeatures_1_1 } from '@spryker-oryx/presets/storefront';
 import { storefrontTheme } from '@spryker-oryx/themes';
 
 const env = import.meta.env;
 
 const features = [
-  ...storefrontFeatures,
+  ...storefrontFeatures_1_1,
```
dunqan: our app should use latest always?

libs/platform/experience/src/routes.ts
```diff
@@ -18,13 +18,35 @@ export const defaultExperienceRoutes: RouteConfig[] = [
...
+    type: RouteType.NotFound,
+    render: (): TemplateResult =>
+      html`<oryx-heading><h1>Error 404</h1></oryx-heading>
+        <p>Page not found</p>`,
+  },
+];
 
-            return !state;
-          })
+export const defaultExperienceRoutes: RouteConfig[] = [
```
dunqan, line 40: what about adding deprecation comments?

dunqan (dismissed):

libs/platform/experience/src/routes.ts
```diff
@@ -18,13 +18,38 @@ export const defaultExperienceRoutes: RouteConfig[] = [
...
+    render: (): TemplateResult =>
+      html`<oryx-heading><h1>Error 404</h1></oryx-heading>
+        <p>Page not found</p>`,
+  },
+];
 
-            return !state;
-          })
+/**
+ * @deprecated: Since 1.1 version will be deleted. Use `experienceRoutes` instead.
```
tolerants: We will remove it in 2.0 most probably. So let's remove the message about deletion.
```suggestion
 * @deprecated: Since 1.1 version. Use `experienceRoutes` instead.
```

libs/platform/experience/src/routes.ts
```diff
@@ -18,13 +18,38 @@ export const defaultExperienceRoutes: RouteConfig[] = [
         .resolveToken('USER.AUTHENTICATED')
         .pipe(
           take(1),
-          map((state) => {
```
tolerants, line 21: Don't get it why we are changing current behaviour - it's a bug fix?

tolerants (changes requested):

libs/platform/experience/src/routes.ts
supproduction: the behaviour is the same we updated lit-router too to support this one

tolerants (approved):

PR #698, opened 2023-08-08
chore: adjust order totals e2e tests

tarashaliuk:
Adjusted e2e tests to test order totals

closes: https://spryker.atlassian.net/browse/HRZ-2810

### DIFF
--- a/apps/storefront-e2e/src/integration/checkout.cy.ts
+++ b/apps/storefront-e2e/src/integration/checkout.cy.ts
@@ -38,7 +38,7 @@ describe('Checkout suite', () => {
         thankYouPage.getOrderEntriesWrapper().should('be.visible');
         thankYouPage.getOrderEntries().should('have.length.at.least', 1);
 
-        thankYouPage.getOrderTotals.getWrapper().should('not.exist');
+        thankYouPage.getOrderTotals().getWrapper().should('not.exist');
 
         verifyThatCartWasCleared();
       });
@@ -74,8 +74,8 @@ describe('Checkout suite', () => {
         thankYouPage.getOrderEntriesWrapper().should('be.visible');
         thankYouPage.getOrderEntries().should('have.length.at.least', 1);
 
-        thankYouPage.getOrderTotals.getWrapper().should('be.visible');
-        thankYouPage.getOrderTotals.checkTotals({
+        thankYouPage.getOrderTotals().getWrapper().should('be.visible');
+        thankYouPage.getOrderTotals().checkTotals({
           subTotal: '€106.80',
           discountsTotal: '-€29.90',
           taxTotal: '€13.06',
--- a/apps/storefront-e2e/src/integration/checkout.cy.ts
+++ b/apps/storefront-e2e/src/integration/checkout.cy.ts
@@ -38,7 +38,7 @@ describe('Checkout suite', () => {
         thankYouPage.getOrderEntriesWrapper().should('be.visible');
         thankYouPage.getOrderEntries().should('have.length.at.least', 1);
 
-        thankYouPage.orderTotalsFragment.getWrapper().should('not.exist');
+        thankYouPage.getOrderTotals.getWrapper().should('not.exist');
 
         verifyThatCartWasCleared();
       });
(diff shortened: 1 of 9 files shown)


apps/storefront-e2e/src/integration/checkout.cy.ts
```diff
@@ -66,6 +71,30 @@ describe('Checkout suite', () => {
           .and('contain', 'Billing address')
           .and('contain', 'Delivery address');
 
+        thankYouPage.getOrderEntriesWrapper().should('be.visible');
+        thankYouPage.getOrderEntries().should('have.length.at.least', 1);
+
+        thankYouPage.orderTotalsFragment.getWrapper().should('be.visible');
```
AndreyYevtukhov: We have a function that checks cart totals prices already here https://github.com/spryker/oryx/blob/development/apps/storefront-e2e/src/integration/cart.cy.ts#L155

Let's move it to https://github.com/spryker/oryx/blob/development/apps/storefront-e2e/src/support/cy-actions/cart.actions.ts and use it in the cart suite and here

apps/storefront-e2e/src/support/page-fragments/totals.fragment.ts
```diff
@@ -1,5 +1,11 @@
-export class CartTotalsFragment {
-  getWrapper = () => cy.get('oryx-cart-totals');
+export class TotalsFragment {
```
AndreyYevtukhov, line 1: This fragment is used in 3 different places, but it is different everywhere 

https://github.com/spryker/oryx/pull/698/files#diff-c53e5dd346244eee79cbbfee128a66db8bdecfda4a019cc634fd92f6376d126aR8

https://github.com/spryker/oryx/pull/698/files#diff-6c3c0d4b06a47339eda5a1c49499ad2caa8f6f4c935395210e43a098189c6773R38

https://github.com/spryker/oryx/pull/698/files#diff-86856059ec914c9e7a08d9dd3bf61fdab5fd7d9f4f7f3552c99bc9ff67426162R14

Can we change it to make it consistent?

AndreyYevtukhov (changes requested):

apps/storefront-e2e/src/support/page-fragments/totals.fragment.ts
tarashaliuk: There are different wrappers in those places (`oryx-order-totals` and `oryx-cart-totals`). That’s why I renamed the fragment and added a constructor param (wrapper: string)

apps/storefront-e2e/src/integration/checkout.cy.ts
tarashaliuk: Can we place the `checkTotals` function in `TotalFragment` or in separate `totals.actions.ts` because it isn't only for the cart process but also for the checkout?

AndreyYevtukhov: Lets put it in `TotalFragment`, should be fine

apps/storefront-e2e/src/support/page-fragments/totals.fragment.ts
AndreyYevtukhov: class is fine, I mean usage:

```
// private member
private cartTotals = new TotalsFragment('oryx-cart-totals');
```

```
// getter
getCartTotals = new TotalsFragment('oryx-cart-totals');
```

```
// public member with initialization in constructor
this.orderTotalsFragment = new TotalsFragment('oryx-order-totals');
```

we can make it `getter` everywhere

apps/storefront-e2e/src/support/page-objects/cart.page.ts
```diff
@@ -23,7 +21,7 @@ export class CartPage extends AbstractSFPage {
       });
   getCartEntriesHeading = () =>
     this.getCartEntriesWrapper().find('oryx-heading');
-  getCartTotals = () => this.cartTotals;
+  getCartTotals = new TotalsFragment('oryx-cart-totals');
```
AndreyYevtukhov: Can we make it a getter, not a parameter, aka `getCartTotals = () => new ....`?

AndreyYevtukhov (changes requested):

AndreyYevtukhov (approved):

PR #703, opened 2023-08-11
feat(preset): add b2b-storefront preset

tolerants:
- Add `b2b-storefront` preset
- Add `ORYX_PRESET` variable to be able to load `b2b-storefront` or `storefront` preset for `storefront` app
- Configured b2b e2e tests
- Added a basic b2b e2e test for contact us page

closes: [HRZ-89464](https://spryker.atlassian.net/browse/HRZ-89464)

[HRZ-89464]: https://spryker.atlassian.net/browse/HRZ-89464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/.env.template
```diff
@@ -26,3 +26,7 @@ ORYX_CONTENTFUL_SPACE=
 
 # The Storyblok Access Token
 ORYX_STORYBLOK_TOKEN=
+
+# Used to set a preset, for now if it's not empty then b2b preset will be loaded.
+# If it's empty then storefront preset will be loaded.
```
dunqan: I'd add a list of available options. Also not empty is not intuitive, i.e. ORYX_PRESET=B2C should not result in B2B, what do you think? Maybe we should add bit more logic to make it already work as intended?

tolerants: I wanted to keep it as simple as possible, but you're right. In any case we will need to make it at some point.

dunqan (approved):

PR #705, opened 2023-08-11
feat(site): breadcrumbs component and service

Sinucid:
- added a couple of specific breadcrumbs resolvers for product list and category pages and default one
- new method `current` was added to the RouterService that gets the combination of routeConfig with current query and params
- resolve was replaced by injectors`s inject in token resolvers services
- FacetQualifier was improved. The required `name` param was made optional and additional optional qualifier `parameter` was added
- added unit tests for the new functionality 
- added demo and static stories for the breadcrumbs component
- improved tests coverage of `DefaultFacetListService`

closes: [HRZ-89551](https://spryker.atlassian.net/browse/HRZ-89551)

[HRZ-89551]: https://spryker.atlassian.net/browse/HRZ-89551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/presets/storefront/experience/pages/category-page.ts
```diff
@@ -19,6 +19,17 @@ export const categoryPage: ExperienceComponent = {
     ],
   },
   components: [
+    {
```
dunqan: we can't add it to existing presets as it will be a breaking change, we need to create "latest" feature.

libs/domain/site/src/services/breadcrumbs/breadcrumbs.service.ts
```diff
@@ -0,0 +1,24 @@
...
+export type BreadcrumbsQualifier = {
+  type: RouteType;
+};
+
+export const BreadcrumbsService = 'oryx.BreadcrumbsService';
+export const BreadcrumbsResolvers = 'oryx.BreadcrumbsResolver*';
+
+declare global {
+  interface InjectionTokensContractMap {
+    [BreadcrumbsService]: BreadcrumbsService;
```
dunqan, line 17: what about adding also resolvers to this map?

libs/platform/router/src/services/router.service.ts
```diff
@@ -9,6 +9,8 @@ export interface RouterService {
   getEvents(type: RouterEventType): Observable<RouterEvent>;
   route(): Observable<string>;
   currentRoute(): Observable<string>;
+  currentRouteWithParams(): Observable<RouteWithParams>;
+  setCurrentRouteConfig(config: RouteConfig): void;
   currentParams(): Observable<RouteParams>;
```
dunqan, line 13: Can we somehow avoid adding this method to public API of this service?

libs/platform/router/src/services/router.service.ts
```diff
@@ -9,6 +9,8 @@ export interface RouterService {
   getEvents(type: RouterEventType): Observable<RouterEvent>;
   route(): Observable<string>;
   currentRoute(): Observable<string>;
+  currentRouteWithParams(): Observable<RouteWithParams>;
```
dunqan: maybe just `current` with all the data for the route, so we could.deprecate old ones?

libs/domain/site/src/services/breadcrumbs/default-breadcrumbs.service.ts
```diff
@@ -0,0 +1,57 @@
+import { PageMetaResolverService } from '@spryker-oryx/core';
+import { resolve } from '@spryker-oryx/di';
```
dunqan: we should inject INJECTOR and use it to resolve tokens dulynamically

libs/template/presets/storefront/experience/pages/category-page.ts
Sinucid: dropped from current implementation, will be added in the following sub-task

libs/domain/site/src/services/breadcrumbs/breadcrumbs.service.ts
Sinucid: decided to leave it as is

libs/platform/router/src/services/router.service.ts
Sinucid: refactored

libs/domain/cart/src/services/resolver/cart.resolver.ts
```diff
@@ -14,11 +14,15 @@ export type CartResolvers = {
 };
 
 export class CartResolver extends BaseResolver<CartResolvers> {
-  protected cartService$ = resolve(CartService);
+  protected cartService: CartService;
+  constructor(protected injector = inject(INJECTOR)) {
```
supproduction:
```suggestion
  constructor(protected cartService = inject(CartService)) {
```

libs/domain/product/src/services/resolvers/breadcrumbs.resolver.ts
```diff
@@ -0,0 +1,37 @@
...
+import { RouteType, RouterService } from '@spryker-oryx/router';
+import {
+  Breadcrumb,
+  BreadcrumbsResolver,
+  BreadcrumbsResolvers,
+} from '@spryker-oryx/site';
+import { Observable, map } from 'rxjs';
+
+export class ProductListBreadcrumbsResolver implements BreadcrumbsResolver {
+  constructor(protected injector = inject(INJECTOR)) {}
```
supproduction:
```suggestion
  constructor(protected injector = inject(RouterService)) {}
```

libs/domain/search/src/services/resolvers/breadcrumbs.resolver.ts
```diff
@@ -0,0 +1,79 @@
...
+  Observable,
+  combineLatest,
+  map,
+  of,
+  switchMap,
+  throwError,
+} from 'rxjs';
+import { FacetListService } from '../facet-list.service';
+
+export type FlatFacet = { name: string; value: string };
```
supproduction: it it internal type lt's not export it
```suggestion
type FlatFacet = { name: string; value: string };
```

libs/domain/search/src/services/resolvers/breadcrumbs.resolver.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export class CategoryBreadcrumbsResolver implements BreadcrumbsResolver {
+  constructor(protected injector = inject(INJECTOR)) {}
+
+  protected getFlattenCategoriesBreadcrumbs(
+    categories: Facet
+  ): Observable<Breadcrumb[]> {
+    const getSelected = (facets?: FacetValue[]): FacetValue | undefined => {
+      return facets?.find(
+        ({ selected, children }) => selected || getSelected(children)
```
supproduction, line 29: Can we avoid recursive function?

libs/domain/search/src/services/resolvers/breadcrumbs.resolver.ts
```diff
@@ -0,0 +1,79 @@
...
+    };
+
+    const flattenSelected = (facet?: FacetValue): FlatFacet[] => {
+      if (!facet) {
+        return [];
+      }
+      const { selected, name, value, children } = facet;
+      return [
+        { name: name ?? '', value: String(value) },
+        ...(!selected ? flattenSelected(children?.[0]) : []),
```
supproduction, line 40: Can we avoid recursive function?

libs/domain/search/src/services/resolvers/breadcrumbs.resolver.ts
```diff
@@ -0,0 +1,79 @@
...
+      ];
+    };
+
+    const selected = getSelected(categories.values as FacetValue[]);
+    const flatten = flattenSelected(selected);
+    return flatten.length
+      ? combineLatest(
+          flattenSelected(selected).map(({ name, value }) =>
```
supproduction:
```suggestion
          flatten.map(({ name, value }) =>
```

libs/domain/search/src/services/resolvers/breadcrumbs.resolver.ts
```diff
@@ -0,0 +1,79 @@
...
+  of,
+  switchMap,
+  throwError,
+} from 'rxjs';
+import { FacetListService } from '../facet-list.service';
+
+export type FlatFacet = { name: string; value: string };
+
+export class CategoryBreadcrumbsResolver implements BreadcrumbsResolver {
+  constructor(protected injector = inject(INJECTOR)) {}
```
supproduction:
```suggestion
  constructor(protected injector = inject(LinkService)) {}
```

libs/domain/search/src/services/resolvers/breadcrumbs.resolver.ts
```diff
@@ -0,0 +1,79 @@
...
+      : of([]);
+  }
+
+  resolve(): Observable<Breadcrumb[]> {
+    return this.injector
+      .inject(FacetListService)
+      .getFacet({ parameter: 'category' })
+      .pipe(
+        switchMap((facet) => this.getFlattenCategoriesBreadcrumbs(facet)),
+        switchMap((breadcrumbs) =>
```
supproduction:
```suggestion
        map((breadcrumbs) =>
```

libs/domain/site/src/services/breadcrumbs/default-breadcrumbs.service.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Observable, map, switchMap } from 'rxjs';
+import { Breadcrumb } from '../../models';
+import {
+  BreadcrumbsResolver,
+  BreadcrumbsResolvers,
+  BreadcrumbsService,
+} from './breadcrumbs.service';
+
+export class DefaultBreadcrumbsService implements BreadcrumbsService {
+  constructor(protected injector = inject(INJECTOR)) {}
```
supproduction:
```suggestion
  constructor(protected injector = inject(RouterService)) {}
```

libs/domain/site/src/services/breadcrumbs/default-breadcrumbs.service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  protected getResolverKey(type: string): string {
+    return `${BreadcrumbsResolvers}${type}`;
+  }
+
+  protected resolveType(): Observable<string | undefined> {
+    return this.injector
+      .inject(RouterService)
+      .current()
+      .pipe(map((route) => route.type as string | undefined));
```
supproduction: do we need typecasting here?

libs/domain/site/src/services/breadcrumbs/default-breadcrumbs.service.ts
```diff
@@ -0,0 +1,55 @@
...
+    return [this.homeBreadcrumb, ...breadcrumbs];
+  }
+
+  protected resolveBreadcrumbs(type = 'DEFAULT'): Observable<Breadcrumb[]> {
+    const key = this.getResolverKey(type);
+    return this.injector
+      .inject<BreadcrumbsResolver>(key)
+      .resolve()
+      .pipe(map((breadcrumbs) => this.mapBreadcrumbs(breadcrumbs)));
+  }
```
supproduction:
```suggestion
  protected resolveBreadcrumbs(type = 'DEFAULT'): Observable<Breadcrumb[]> {
    const key = this.getResolverKey(type);
    return this.injector
      .inject<BreadcrumbsResolver>(key)
      .resolve()
      .pipe(map((breadcrumbs) => [this.homeBreadcrumb, ...breadcrumbs]));
  }
```

libs/domain/site/src/services/breadcrumbs/default-breadcrumbs.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  get(): Observable<Breadcrumb[]> {
+    return this.resolveType().pipe(
+      switchMap((type) => {
+        try {
+          if (!type) throw 'Incorrect route type!';
+          return this.resolveBreadcrumbs(type.toUpperCase());
+        } catch {
+          //Fallback breadcrumbs
+          return this.resolveBreadcrumbs();
+        }
```
supproduction:
```suggestion
 if (!type) throwError 'Incorrect route type!';
        return this.resolveBreadcrumbs(type?.toUpperCase());
```

libs/domain/site/src/services/resolvers/breadcrumbs.resolver.ts
```diff
@@ -0,0 +1,21 @@
+import { PageMetaResolverService } from '@spryker-oryx/core';
+import { INJECTOR, Provider, inject } from '@spryker-oryx/di';
+import { Observable, map } from 'rxjs';
+import { Breadcrumb } from '../../models';
+import { BreadcrumbsResolver, BreadcrumbsResolvers } from '../breadcrumbs';
+
+export class DefaultBreadcrumbsResolver implements BreadcrumbsResolver {
+  constructor(protected injector = inject(INJECTOR)) {}
```
supproduction:
```suggestion
  constructor(protected injector = inject(PageMetaResolverService)) {}
```

libs/domain/user/src/services/resolver/user.resolver.ts
```diff
@@ -14,7 +14,11 @@ export type UserResolvers = {
 };
 
 export class UserResolver extends BaseResolver<UserResolvers> {
-  protected user$ = resolve(UserService).getUser();
+  constructor(protected injector = inject(INJECTOR)) {
```
supproduction:
```suggestion
  constructor(protected injector = inject(UserService)) {
```

libs/platform/router/src/services/default-router.service.spec.ts
```diff
@@ -1,9 +1,16 @@
 import { StorageService, StorageType } from '@spryker-oryx/core';
 import { createInjector, destroyInjector } from '@spryker-oryx/di';
 import { of } from 'rxjs';
+import 'urlpattern-polyfill';
+import { RouteConfig } from '../../lit/lit-router';
```
supproduction:
```suggestion
import { RouteConfig } from '@spryker-oryx/router/lit';
```

libs/platform/router/src/services/default-router.service.ts
```diff
@@ -21,6 +22,16 @@ import {
 const CURRENT_PAGE = 'currentPage';
 const PREVIOUS_PAGE = 'previousPage';
 
+const isPatternConfig = (route: RouteConfig): route is URLPatternRouteConfig =>
+  (route as URLPatternRouteConfig).pattern !== undefined;
+
+const getPattern = (route: RouteConfig) => {
```
supproduction, line 28: Just add `export` in `lit-router` don't do copy paste

libs/platform/router/src/services/router.service.ts
```diff
@@ -61,3 +62,8 @@ export interface RouteParams {
 }
 
 export type QueryParamsHandling = 'merge' | '';
+
+export type RouteWithParams = RouteConfig & {
```
supproduction, line 66:
```suggestion
export interface RouteWithParams extends RouteConfig {
```

supproduction (changes requested):

libs/domain/search/src/services/resolvers/breadcrumbs.resolver.ts
Sinucid: Unfortunately, this won't work due to `getFlattenCategoriesBreadcrumbs` returns observable that is array of breadcrumbs that build their links by the means of `linkService`.

libs/domain/site/src/services/breadcrumbs/default-breadcrumbs.service.ts
Sinucid: route type has special type `RouteType` those values are strings. Unfortunately we need to cast the type

libs/domain/site/src/services/breadcrumbs/default-breadcrumbs.service.ts
Sinucid: Won't work. Throw will break the execution of the method and breadcrumbs won't be resolved. The idea of `try - catch` - to have a fallback for any cases, for example when page does not have it's own breadcrumbs resolver, then injector throws an error and fallback will be used (home \ page title)

libs/platform/router/src/services/default-router.service.ts
Sinucid: It is not just a copy-paste, the implementation in service was simplified. We discussed that with Stan, and decided not to touch the router (manage exports, change descriptors) due to router is temporary solution and can be changed or even dropped in the future

libs/platform/router/src/services/router.service.ts
Sinucid: Is not possible, due to `RouteConfig = PathRouteConfig | URLPatternRouteConfig`. Interface can extend statically known members only

libs/domain/search/src/services/resolvers/breadcrumbs.resolver.ts
Sinucid: I asked **Denys** about existing limits of categories nesting and he answered that this was not limited. Due to that category facets have various nesting I don't see another way how to handle this without recursion. Any suggestion?

supproduction (approved):

libs/domain/product/src/services/resolvers/breadcrumbs.resolver.ts
```diff
@@ -0,0 +1,34 @@
...
+        },
+      ])
+    );
+  }
+}
+
+export const ProductListBreadcrumbs: Provider = {
+  provide: `${BreadcrumbsResolvers}${String(
+    RouteType.ProductList
+  ).toUpperCase()}`,
```
dunqan: do we need this `toUpperCase()` as it's const anyway?

libs/domain/site/src/models/breadcrumbs.model.ts
```diff
@@ -0,0 +1,5 @@
+export interface Breadcrumb {
+  text?: string;
```
dunqan, line 2: What about having generic text resource, that can either be a string or object with token?

libs/domain/site/src/services/breadcrumbs/breadcrumbs.service.ts
dunqan: wby? :) we can aways improve typesafety of injection tokens in the future with this mapping

libs/domain/site/src/services/resolvers/breadcrumbs.resolver.ts
```diff
@@ -0,0 +1,18 @@
...
+export class DefaultBreadcrumbsResolver implements BreadcrumbsResolver {
+  constructor(protected pageMetaService = inject(PageMetaResolverService)) {}
+
+  resolve(): Observable<Breadcrumb[]> {
+    return this.pageMetaService.getTitle().pipe(map((text) => [{ text }]));
+  }
+}
+
+export const DefaultBreadcrumbs: Provider = {
+  provide: `${BreadcrumbsResolvers}DEFAULT`,
```
dunqan: can we have one const for this one?

libs/domain/product/src/services/resolvers/breadcrumbs.resolver.ts
Sinucid: Yes, upperCase is required to make service's name more consistent, and I would like to connect route type with resolvers type and do not use the hardcoded strings

libs/domain/site/src/models/breadcrumbs.model.ts
Sinucid: dunno, lets ask Tobi when he is back

libs/domain/site/src/services/resolvers/breadcrumbs.resolver.ts
Sinucid: lets implement it in the following subtask

dunqan (approved):

PR #708, opened 2023-08-14
feat(user): implement header menu account link component

aleespryker:
closes: [HRZ-89563](https://spryker.atlassian.net/browse/HRZ-89563)

[HRZ-89563]: https://spryker.atlassian.net/browse/HRZ-89563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/user/account-link/account-link.component.ts
```diff
@@ -0,0 +1,30 @@
+import { AuthService } from '@spryker-oryx/auth';
+import { resolve } from '@spryker-oryx/di';
+import { ButtonType } from '@spryker-oryx/ui/button';
+import { IconTypes } from '@spryker-oryx/ui/icon';
+import { hydrate, I18nMixin, signal } from '@spryker-oryx/utilities';
+import { html, LitElement, TemplateResult } from 'lit';
+import { styles } from './account-link.styles';
+
+@hydrate({ event: 'window:load' })
+export class AccountLinkComponent extends I18nMixin(LitElement) {
```
Sinucid: Can we use just content-link instead of this specific component?

aleespryker: I gave it a try. Unfortunately, the content link button style which is closest in appearance gives me no way to get rid of the box shadow that appears on hover.

Without the button style, the icon is the wrong size.

Should we make the box shadow color css configurable?

libs/domain/content/link/link.component.ts
```diff
@@ -22,10 +31,15 @@ export class ContentLinkComponent extends ContentMixin<
   });
 
   protected override render(): TemplateResult | void {
-    const { button, icon, singleLine, color } = this.$options();
+    const { appearance, button, icon, singleLine, color } = this.$options();
+
+    this.dropdown = appearance === LinkAppearance.DROPDOWN;
```
Sinucid: super bad idea to change reactive properties inside render

libs/base/ui/action/link/src/link.model.ts
```diff
@@ -23,3 +23,8 @@ export const enum ColorType {
   Primary = 'primary',
   Neutral = 'neutral',
 }
+
+export const enum LinkAppearance {
```
Sinucid: Why do you need this enum? And why is it related to the ui link?

The initial idea was to add a new option for the content link that will make it look like menu item. It has already had a `button` option, that makes it look like just a button, that is completely strange. I would suggest to add a new option -> `appearance` that will have with possible values:

```
export const enum ContentLinkAppearance {
  Link = 'link',
  Button = 'button',
  MenuLink = 'menuLink',
}
```

ContentLinkAppearance.Link will be the default value for the option, and in the future we will deprecate the `button` option

@dunqan What do you think?

libs/domain/content/link/link.component.ts
```diff
@@ -1,17 +1,26 @@
...
 
 @hydrate()
 export class ContentLinkComponent extends ContentMixin<
   ContentLinkOptions,
   ContentLinkContent
 >(LitElement) {
+  static styles = styles;
+
+  @property({ reflect: true, type: Boolean })
+  protected dropdown = false;
```
Sinucid: Why do you need a separate property if you have `appearance` option?

libs/domain/content/link/link.styles.ts
```diff
@@ -0,0 +1,13 @@
+import { css } from 'lit';
+
+export const styles = css`
+  :host([dropdown]) {
+    --oryx-button-color: var(--oryx-color-neutral-12);
+
+    padding: 9px 15px;
+  }
+
+  :host([dropdown]:hover) {
```
Sinucid: If you need just an attribute for stylisation, try to toggle it inside `elementEffect`, or composition supports hover query, check the **home-page.ts** for some examples

Sinucid: Or we can just apply that stylisation directly to the button, ie:
```
oryx-button[type="text"] {
...
}
```

Sinucid (changes requested):

libs/base/ui/action/link/src/link.model.ts
aleespryker: I tried putting it in content library originally but then it couldn't be used in the auth `login-link` component because content library is in `domain`, and I get warnings that I'm not allowed to use it unless it's in `base` or `platform`.

Otherwise I can just add the extra button option to fill in for the option we plan to deprecate.

libs/domain/content/link/link.component.ts
aleespryker: It was for styling, but I see there's a comment below that addresses it.

libs/base/ui/action/link/src/link.model.ts
aleespryker: We will use `oryx-option` in `login-link` to avoid this problem and keep `ContentLinkAppearance` in the content package.

libs/domain/content/link/link.model.ts
```diff
@@ -5,6 +5,12 @@ export interface ContentLinkContent {
   text?: string;
 }
 
+export const enum ContentLinkAppearance {
+  LINK = 'link',
```
tolerants: Please follow our conventions about enums.

libs/domain/content/link/link.component.ts
```diff
@@ -1,19 +1,42 @@
...
+  static styles = styles;
+
   protected semanticLinkService = resolve(LinkService);
 
+  @elementEffect()
+  protected setDropdown = () => {
+    const { appearance } = this.$options();
+    this.toggleAttribute(
+      'dropdown',
+      appearance === ContentLinkAppearance.DROPDOWN
```
tolerants: Maybe it makes sense to have an `appearance` as an attribute as long as we need it for styling? To not have this additional attribute.

libs/domain/content/link/link.model.ts
```diff
@@ -50,4 +56,9 @@ export interface ContentLinkOptions {
    * Indicates that the link text should truncate when it has multiple lines.
    */
   singleLine?: boolean;
+
+  /**
+   * Changes the appearance of the content link
```
tolerants:
```suggestion
   * Changes the appearance of the content link.
```

libs/platform/auth/login-link/login-link.styles.ts
```diff
@@ -1,10 +1,7 @@
 import { css } from 'lit';
 
 export const styles = css`
-  oryx-button {
-    --oryx-icon-size: 24px;
-
-    justify-content: flex-start;
-    padding: 9px 15px;
+  oryx-content-link {
```
tolerants: Why we are writing styles for `oryx-content-link`?

tolerants (changes requested):

libs/domain/content/link/link.model.ts
```diff
@@ -5,6 +5,12 @@ export interface ContentLinkContent {
   text?: string;
 }
 
+export const enum ContentLinkAppearance {
+  Link = 'link',
+  Dropdown = 'dropdown',
```
tobi-or-not-tobi: I'd prefer to try to make this `Navigation`, so we can use it in consistently in all navigations, such as the category navigation and the in-page my account navigation. It would mean that the style rules are generic enough. If we cannot do it like this, we can stick with Dropdown. 
The problem with making it `dropdown` is that we cannot easily move an navigation from one composition to another without changing the appearance on the components. We'd also not be able to reuse it. 

We'll also tackle this with UX to see if we can make those navigations fairly consistent from style rules perspective.

libs/template/presets/storefront/experience/header.ts
```diff
@@ -105,7 +107,24 @@ export const HeaderTemplate: ExperienceComponent = {
...
+                        url: '/account/overview',
+                        appearance: ContentLinkAppearance.Dropdown,
+                      },
+                      content: {
+                        data: {
+                          text: i18n('user.account'),
+                        },
+                      },
+                    }
+                  : {},
```
tobi-or-not-tobi: Can we avoid adding an empty object here? I'd rather keep the component list as-is and only add a component when the featureVersion is used.

libs/domain/content/link/link.component.ts
```diff
@@ -22,10 +36,15 @@ export class ContentLinkComponent extends ContentMixin<
...
+
+    const dropdown = appearance === ContentLinkAppearance.Dropdown;
 
-    if (button) {
-      return html`<oryx-button>${this.renderLink(true)}</oryx-button>`;
+    if (button || dropdown || appearance === ContentLinkAppearance.Button) {
+      return html`<oryx-button
+        .type=${ifDefined(dropdown ? ButtonType.Text : undefined)}
+        >${this.renderLink(true)}</oryx-button
+      >`;
```
tobi-or-not-tobi: Our initial designs started to use buttons a lot, but in reality those are links.
We like the UI/UX to start to use native links and we already agreed with Helen on this approach. 

Can we already start using `<oryx-link>` here, and apply the navigation styles to it?

libs/domain/content/link/link.styles.ts
```diff
@@ -0,0 +1,15 @@
...
+    --oryx-button-color: var(--oryx-color-neutral-12);
+
+    padding: 9px 5px;
+    display: flex;
+    height: 24px;
+  }
+
+  :host([appearance='dropdown']:hover) {
+    background-color: var(--oryx-color-neutral-3);
+  }
```
tobi-or-not-tobi: We like those styles to be applied to the link component.
We like to try to make them generic enough so they can serve more navigation items in other places (see other comments).

libs/domain/content/link/link.styles.ts
```diff
@@ -0,0 +1,15 @@
+import { css } from 'lit';
+
+export const styles = css`
+  :host([appearance='dropdown']) {
+    --oryx-button-color: var(--oryx-color-neutral-12);
+
+    padding: 9px 5px;
+    display: flex;
+    height: 24px;
```
tobi-or-not-tobi: we should use the max available width, so that you can access/hover anywhere on the real estate of the component.

tobi-or-not-tobi (changes requested):

libs/domain/site/dropdown-item/dropdown-item.component.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  it('should be defined', () => {
+    expect(element).toBeInstanceOf(DropdownItemComponent);
+  });
+
+  it('passes the a11y audit', async () => {
+    await expect(element).shadowDom.to.be.accessible();
+  });
+
+  it('should render components', () => {
+    expect(element.renderRoot.querySelector('oryx-button')).not.toBeNull();
```
Sinucid:
```suggestion
    expect(element).toContainElement('oryx-button');
```

libs/domain/site/dropdown-item/dropdown-item.component.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    beforeEach(async () => {
+      element = await fixture(
+        html`<oryx-site-dropdown-item
+          .content=${{ text: 'mock' }}
+        ></oryx-site-dropdown-item>`
+      );
+    });
+
+    it('should render text', () => {
+      expect(element).toContainElement('span');
```
Sinucid: The test below is doing the same check,` innerText === 'mock'` means the element exists

libs/domain/site/dropdown-item/dropdown-item.component.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+          .options=${{ url: 'mock' }}
+        ></oryx-site-dropdown-item>`
+      );
+    });
+
+    it('should render anchor link', () => {
+      expect(element).toContainElement('a[href]');
+      expect(
+        (element.renderRoot.querySelector('a[href]') as HTMLAnchorElement).href
+      ).toContain('mock');
```
Sinucid:
```suggestion
      expect(element).toContainElement('a[href="mock"]');
```

libs/domain/site/dropdown-item/dropdown-item.component.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+        html`<oryx-site-dropdown-item
+          .options=${{ icon: IconTypes.Add }}
+        ></oryx-site-dropdown-item>`
+      );
+    });
+    it('should render icon', () => {
+      expect(element).toContainElement('oryx-icon');
+      expect(
+        element.renderRoot.querySelector('oryx-icon')?.getAttribute('type')
+      ).toBe(IconTypes.Add);
```
Sinucid:
```suggestion
      expect(element).toContainElement(`oryx-icon[type="${IconTypes.Add}"]`);
```

libs/domain/site/dropdown-item/dropdown-item.styles.ts
```diff
@@ -0,0 +1,29 @@
+import { css } from 'lit';
+
+export const styles = css`
+  oryx-button {
+    display: flex;
+    --oryx-button-height: auto;
+    --oryx-button-border: none;
+    --oryx-button-padding: 0;
+    --oryx-button-color: var(--oryx-color-neutral-12);
+    --oryx-icon-color: var(--oryx-color-neutral-9);
```
Sinucid:
```suggestion
    --oryx-button-height: auto;
    --oryx-button-border: none;
    --oryx-button-padding: 0;
    --oryx-button-color: var(--oryx-color-neutral-12);
    --oryx-icon-color: var(--oryx-color-neutral-9);
    
    display: flex;
```

libs/domain/site/dropdown-item/stories/demo.stories.ts
```diff
@@ -0,0 +1,36 @@
+import { getAppIcons } from '@/tools/storybook';
+import { Meta, Story } from '@storybook/web-components';
+import { TemplateResult, html } from 'lit';
+import { storybookPrefix } from '../../.constants';
+import {
+  DropdownItemContent,
+  DropdownItemOptions,
+} from '../dropdown-item.model';
+
+const icons = getAppIcons();
```
Sinucid, line 10: component has hover state, better to ignore this story for chromatic and add static one, with default and hovered state

libs/platform/auth/login-link/login-link.component.ts
```diff
@@ -31,13 +27,12 @@ export class LoginLinkComponent extends ContentMixin<LoginLinkOptions>(
     const i18nToken = this.$isAuthenticated() ? 'auth.logout' : 'auth.login';
 
     return html`
-      <oryx-button
-        .type=${ButtonType.Text}
-        .text=${this.i18n(i18nToken)}
-        .icon=${IconTypes.Login}
+      <oryx-site-dropdown-item
```
Sinucid: Looks like a BC. Content element was changed that potentially can break custom styles. I would suggest to render an `oryx-button` for the < 1.1, and  `oryx-site-dropdown-item` for >= 1.1. Don't forgot to restore the styles and make them also conditional

Sinucid (changes requested):

libs/domain/site/dropdown-item/dropdown-item.component.ts
```diff
@@ -0,0 +1,36 @@
...
+import { ButtonType } from '@spryker-oryx/ui/button';
+import { LitElement, TemplateResult, html } from 'lit';
+import { when } from 'lit/directives/when.js';
+import {
+  DropdownItemContent,
+  DropdownItemOptions,
+} from './dropdown-item.model';
+import { styles } from './dropdown-item.styles';
+
+export class DropdownItemComponent extends ContentMixin<
```
tolerants, line 11: I think we agreed with Helen it's a `MenuItemComponent`. @aleespryker can you double check with her?

tolerants, line 13: I'm wondering why it's in the site domain and can we make it a general component inside ui?

tolerants (changes requested):

libs/domain/site/dropdown-item/dropdown-item.component.ts
Sinucid: UI components are dummy, but this one uses content mixin under the hood (with content service) and can be used in EB compositions. This is definitely domain component

libs/base/ui/action/dropdown-item/dropdown-item.component.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Icons } from '@spryker-oryx/ui/icon';
+import { hydrate } from '@spryker-oryx/utilities';
+import { LitElement, TemplateResult, html } from 'lit';
+import { property } from 'lit/decorators.js';
+import { when } from 'lit/directives/when.js';
+import { DropdownItemComponentAttributes } from './dropdown-item.model';
+import { styles } from './dropdown-item.styles';
+
+@hydrate()
+export class DropdownItemComponent
```
Sinucid, line 11: Not sure about moving it to the UI. What is the aim of this component then? It is just an oryx-button with a bit of additional styles. Will we use it directly in EB compositions?

Sinucid (changes requested):

libs/domain/site/dropdown-item/dropdown-item.component.ts
tolerants: It looks like a UI component to me, but details of the implementation should be clarified by @tobi-or-not-tobi together with Helen.

libs/base/ui/action/dropdown-item/dropdown-item.component.ts
aleespryker: I was trying to follow the suggestion above https://github.com/spryker/oryx/pull/708#discussion_r1325740926

and after discovering it wouldn't work because of a circular dependency with `ContentMixin`, we decided to just split the component like this. Would it have been better to keep it under the site package?

The only other minor difference is that we can use the ui dropdown-item component directly in login link instead of the site version (minor difference I guess).

PR #712, opened 2023-08-15
fix(auth): OauthService wait for initial restoreState

aleespryker:
closes: [HRZ-89566](https://spryker.atlassian.net/browse/HRZ-89566)

[HRZ-89566]: https://spryker.atlassian.net/browse/HRZ-89566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/auth/src/oauth/oauth.service.ts
```diff
@@ -192,11 +203,20 @@ export class OauthService implements AuthService, AuthTokenService {
     return this.state$;
   }
 
-  protected restoreState(): void {
-    this.storageService
+  protected initState(): Observable<OauthServiceState | null> {
```
tolerants: Shoudn't we update tests?

tolerants (changes requested):

libs/platform/auth/src/oauth/oauth.service.ts
```diff
@@ -192,11 +203,20 @@ export class OauthService implements AuthService, AuthTokenService {
...
+      .pipe(
+        distinctUntilChanged(),
+        map((state) => {
+          this.state$.next({ ...state });
+          return state;
+        })
+      );
+  }
+
+  protected restoreState(): void {
```
nsilva-spryker: Do we still need it and why?

nsilva-spryker (changes requested):

libs/platform/auth/src/oauth/oauth.service.ts
aleespryker: It's still being used by `invokeStoredToken`, but I can move the implementation of `restoreState` there so we won't need it anymore.

nsilva-spryker: yes, let's do this small refactor and move implementation to `invokeStoredToken`

nsilva-spryker (dismissed):

tarashaliuk (approved):

tolerants (approved):

nsilva-spryker (approved):

PR #714, opened 2023-08-16
test: added category page e2e tests

Ivan-Vidiaiev:
- added category page e2e tests
- added one category e2e test to smoke suite

closes: [HRZ-89598](https://spryker.atlassian.net/browse/HRZ-89598)

[HRZ-89598]: https://spryker.atlassian.net/browse/HRZ-89598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -161,6 +161,7 @@
     "abort-controller": "^3.0.0",
     "body-parser": "^1.20.2",
     "chai-a11y-axe": "^1.4.0",
+    "chai-sorted": "^0.2.0",
```
supproduction: where do we use it?

apps/storefront-e2e/src/integration/category.cy.ts
```diff
@@ -0,0 +1,124 @@
...
+    categoryPage = new CategoryPage({ id: '15' });
+    cy.visit(categoryPage.url);
+  });
+
+  it('default sorting applied even after clearing the sort parameter', () => {
+    cy.intercept('GET', '**/catalog-search?*').as('catalog-search');
+    cy.wait('@catalog-search', { timeout: 5000 });
+    //Wait till JS build the template
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(200);
```
supproduction, line 80: Where we get 200 ms from?

supproduction (changes requested):

apps/storefront-e2e/src/integration/category.cy.ts
```diff
@@ -0,0 +1,124 @@
...
+  let categoryPage;
+  beforeEach(() => {
+    categoryPage = new CategoryPage({ id: '15' });
+    cy.visit(categoryPage.url);
+  });
+
+  it('default sorting applied even after clearing the sort parameter', () => {
+    cy.intercept('GET', '**/catalog-search?*').as('catalog-search');
+    cy.wait('@catalog-search', { timeout: 5000 });
+    //Wait till JS build the template
```
tolerants: Isn't is more readable? :)
```suggestion
    // Wait till JS build the template
```

apps/storefront-e2e/src/integration/category.cy.ts
tolerants: Yeah we were talking with Andrii about all these delays to be stored in one place... so it have to refactored at some point.

Sinucid: It has a delay between data is populated from the request and template is completely built. It is small, but not equal 0. We decided to add 200ms cy.wait to be sure that list of the products is rendered

apps/storefront-e2e/src/support/page-objects/category.page.ts
```diff
@@ -18,6 +18,10 @@ export class CategoryPage extends AbstractSFPage {
   }
 
   getFacets = () => cy.get('oryx-search-facet');
-  getOryxCards = () => cy.get('oryx-product-card');
+  getRadio = () => cy.get('input[type="radio"]');
+  getOryxCard = () => cy.get('oryx-product-card');
```
tolerants: Why not:
```suggestion
  getProductCard = () => cy.get('oryx-product-card');
```

tolerants (changes requested):

package.json
Ivan-Vidiaiev: Not using it, deleted

apps/storefront-e2e/src/integration/category.cy.ts
supproduction: Yeah lets move it in separate variable with understandable naming

supproduction (approved):

package-lock.json
supproduction, line 1: Please, revert

supproduction (changes requested):

apps/storefront-e2e/src/integration/category.cy.ts
Ivan-Vidiaiev: But can we make it a separate PR or task? It would be good if we can enable these tests from now, as it also includes the Smoke test as well and we will do a refactor anyway later, as we will need to cover pagination here as well.

tolerants (approved):

supproduction (approved):

PR #722, opened 2023-08-23
feat(presets): breadcrumbs storefront integration and e2es

Sinucid:
- breadcrumbs component added to the search, category and product pages with dynamic visibility rules to hide it on small screens
- added e2e tests coverage for appropriate pages

closes: [HRZ-2318](https://spryker.atlassian.net/browse/HRZ-2318), [HRZ-89604](https://spryker.atlassian.net/browse/HRZ-89604)

[HRZ-2318]: https://spryker.atlassian.net/browse/HRZ-2318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-89604]: https://spryker.atlassian.net/browse/HRZ-89604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/presets/storefront/experience/pages/search-page.ts
```diff
@@ -17,6 +18,18 @@ export const searchPage: ExperienceComponent = {
     ],
   },
   components: [
+    {
+      type: 'oryx-site-breadcrumbs',
+      id: 'breadcrumbs',
```
tobi-or-not-tobi: can we leave out the `id`?

tobi-or-not-tobi (dismissed):
I have missed the previous PRs due to holidays, but have a few comments on this work:
- can we rename to breadcrumb? (singular)
- the last part of the breadcrumb should have an `inert` attribute to ensure it's not accessible
- we need to use standard font based icons for storefront. Let's use `chevron_right` or `navigate_next` for the delimiter. 
- I'd drop the `showDivider` and allow to have a nullish `divider` option that steers the usage of divider icon in the first place. 
- can we get an optional icon per breadcrumb? So that we can have the home icon to start the breadcrumb with, it's a common practice. Other pages could also add an icon in between, but that's not so common. Making it part of the breadcrumb parts, ensures it's clickable.
- Can we get the product category/ies in the pdp breadcrumb? Ideally we track the categories from the URL or history, but if not provided, we should rely on the product data. We have categories integrated in product data nowadays. 

We can address this feedback in separate subtask(s).

apps/storefront-e2e/src/support/page-fragments/breadcrumbs.fragment.ts
```diff
@@ -0,0 +1,15 @@
+export class BreadcrumbsFragment {
+  get = () => cy.get('oryx-site-breadcrumbs');
+  lastBreadcrumb = () => this.get().find('a:last-of-type');
+  nthBreadcrumb = (n: number) => this.get().find(`a:nth-of-type(${n})`);
```
supproduction: can we use nthBreadcrumb(-1) for last element?
`this.get().find('a').eq(number) `

apps/storefront-e2e/src/support/page-fragments/breadcrumbs.fragment.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  shouldHaveHomePageLink(): void {
+    const link = this.get().find('a:first-of-type');
```
supproduction:
```suggestion
    const link = this.nthBreadcrumb('1');
```

apps/storefront-e2e/src/support/page-fragments/breadcrumbs.fragment.ts
```diff
@@ -0,0 +1,15 @@
...
+    link.should('contain.text', 'Home');
+    link.invoke('attr', 'href').should('eq', '/');
```
supproduction:
```suggestion
    link.should('contain.text', 'Home').invoke('attr', 'href').should('eq', '/');
```

libs/domain/product/src/services/routes.ts
```diff
@@ -17,3 +17,10 @@ export const productRoutes: RouteConfig[] = [
         ),
   },
 ];
+
+export const productListRoutes: RouteConfig[] = [
+  {
+    path: '/search',
+    type: RouteType.ProductList,
+  },
+];
```
supproduction: just add it to the product routes

supproduction (changes requested):

supproduction (approved):

PR #725, opened 2023-08-25
feat: build time version flag

dunqan:
Introduce build time feature level flag, to accomodate breaking changes

Closes [HRZ-89610](https://spryker.atlassian.net/browse/HRZ-89610)

[HRZ-89610]: https://spryker.atlassian.net/browse/HRZ-89610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/src/routes.ts
```diff
@@ -52,14 +50,19 @@ export const defaultExperienceRoutes: RouteConfig[] = [
...
   },
-  {
-    path: '/product/:sku',
-    type: RouteType.Product,
-  },
-  {
-    path: '/category/:id',
-    type: RouteType.Category,
-  },
+  ...(OryxFeatureLevel < 110
```
tolerants: How to differentiate `1.10` and `11.0`?

tolerants (changes requested):

libs/platform/experience/src/routes.ts
tobi-or-not-tobi: what about `1_1_0`?

tobi-or-not-tobi: https://github.com/spryker/oryx/pull/725#discussion_r1305358958

dunqan: 👍 we decided to go with `1_1`

dunqan: we decided to sacrifice support for `1.10` for the time being.

apps/storefront/.env.template
```diff
@@ -30,3 +30,6 @@ ORYX_STORYBLOK_TOKEN=
 # Used to set a preset name, available options: b2b, b2c.
 # Default preset is b2c (storefront).
 ORYX_PRESET=
+
+# Used to set a feature level, available options: latest, 1_0, 1_1, etc.
```
tobi-or-not-tobi: can we add that only major and minor are needed?

apps/storefront/vite.config.common.ts
```diff
@@ -1,12 +1,18 @@
...
 
+dotenv.config();
+
 export const viteConfig = {
   index: './src',
   // vite starts paths from index dir
   root: '../',
   envPrefix: ['ORYX', 'FES', 'SCOS', 'STORE'],
+  define: {
+    __ORYX_FEATURE_LEVEL__: `${process.env.ORYX_FEATURE_LEVEL ?? 'latest'}`,
```
tobi-or-not-tobi: Wonder if latest is a good fallback. It means that if customers do not use this, they will get breaking changes. I'd rather have it more conservative.

apps/storefront/.env.template
```diff
@@ -30,3 +30,6 @@ ORYX_STORYBLOK_TOKEN=
...
+ORYX_FEATURE_LEVEL=
```
tolerants: Maybe `ORYX_FEATURE_VERSION`?

tobi-or-not-tobi (changes requested):

apps/storefront/.env.template
tobi-or-not-tobi: `ORYX_FEATURE_VERSION` doesn't sound right, as this is about opt-in to the latest minor.

apps/storefront/vite.config.common.ts
dunqan: 👍

apps/storefront/.env.template
dunqan: 👍

tobi-or-not-tobi (approved):

tolerants (approved):

PR #726, opened 2023-08-25
feat(cart): separate cart heading component

aleespryker:
The separation of the cart heading and the cart entries improves the reusability of both elements and will help to add more features to the cart heading going forward. Consequently, we introduced a checkout and order heading. 

The heading components are considered new features that will official land in 2.0, but can already be used and applied automatically when the `ORYX_FEATURE_VERSION` (min. `1.2`) is used.

closes: [HRZ-3262](https://spryker.atlassian.net/browse/HRZ-3244)

[HRZ-3262]: https://spryker.atlassian.net/browse/HRZ-3262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/cart/heading/index.ts
```diff
@@ -0,0 +1 @@
+export * from './heading.component';
```
supproduction, line 1: do we want to export styles and def?

libs/domain/order/heading/index.ts
```diff
@@ -0,0 +1 @@
+export * from './heading.component';
```
supproduction, line 1: same for styles and def

libs/template/presets/storefront/app/1_1/experience/experience-data.ts
```diff
@@ -0,0 +1,37 @@
...
+  editAddressPage,
+  homePage,
+  loginPage,
+  productPage,
+  searchPage,
+} from '../../1_0/experience/pages';
+import { cartPage, checkoutPage, orderConfirmationPage } from './pages';
+
+export const staticExperienceLatestFeature: AppFeature = {
+  providers: [
```
supproduction: Can we just use featureVersion instead of copy paste templates?

supproduction (changes requested):

libs/domain/order/heading/index.ts
tobi-or-not-tobi: there are no styles and we don't want to export def

libs/domain/cart/heading/index.ts
tobi-or-not-tobi: yes to styles, no to def (as it breaks lazy loading)

libs/template/presets/storefront/app/1_1/experience/experience-data.ts
tobi-or-not-tobi: actually, big left over, as I dropped the full copy/paste templates as part of the revamp of this one ;) So we think the same direction, just forgot to drop this stuff.

libs/domain/checkout/heading/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './heading.component';
+export * from './heading.def';
```
supproduction: Drop def here, please

supproduction (changes requested):

supproduction (approved):

PR #731, opened 2023-08-28
fix(order): clear guest order session storage on login

aleespryker:
closes: [HRZ-89571](https://spryker.atlassian.net/browse/HRZ-89571)

[HRZ-89571]: https://spryker.atlassian.net/browse/HRZ-89571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/order/src/services/order.routes.ts
```diff
@@ -1,24 +1,9 @@
...
 import { RouteConfig } from '@spryker-oryx/router/lit';
-import { combineLatest, map, take } from 'rxjs';
-import { OrderService } from './order.service';
 
 export const orderRoutes: RouteConfig[] = [
   {
     path: '/order/:id',
     type: RouteType.Order,
-    enter: ({ id }) =>
-      combineLatest([
```
supproduction, line 12: Why did we drop it?
In this case we don't have gurad for incorrect order

libs/domain/order/src/services/order.plugin.ts
```diff
@@ -0,0 +1,30 @@
+import { AuthService } from '@spryker-oryx/auth';
+import { App, AppPlugin } from '@spryker-oryx/core';
+import { resolve } from '@spryker-oryx/di';
+import { Subscription, tap } from 'rxjs';
+import { OrderService } from './order.service';
+
```
supproduction: Why do we need this plugin. Can we just clear storage on login?

supproduction (changes requested):

libs/domain/order/src/services/order.routes.ts
aleespryker: It doesn't work in SSR because it needs session data from the client side, so we always end up showing 404 pages except directly after checkout.

Discussed it with Stan, and he said to remove it for now. Feel free to ask him more about it.

libs/domain/order/src/services/order.plugin.ts
aleespryker: I wanted to keep it in the order package instead of the auth one.

supproduction: but you can check state auth in the order service. Not sure why we need plugin

libs/domain/order/src/controllers/order.controller.ts
```diff
@@ -24,10 +25,20 @@ export class OrderController {
...
+            ),
+            tap((lastOrder) => {
+              if (!lastOrder) {
+                this.orderService.clearLastOrder();
+              }
+            })
+          );
         }
+        this.orderService.clearLastOrder();
         return this.orderService.get({ id });
```
supproduction: we should do it in the service and here just use `get` method

libs/domain/order/src/mocks/src/mock-order.service.ts
```diff
@@ -14,4 +14,8 @@ export class MockOrderService implements Partial<OrderService> {
   getLastOrder(): Observable<OrderData | null> {
     return of(MockOrderService.mockOrder);
   }
+
+  clearLastOrder(): void {
+    //mock
+  }
```
supproduction: Leftover?

libs/domain/order/src/services/default-order.service.ts
```diff
@@ -37,7 +45,23 @@ export class DefaultOrderService implements OrderService {
 
   storeLastOrder(order: OrderData): void {
     // For privacy reasons, we cannot store the address in session storage.
-    const { billingAddress, shippingAddress, ...sanitized } = order;
-    this.storage.set(orderStorageKey, sanitized, StorageType.Session);
+    this.identity
```
supproduction: Can we get user data from order?

supproduction (changes requested):

libs/domain/order/src/mocks/src/mock-order.service.ts
aleespryker: We do actually need a clearLastOrder function in the mock or else the services/controllers using it in storybook and unit tests will not work.

libs/domain/order/src/services/default-order.service.ts
```diff
@@ -16,28 +24,55 @@ export class DefaultOrderService implements OrderService {
...
+    this.storage.set(
+      orderStorageKey,
+      { ...sanitized, userId },
+      StorageType.Session
+    );
+  }
+
+  clearLastOrder(): void {
+    this.storage.remove(orderStorageKey, StorageType.Session);
   }
```
supproduction: Ler's not create ne w public method. Change it to protected

libs/domain/order/src/services/default-order.service.ts
```diff
@@ -16,28 +24,55 @@ export class DefaultOrderService implements OrderService {
...
+          );
+        }
+        if (!this.orders$.has(id)) {
+          this.orders$.set(
+            id,
+            this.identity.get().pipe(
+              switchMap(() => this.adapter.get({ id })),
+              catchError(() => of(null)),
+              shareReplay({ bufferSize: 1, refCount: true })
+            )
```
supproduction:
```suggestion
            this.adapter.get({ id })).pipe(
              catchError(() => of(null)),
              shareReplay({ bufferSize: 1, refCount: true })
            )
```

libs/domain/order/src/services/default-order.service.ts
supproduction:
```suggestion
  protected clearLastOrder(): Observable<void> {
    return this.storage.remove(orderStorageKey, StorageType.Session);
  }
```

libs/domain/order/src/services/default-order.service.ts
```diff
@@ -16,28 +24,55 @@ export class DefaultOrderService implements OrderService {
...
+    return this.identity.get().pipe(
+      switchMap((user) => {
+        const { id } = data;
+        if (!user.isAuthenticated || !id) {
+          return this.getLastOrder().pipe(
+            map((lastOrder) =>
+              lastOrder?.userId === user.userId && lastOrder?.id === id
+                ? lastOrder
+                : null
+            ),
```
supproduction:
```suggestion
            switchMap((lastOrder) =>
              lastOrder?.userId === user.userId && lastOrder?.id === id
                ?of(lastOrder)
                : this.this.clearLastOrder()
            ),
```

libs/domain/order/src/services/default-order.service.ts
```diff
@@ -16,28 +24,55 @@ export class DefaultOrderService implements OrderService {
...
+            this.identity.get().pipe(
+              switchMap(() => this.adapter.get({ id })),
+              catchError(() => of(null)),
+              shareReplay({ bufferSize: 1, refCount: true })
+            )
+          );
+        }
+        this.clearLastOrder();
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        return this.orders$.get(id)!;
```
supproduction:
```suggestion
       return  this.clearLastOrder().pipe(switchMap() => this.orders$.get(id)!;
```

supproduction (changes requested):

supproduction (approved):

libs/domain/order/src/services/order.service.ts
```diff
@@ -3,9 +3,9 @@ import { OrderData } from '../models';
 import { GetOrderDataProps } from './adapter';
 
 export interface OrderService {
-  get(data: GetOrderDataProps): Observable<OrderData | null>;
+  get(data: GetOrderDataProps): Observable<OrderData | null | void>;
   getLastOrder(): Observable<OrderData | null>;
```
supproduction: Lett's change this method to protected

supproduction (changes requested):

libs/domain/order/src/controllers/order.controller.spec.ts
```diff
@@ -29,42 +28,21 @@ vi.spyOn(litRxjs, 'ObserveController') as SpyInstance;
...
-const mockUser: AuthIdentity = {
-  userId: 'userId',
-  isAuthenticated: true,
-};
-
-class MockIdentityService implements Partial<IdentityService> {
-  get = vi
-    .fn<[], Observable<AuthIdentity>>()
-    .mockReturnValue(of(mockAnonymousUser));
+  clearLastOrder = vi.fn();
```
supproduction: Do we need all this methods?

supproduction (dismissed):

libs/domain/order/src/controllers/order.controller.spec.ts
supproduction: clearLastOrder and getLastOrder

supproduction (approved):

PR #735, opened 2023-08-30
feat(site): breadcrumbs

Sinucid:
- new 'current' method was added to the router service that returns current route config with query and params
- added breadcrumb component to the site domain
- added breadcrumbs service and separate breadcrumbs resolvers to resolve product-details, categories, search and fallback breadcrumbs
- added category service and adapter to fetch and store categories data
- breadcrumbs components were added to the product, search, category pages
- sets of filed were limited for product and product-list requests
- optional 'parameter' field were added as qualifier for FacetQualifier

closes: [HRZ-2262](https://spryker.atlassian.net/browse/HRZ-2262)

[HRZ-2262]: https://spryker.atlassian.net/browse/HRZ-2262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tobi-or-not-tobi (approved):

libs/domain/cart/src/services/resolver/cart.resolver.ts
```diff
@@ -14,11 +14,13 @@ export type CartResolvers = {
 };
 
 export class CartResolver extends BaseResolver<CartResolvers> {
-  protected cartService$ = resolve(CartService);
+  constructor(protected cartService = inject(CartService)) {
+    super();
+  }
```
tobi-or-not-tobi: BC:
- existing protected `cartService$` no longer exists
- existing class extensions would fail since this class has a mandatory constructor

Fix:
- make cartService optional, keep `cartService$`, but deprecate it and support both during 1.0
- optionally, try to use feature flag to bring changes in

libs/domain/product/src/services/adapter/default-product.adapter.ts
```diff
@@ -29,11 +29,23 @@ export class DefaultProductAdapter implements ProductAdapter {
...
+
     return this.http
       .get<ApiProductModel.Response>(
         `${this.SCOS_BASE_URL}/${this.productEndpoint}/${sku}${
           include ? '?include=' : ''
-        }${include?.join(',') || ''}`
+        }${include?.join(',') || ''}
+        &fields[${
+          ApiProductModel.Includes.CategoryNodes
+        }]=${categoryNodeFields.join(',')}`
```
tobi-or-not-tobi: Potential BC on the consumption of this data. Let's assume this method was customized, and it would not return catergory nodes, would the normalizers break? Or still work?
To ensure a non BC DX, the code should ideally fail with a warning. 

We could consider to add the additional fields only when feature flag is on.

libs/domain/product/src/services/adapter/normalizers/concrete-products/concrete-products.normalizer.ts
```diff
@@ -27,7 +27,12 @@ export function concreteProductsNormalizer(
             ProductNormalizer
           ),
           transformer.transform(abstract[categoryKey], CategoryIdNormalizer),
-        ]).pipe(map(([product, nodeId]) => ({ ...product, ...nodeId })))
+        ]).pipe(
+          map(([product, nodeId]) => ({
+            ...product,
+            ...nodeId,
+          }))
+        )
```
tobi-or-not-tobi, line 35: Potential BC. Would the consuming code still work if a custom normalizer is used? Could we add this in a feature flag?

libs/domain/user/src/services/resolver/user.resolver.ts
```diff
@@ -14,7 +14,11 @@ export type UserResolvers = {
 };
 
 export class UserResolver extends BaseResolver<UserResolvers> {
-  protected user$ = resolve(UserService).getUser();
+  constructor(protected userService = inject(UserService)) {
+    super();
+  }
```
tobi-or-not-tobi: BC. What about overloading the constructor?

libs/platform/core/src/services/token-resolver/default-token-resolver.service.ts
```diff
@@ -12,6 +12,8 @@ const tokenRE = /^[A-Z_-]+\.(!?)[A-Z_-]+$/;
 export class DefaultTokenService implements TokenResolver {
   protected resolvers = new Map<string, TokenResourceResolver>();
 
+  constructor(protected injector = inject(INJECTOR)) {}
```
tobi-or-not-tobi: BC

tobi-or-not-tobi (changes requested):

libs/domain/product/src/services/adapter/normalizers/concrete-products/concrete-products.normalizer.ts
Sinucid: I didn't add there something, prettier just formatted it differently

libs/domain/cart/src/services/resolver/cart.resolver.ts
```diff
@@ -14,11 +15,20 @@ export type CartResolvers = {
...
+  /** @deprecated since 1.1 use cartService instead*/
   protected cartService$ = resolve(CartService);
+  protected cartService?: CartService;
+
+  constructor() {
+    super();
+    if (featureVersion >= '1.1') {
+      this.cartService = inject(CartService);
+    }
+  }
```
tobi-or-not-tobi:
```suggestion
  protected cartService = inject(CartService);
```

libs/domain/cart/src/services/resolver/cart.resolver.ts
```diff
@@ -14,11 +15,20 @@ export type CartResolvers = {
...
 
   protected resolvers = {
     SUMMARY: (): ResolvedToken => {
-      return this.cartService$.getCart().pipe(
+      return (this.cartService ?? this.cartService$).getCart().pipe(
```
tobi-or-not-tobi:
```suggestion
      return ((featureVersion >= '1.1' ? this.cartService : this.cartService$)).getCart().pipe(
```

libs/domain/cart/src/services/resolver/cart.resolver.ts
```diff
@@ -35,7 +45,7 @@ export class CartResolver extends BaseResolver<CartResolvers> {
       );
     },
     EMPTY: (): ResolvedToken => {
-      return this.cartService$
+      return (this.cartService ?? this.cartService$)
```
tobi-or-not-tobi:
```suggestion
      return (featureVersion >= '1.1' ? this.cartService : this.cartService$)
```

libs/domain/cart/src/services/resolver/cart.resolver.ts
tobi-or-not-tobi: This is causing:
- optional cartService field, but we rather make it required in future version
- constructor without arguments, that might be fine, but normally you'd add a constructor arg

Therefor we like: 
- add new field, using inject outside constructor. Ensure the new field is not optional
- apply conditional logic in the consuming code (see below comments).

tobi-or-not-tobi: We should use this pattern in other cases as well (e.g. `UserResolver`).

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #739, opened 2023-09-02
feat(site): pdp breadcrumb

Sinucid:
- added category service and adapter to fetch and store categories data
- added additional breadcrumbs resolver for PDP page

closes: [HRZ-89635](https://spryker.atlassian.net/browse/HRZ-89635)

[HRZ-89635]: https://spryker.atlassian.net/browse/HRZ-89635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/product/src/models/product.model.ts
```diff
@@ -31,6 +31,7 @@ export interface Product {
    * Ids of the category to which the product belongs.
    */
   categoryIds?: string[];
+  categories?: ProductCategory[];
```
tobi-or-not-tobi: how much is this different from the `categoryIds`? Should we not only main category id's and resolve other data from a category service instead? Categories will play a larger role going forward (e.g. category list component, category driven navigations, etc.).

That being said, we can consider keeping them with products, as it's closely related.

libs/domain/site/src/services/site.providers.ts
```diff
@@ -98,18 +97,14 @@ export const siteProviders: Provider[] = [
...
-      ]
-    : []),
+  {
+    provide: BreadcrumbService,
+    useClass: DefaultBreadcrumbService,
+  },
+  {
+    provide: FallbackBreadcrumbResolver,
+    useClass: DefaultFallbackBreadcrumbResolver,
+  },
```
tobi-or-not-tobi, line 107: cool, these services/resolvers just new features and will not break anything. only the usage will be considered a breaking change.

tobi-or-not-tobi (changes requested):

libs/domain/site/src/services/site.providers.ts
Sinucid: Yup, and currently resolvers are used by breadcrumbs service, and the service itself by the breadcrumb component

libs/domain/product/src/models/product.model.ts
Sinucid: I would say, if we introduce a new service (ProductCategoriesService), to keep only `categoryIds` field has much more sense. What would this new service be? Will it just store categories from `concrete-product` and `catalog-search` requests and with ability to get them by ids? Need more details, because I learned about this feature for the first time.

One note, wee need to integrate this service with facets as well

tobi-or-not-tobi (approved):

dunqan (approved):

dunqan (approved):

PR #740, opened 2023-09-03
feat: password policy validation

tarashaliuk:
Added validation rules to the password input component

closes: https://spryker.atlassian.net/browse/HRZ-89611

### DIFF
--- a/libs/base/ui/form/password/src/password-input.model.ts
+++ b/libs/base/ui/form/password/src/password-input.model.ts
@@ -31,25 +31,25 @@ export const enum PasswordVisibilityStrategy {
   Hover = 'hover',
 }
 
-export interface ValidationOptions {
+export interface PasswordValidationOptions {
   /**
-   * Minimum length the password must have
+   * Specifies the minimum length required for the password.
    */
   minLength?: number;
   /**
-   * Maximum allowable length for the password
+   * Specifies the maximum allowable length for the password.
    */
   maxLength?: number;
   /**
-   * Requires at least one uppercase letter in the password
+   * Specifies the minimum number of uppercase characters required in the password.
    */
-  requireUpperLetter?: boolean;
+  minUppercaseChars?: number;
   /**
-   * Requires at least one numeric character in the password
+   * Specifies the minimum number of numeric characters required in the password.
    */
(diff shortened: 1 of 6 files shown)


libs/base/ui/form/password/src/password-input.model.ts
```diff
@@ -30,3 +30,26 @@ export const enum PasswordVisibilityStrategy {
...
+  requireUpperLetter?: boolean;
+  /**
+   * Requires at least one numeric character in the password
+   */
+  requireNumber?: boolean;
+  /**
+   * Requires at least one special character in the password
+   */
+  requireSpecialChar?: boolean;
+}
```
tobi-or-not-tobi: What about:

```ts
export interface PasswordValidationOptions {
  /**
   * Specifies the minimum length required for the password.
   */
  minLength?: number;
  /**
   * Specifies the maximum allowable length for the password.
   */
  maxLength?: number;
  /**
   * Specifies the minimum number of uppercase characters required in the password.
   */
  minUppercaseChars?: number;
  /**
   * Specifies the minimum number of numeric characters required in the password.
   */
  minNumbers?: number;
  /**
   * Specifies the minimum number of special characters required in the password.
   */
  minSpecialChars?: number;
}
```

Makes it a little more flexible, we could default the values to `1`.

libs/base/ui/form/password/src/password-input.model.ts
```diff
@@ -30,3 +30,26 @@ export const enum PasswordVisibilityStrategy {
    */
   Hover = 'hover',
 }
+
+export interface ValidationOptions {
```
tobi-or-not-tobi: you probably want to use this in the component.

tobi-or-not-tobi (changes requested):

libs/base/ui/form/password/src/password-input.model.ts
tarashaliuk: Thanks, I like your idea about number types instead of boolean. But I think it is better to left props without default values because of flexibility (e.g. use just minLength rule). What do you think?

tobi-or-not-tobi (approved):

PR #744, opened 2023-09-04
fix(ui): give font icons width to ensure proper word wrapping in buttons

aleespryker:
closes: [HRZ-89670](https://spryker.atlassian.net/browse/HRZ-89670)

[HRZ-89670]: https://spryker.atlassian.net/browse/HRZ-89670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/base/ui/graphical/icon/src/icon.styles.ts
+++ b/libs/base/ui/graphical/icon/src/icon.styles.ts
@@ -9,8 +9,8 @@ export const styles = css`
     display: flex;
     align-items: center;
     justify-content: center;
-    width: var(--oryx-icon-size, var(--_size, 24px));
-    aspect-ratio: 1 / 1;
+    height: var(--oryx-icon-size, var(--_size, 24px));
+    width: max-content;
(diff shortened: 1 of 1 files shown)


libs/base/ui/graphical/icon/src/icon.styles.ts
```diff
@@ -10,6 +10,7 @@ export const styles = css`
     align-items: center;
     justify-content: center;
     height: var(--oryx-icon-size, var(--_size, 24px));
+    width: var(--oryx-icon-size, var(--_size, 24px));
```
tobi-or-not-tobi, line 13: we can probably drop `aspect-ratio` with this change.

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #748, opened 2023-09-06
feat: remove patch on cart when changing price mode

nsilva-spryker:
Requirements have change regarding price mode patch update. Right now **we will not patch cart with price mode**


libs/domain/cart/src/services/adapter/default-cart.adapter.ts
```diff
@@ -79,44 +78,6 @@ export class DefaultCartAdapter implements CartAdapter {
     );
   }
 
-  update(data: UpdateCartQualifier): Observable<Cart> {
```
supproduction: I would keep update method, just delete headers

libs/domain/cart/src/services/default-cart.service.ts
```diff
@@ -112,28 +111,6 @@ export class DefaultCartService implements CartService {
     },
   ]);
 
-  protected updatePriceMode$ = createEffect<Cart>([
```
supproduction, line 115: Let's create update method in service

supproduction (changes requested):

libs/domain/cart/src/services/cart.providers.ts
```diff
@@ -29,10 +27,6 @@ export const cartProviders: Provider[] = [
     provide: TotalsService,
     useClass: DefaultTotalsService,
   },
-  {
-    provide: PriceModeChangeGuard,
-    useClass: CartPriceModeChangeGuard,
-  },
```
supproduction, line 35: Also, please delete interceptor from public API? (means export)

supproduction (changes requested):

libs/domain/cart/src/services/default-cart.service.ts
```diff
@@ -112,27 +111,24 @@ export class DefaultCartService implements CartService {
...
-              });
-            })
-          )
-        );
-      }
+  update(data: Cart): Observable<Cart | undefined> {
+    if (data) {
+      subscribeReplay(
+        this.getCart().pipe(
+          take(1),
```
supproduction: Please, use similar flow as `addEntry...`

supproduction (changes requested):

libs/domain/cart/src/services/cart.providers.ts
nsilva-spryker: `export const PriceModeChangeGuard = 'oryx.PriceModeChangeGuard*'`
Was deleted from file `price-mode.service.ts`

Also CartVersionInterceptor deleted

supproduction (approved):

PR #751, opened 2023-09-08
feat: asynchronous providers

dunqan:
With this PR, we've added infrastructure for asynchronous loading of services, normalizers and serializers. This improves the initial loading of the application, as the implementations are only loaded when needed.  

The lazy loading technique is applied to the checkout and cart domains. Their services, adapters an converters are no longer part of the main javascript chunk. Only when the user navigates to the cart of checkout, additional javascript is loaded. 

Closes [HRZ-2592](https://spryker.atlassian.net/browse/HRZ-2592)

[HRZ-2592]: https://spryker.atlassian.net/browse/HRZ-2592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/checkout/src/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from './deprecated';
```
tolerants: Not sure about the name, as long as having a `deprecated` folder can be confusing to customers. Maybe we can rename it into `reexports` for example?

tolerants:
Shouldn't it be merged to a feature branch?

libs/base/di/src/injector.ts
```diff
@@ -127,21 +135,27 @@ export class Injector {
...
+      if (isAsyncClassProvider(providerInstance.provider)) {
+        providerInstance.instance = createLazyProxy(
+          providerInstance.provider.asyncClass,
+          this,
+          (instance) => (providerInstance.instance = instance)
+        );
+      } else if (isClassProvider(providerInstance.provider)) {
+        const restoreInjector = setCurrentInjector(this);
         providerInstance.instance = new providerInstance.provider.useClass();
+        restoreInjector();
```
tolerants, line 147: It looks like we are changing a logic of the method a bit. Are we considering it as a fix or a breaking change?

tolerants (changes requested):

libs/base/di/src/injector.ts
dunqan: the logic stays the same, we just do it now for each case separately (as we don't have to do it for asycClass)

libs/domain/checkout/src/index.ts
dunqan: customers shouldn't see it at all, as they should use npm libs :) but yeah, thinking about improving it to 'deprecated-reexports' maybe? 🤔

libs/domain/checkout/src/deprecated-reexports.ts
```diff
@@ -1,11 +1,11 @@
-import * as services from '@spryker-oryx/checkout/services';
 import { featureVersion } from '@spryker-oryx/utilities';
+import * as services from '../services/src/index';
 
 const reexports: typeof services =
   featureVersion < '1.1' ? services : (undefined as any);
 
 /** @deprecated since 1.1, use DefaultCartService from @spryker-oryx/cart/services */
```
tolerants: `use DefaultCartService from @spryker-oryx/cart/services` we should use Cart service instead of Checkout? Is it just a copy-paste?

tolerants: Can you also check the whole file, there are a lot of mentions of `@spryker-oryx/cart/services`?

libs/domain/checkout/src/index.ts
```diff
@@ -1,4 +1,5 @@
-export * from './deprecated';
+export * from './checkout.providers';
+export * from './deprecated-reexports';
```
tolerants: Maybe we can remove `deprecated`?

tolerants (changes requested):

libs/domain/checkout/src/deprecated-reexports.ts
dunqan: 👍  thanks, fixed

libs/domain/checkout/src/index.ts
dunqan: changed to `services-reexports` :)

tolerants (dismissed):

tolerants:
Shouldn't we merge it into the feature branch?

libs/base/di/src/dynamic-proxy/dynamic-proxy-instance.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { inject, Injector } from '@spryker-oryx/di'; // Replace with your actual import
```
tolerants: Is it a leftover? Shouldn't be a relative path?

libs/base/di/src/dynamic-proxy/dynamic-proxy-instance.ts
```diff
@@ -0,0 +1,71 @@
+import { Injector } from '@spryker-oryx/di';
```
tolerants: Shouldn't be a relative path?

libs/base/di/src/injector.ts
tolerants: But in previous implementation `restoreInjector();` was a default behaviour, and now it's not.

libs/domain/checkout/services/src/adapter/default-checkout.adapter.ts
```diff
@@ -1,16 +1,34 @@
 import { IdentityService } from '@spryker-oryx/auth';
 import { HttpService, JsonAPITransformerService } from '@spryker-oryx/core';
 import { inject } from '@spryker-oryx/di';
-import { combineLatest, map, Observable, switchMap, take } from 'rxjs';
+import { Observable, combineLatest, map, switchMap, take } from 'rxjs';
+
+/* deprecated since 1.1, we should use imports from @spryker-oryx/checkout:
 import {
```
tolerants, line 2: Not sure we need all the imports here. And there are a lot of places like this.

tolerants (dismissed):

libs/base/di/src/injector.ts
dunqan: we need to restore it, only when setCurrentInjector was called first (when we instantiate service), and in case of asnycClass, we don't need it (as we delay instantation). So the logic for other cases (where it's needed) behaves the same.

libs/domain/checkout/services/src/adapter/default-checkout.adapter.ts
dunqan: it's a reminder, we should fix it in the future (when dropping 1.0 support)

libs/domain/checkout/src/index.ts
dunqan: btw, updated, is it OK now?

libs/domain/checkout/services/src/adapter/default-checkout.adapter.ts
dunqan: actually I switched to the target ones, and checking if it not breaks anything

libs/domain/checkout/src/index.ts
tolerants: Yeah, works for me

libs/domain/checkout/services/src/adapter/default-checkout.adapter.ts
tolerants: But what about other places?

dunqan: done

tobi-or-not-tobi (approved):

PR #753, opened 2023-09-08
feat(user): user registration component

tarashaliuk:
Implemented user registration component and placed it on the registration page

closes: https://spryker.atlassian.net/browse/HRZ-89612


libs/domain/site/src/services/gender/default-gender.service.ts
```diff
@@ -0,0 +1,14 @@
+import { Observable, of } from 'rxjs';
+import { Option } from '../../models';
+import { GenderService } from './gender.service';
+
+export class DefaultGenderService implements GenderService {
+  protected genders = [
+    { text: 'male', value: 'Male' },
+    { text: 'female', value: 'Female' },
+  ];
```
tobi-or-not-tobi, line 9: Is there no Spryker API that exposes this list?

libs/domain/site/src/models/option.model.ts
```diff
@@ -1,4 +1,4 @@
-export interface Salutation {
+export interface Option {
```
tobi-or-not-tobi: this is considered a breaking change

libs/domain/user/registration/registration.component.ts
```diff
@@ -0,0 +1,217 @@
...
+              terms: {
+                value: 'terms and conditions',
+                link: {
+                  href: this.$options()?.termsAndConditionsLink || '',
+                  target: '_blank',
+                },
+              },
+            })}
+          </span>
+        </div>
```
tobi-or-not-tobi: we should add text to the checkbox component rather than adding separate span; the span text should be clickable as well. We should stil be able to click on the link though.

libs/domain/user/registration/registration.component.ts
```diff
@@ -0,0 +1,217 @@
...
+        this.hasGenericError,
+        () => html`
+          <oryx-notification type="error" scheme="dark">
+            ${this.i18n('user.registration.something-went-wrong')}.
+          </oryx-notification>
+        `
+      )}
+
+      <form @submit=${this.onSubmit}>
+        ${this.fieldRenderer.buildForm(this.getFields())}
```
tobi-or-not-tobi, line 139: you could actually register the password and terms and conditions as form field renderers. That would keep this class much cleaner, as it's a bit lengthy. 

It means that all form fields are data driven, and when password or t&c is rendered, your registered Form field renderer would be used.

libs/template/presets/storefront/experience/pages/registration-page.ts
```diff
@@ -0,0 +1,31 @@
+import { ExperienceComponent } from '@spryker-oryx/experience';
+
+export const registrationPage: ExperienceComponent = {
```
tobi-or-not-tobi, line 8: this is considered a breaking change:
- customer might already have their own at the same route
- a customer that would get this page, would not be prepared for this, leading to operation issues, broken integrations, misalignment of UI, etc.

Can we wrap it inside the latest feature flag?

tobi-or-not-tobi (changes requested):

libs/domain/site/src/models/option.model.ts
tarashaliuk: Should we avoid it and create the same Gender interface or just describe this BC in PR description?

tobi-or-not-tobi: If we want to move away from the current design, we should:
- keep Salutation as-is, but deprecate it
- introduce new Option
- ensure code can still work with Saluation, but use Option as the main model going forward

libs/domain/site/src/services/gender/default-gender.service.ts
tarashaliuk: No, there is no endpoint(s) that exposes the list of genders and salutations

libs/domain/user/registration/registration.component.ts
```diff
@@ -0,0 +1,260 @@
...
+      },
+      {
+        id: 'acceptedTerms',
+        type: FormFieldType.Boolean,
+        required: true,
+        label: this.i18n('user.registration.i-accept-the-<terms>', {
+          terms: {
+            value: 'terms and conditions',
+            link: {
+              href: this.$options()?.termsAndConditionsLink || '',
```
tobi-or-not-tobi, line 229: Let's align the link and text of the link with the t&a from the footer. 
we should probably add a default value for the option here.

libs/template/presets/storefront/experience/pages/registration-page.ts
```diff
@@ -0,0 +1,35 @@
...
+              ],
+              minLength: 8,
+              minUppercaseChars: 1,
+              minNumbers: 1,
+              minSpecialChars: 1,
+            },
+          },
+        ],
+      }
+    : {};
```
tobi-or-not-tobi: can we nog make it undefined?

libs/platform/form/src/renderers/default-form.renderer.ts
```diff
@@ -125,11 +125,52 @@ export class DefaultFormRenderer implements FormRenderer {
       case FormFieldType.RadioList: {
         return this.buildRadioList(field, value as string);
       }
+      case FormFieldType.Password: {
+        return this.buildPasswordField(field, value as string);
+      }
```
tobi-or-not-tobi, line 167: let's add some tests

tobi-or-not-tobi (changes requested):

libs/domain/site/src/models/salutation.model.ts
```diff
@@ -1,3 +1,6 @@
+/**
+ * @deprecated Use the `Option` interface instead.
```
tolerants: Please add a version of deprecation using a format described [here](https://spryker.atlassian.net/wiki/spaces/FA/pages/3870228702/Draft+Dealing+with+breaking+changes), it will help us later when we will clean up these deprecations.

libs/domain/site/src/models/option.model.ts
```diff
@@ -0,0 +1,4 @@
...
```
tolerants: Please add a description. Also I'm not sure what entity in site domain this interface represents. Isn't it better to move it to the core/utilities?

tolerants:
Suggest to align form a bit on mobile:
image

libs/domain/user/registration/index.ts
```diff
@@ -0,0 +1,4 @@
+// export * from './default-registration.strategy';
```
tolerants: Leftover?

libs/domain/user/registration/index.ts
```diff
@@ -0,0 +1,4 @@
...
+export * from './registration.component';
+export * from './registration.model';
+// export * from './registration.strategy';
```
tolerants: Leftover?

libs/domain/user/registration/registration.component.ts
```diff
@@ -0,0 +1,262 @@
...
+
+  protected setPasswordValidity(): void {
+    this.hasPasswordError = !this.passwordComponent?.isValid();
+  }
+
+  protected async registerUser(): Promise<void> {
+    this.isLoading = true;
+
+    try {
+      await firstValueFrom(
```
tolerants, line 116: I personally like this approach, probably worth to add as a good example of allowed usage of rxjs operators inside components.

libs/domain/user/registration/registration.component.ts
```diff
@@ -0,0 +1,262 @@
...
+        this.hasGenericError,
+        () => html`
+          <oryx-notification type="error" scheme="dark">
+            ${this.i18n('user.registration.something-went-wrong')}.
+          </oryx-notification>
+        `
+      )}
+
+      <form @submit=${this.onSubmit}>
+        <oryx-layout
```
tolerants: It looks like we have an unneeded space at the right side
image

libs/domain/user/registration/registration.component.ts
```diff
@@ -0,0 +1,262 @@
...
+      gender: this.gender?.value ?? '',
+      firstName: this.firstName?.value ?? '',
+      lastName: this.secondName?.value ?? '',
+      email: this.email?.value ?? '',
+      password: this.password?.value ?? '',
+      acceptedTerms: this.acceptedTerms?.checked,
+    };
+  }
+
+  constructor() {
```
tolerants: Can you please consider moving constructor to to the start of the Class definition? At the very bottom it looks a bit weird.

libs/template/presets/storefront/experience/pages/registration-page.ts
```diff
@@ -0,0 +1,35 @@
+import { ExperienceComponent } from '@spryker-oryx/experience';
+import { featureVersion } from '@spryker-oryx/utilities';
+
+export const registrationPage: ExperienceComponent | undefined =
+  featureVersion === 'latest'
```
tolerants:
```suggestion
  featureVersion >= '1.1'
```
It can be changed to `1.2`, depends on release schedule.

libs/template/presets/storefront/experience/pages/registration-page.ts
```diff
@@ -0,0 +1,35 @@
+import { ExperienceComponent } from '@spryker-oryx/experience';
```
tolerants: Shouldn't we also add a ling to this page somewhere? Or I missed it?

libs/platform/form/src/renderers/default-form.renderer.ts
```diff
@@ -191,14 +247,10 @@ export class DefaultFormRenderer implements FormRenderer {
...
       >
-        <input
-          type="checkbox"
-          .name=${field.id}
-          ?checked=${!!value}
-          ?required=${field.required}
-        />
-        ${field.label}
+        ${this.renderInput(field, value)}
+        <span>${field.label}</span>
```
tolerants: Can be a breaking change.

tolerants (changes requested):

libs/domain/user/registration/registration.component.ts
tarashaliuk: Where add it?

libs/domain/site/src/models/option.model.ts
tarashaliuk: Maybe this interface can be replaced with FormFieldOption?

libs/template/presets/storefront/experience/pages/registration-page.ts
tarashaliuk: Added the button on Login page. Also, aligned the Registration page with designs to make user able to move back to Login page

libs/platform/form/src/renderers/default-form.renderer.ts
tarashaliuk: Added in BC description

libs/template/presets/storefront/experience/pages/registration-page.ts
tobi-or-not-tobi: Please make sure you add the `>=` condition

libs/platform/form/src/renderers/default-form.renderer.ts
```diff
@@ -79,6 +79,52 @@ export class DefaultFormRenderer implements FormRenderer {
...
+      case FormFieldType.Color:
+        inputType = 'color';
+        break;
+      case FormFieldType.Boolean:
+        inputType = 'checkbox';
+        break;
+      case FormFieldType.Toggle:
+        inputType = 'radio';
+        break;
+    }
```
tobi-or-not-tobi: Seems wrong. We assign the filed.type already. 

What about this?

```ts
inputType = field.attributes?.type ?? field.type
```

we only should make an exception to the `toggle` as it requires to become `radio`.

libs/platform/auth/login/login.component.ts
```diff
@@ -87,76 +101,73 @@ export class AuthLoginComponent extends ContentMixin<LoginOptions>(LitElement) {
...
+        <h2>${this.i18n('user.login.new-customer')}</h2>
+      </oryx-heading>
+      <p>${this.i18n('user.login.benefits-of-registering-to-the-website')}</p>
+
+      <oryx-button
+        href=${this.$options()?.registrationLink}
+        type=${ButtonType.Outline}
+      >
+        ${this.i18n('user.login.create-account')}
+      </oryx-button>
```
tobi-or-not-tobi: We should not add this content to the login component. Instead, we like to have the following:
- use the `oryx-content-text` component for the title, text and button
- add the component to the login page, only add it for >= 1.1 version

This makes the content content manageable, as well as the layout. Some stores use the text next to the login, in a 50-50 layout.

libs/platform/auth/login/login.component.ts
```diff
@@ -75,7 +84,12 @@ export class AuthLoginComponent extends ContentMixin<LoginOptions>(LitElement) {
   }
 
   protected override render(): TemplateResult {
-    return html` ${when(
+    return html`
+      <oryx-heading as=${HeadingTag.H2}>
+        <h1>${this.i18n('user.login.log-in')}</h1>
+      </oryx-heading>
```
tobi-or-not-tobi: We could add the text outside the login component inside a text component, to avoid hard coding it.

libs/platform/auth/login/login.component.ts
```diff
@@ -87,76 +101,73 @@ export class AuthLoginComponent extends ContentMixin<LoginOptions>(LitElement) {
...
-          <input
-            type="password"
-            name="password"
-            required
-            placeholder=${this.i18n('login.password')}
-          />
-        </oryx-password-input>
-
-        ${this.renderLoginOptions()}
+        ${this.fieldRenderer.buildForm(this.getFields())}
```
tobi-or-not-tobi, line 99: ❤️

tobi-or-not-tobi (changes requested):

tolerants:
We are not allowed to make any breaking changes :) In case there is a breaking change we have to release `2.0` immediately.
So 

> deprecated Salutation interface. Use FormFieldOption interface instead

Deprecation is not a breaking change as long as we haven't removed an original code.

> added registration page

It's not a breaking change in case we added it under the feature version, so existing customers will not be affected by this change.

> wrapped checkbox text into span (default-form.renderer.ts) to be able insert elements without style issues 

And here we probably have to wrap new html into the feature flag to not produce a breaking change.

libs/platform/form/src/renderers/default-form.renderer.ts
tolerants: This is not how we are dealing with breaking changes :) Added a general comment https://github.com/spryker/oryx/pull/753#issuecomment-1723353836

libs/template/presets/storefront/experience/pages/login-page.ts
```diff
@@ -15,7 +16,25 @@ export const loginPage: ExperienceComponent = {
     {
       type: 'oryx-auth-login',
       options: {
-        rules: [{ width: '50%', margin: 'auto' }],
```
tolerants, line 18: I'd say we have to add new rules under the feature flag.

tolerants (changes requested):

libs/platform/form/src/renderers/default-form.renderer.ts
tarashaliuk: Set exception for toggle (radio) and boolean (checkbox)

libs/template/presets/storefront/experience/pages/login-page.ts
tarashaliuk: I have refactored the page structure - just moved this rule to the parent object

tolerants: Not sure if a get your point. We need to keep the same structure for `1.0`. And provide new refactored for 1.1 and later.

tolerants (changes requested):

libs/template/labs/src/i18n/translations/en.ts
```diff
@@ -35,7 +35,9 @@ const order = {
 
 const user = {
   'user.address.remove-info':
-    'Removing this address will not remove any pending orders being dispatched to this address',
+    'Removing this address will not remove any pending orders being dispatched to this` address',
+  'user.login.benefits-of-registering-to-the-website':
```
tolerants: But are we using this key anywhere?

tolerants (changes requested):

libs/template/labs/src/i18n/translations/en.ts
tarashaliuk: it is leftover from prev implementation. Removed it

tolerants (approved):

libs/template/presets/storefront/experience/pages/registration-page.ts
```diff
@@ -0,0 +1,129 @@
+import {
+  CompositionLayout,
+  ExperienceComponent,
+} from '@spryker-oryx/experience';
+import { ButtonSize, ButtonType } from '@spryker-oryx/ui/button';
+import { Size, featureVersion } from '@spryker-oryx/utilities';
+
+export const registrationPage: ExperienceComponent | undefined =
+  featureVersion >= '1.1'
```
tobi-or-not-tobi: What about have `registrationPage` without any versioning constraints, but only import it in `StaticExperienceFeature` with the version condition?

tobi-or-not-tobi (changes requested):

libs/platform/form/src/renderers/default-form.renderer.ts
tobi-or-not-tobi: But `field.attributes?.type` will be ignore in that case. I thought the specific attribute type should be preferred

libs/platform/auth/login/login.component.ts
```diff
@@ -87,76 +95,65 @@ export class AuthLoginComponent extends ContentMixin<LoginOptions>(LitElement) {
...
-          </oryx-checkbox>`
-        )}
         ${when(
           this.$options()?.forgotPasswordLink,
-          () => html`<oryx-link>
-            <a href=${this.$options()?.forgotPasswordLink}
-              >${this.i18n('user.login.forgot-password?')}</a
-            >
-          </oryx-link>`
+          () => html`
```
tobi-or-not-tobi, line 109: This part is very broken. We should probably go with an oryx-link here but let's double check the designs.

tobi-or-not-tobi (changes requested):

libs/platform/form/src/renderers/default-form.renderer.ts
tarashaliuk: Not sure that I got your POV. 
`field.attributes?.type` will be ignored only in the case of `boolean` and `toggle` values which should be mapped to "checkbox" and "radio" because those input types aren't accepted.

libs/platform/auth/login/login.component.ts
tarashaliuk: Yes, it should be oryx-link according to designs. Changed the oryx-button to oryx-link

libs/platform/form/src/renderers/default-form.renderer.ts
tobi-or-not-tobi: When you use `field.attributes?.type` you bring a very specific editor type into the mix. You could imagine that the type in the model is boolean, but you want have a custom editor that at then will resolve the boolean attribute. With this implementation we cannot achieve this, as the standard boolean editor is always used, regardless of a custom type.

tarashaliuk: Should the `toggle` not be mapped to `radio` as well? Because a custom editor can resolve `toggle`?

tobi-or-not-tobi: I was thinking of this:

```ts
 let inputType = field.type;

    switch (inputType) {
      case FormFieldType.Boolean:
        inputType = 'checkbox';
        break;
      case FormFieldType.Toggle:
        inputType = 'radio';
        break;
    }

    return html`
      <input
        name=${field.id}
        type=${field.attributes?.type ?? inputType}
```

tobi-or-not-tobi (approved):

PR #754, opened 2023-09-10
fix(search): search box on small screens

Sinucid:
Fix small screens experience of site search box component

closes: [HRZ-89724](https://spryker.atlassian.net/browse/HRZ-89724)

[HRZ-89724]: https://spryker.atlassian.net/browse/HRZ-89724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/search/searchbox/src/searchbox.controller.ts
```diff
@@ -186,8 +186,9 @@ export class SearchboxController implements ReactiveController {
   }
 
   protected onTriggerClick(): void {
-    const isOpened = this.host.hasAttribute('open');
-    this.host.toggleAttribute('open', !isOpened);
+    const isOpened = this.host.open; //this.host.hasAttribute('open');
+    this.host.open = !isOpened;
+    // this.host.toggleAttribute('open', !isOpened);
```
aleespryker: Do we still need these comments?

libs/base/ui/search/searchbox/src/searchbox.model.ts
```diff
@@ -1,9 +1,9 @@
 export interface SearchAttributes {
   open?: boolean;
   /**
-   * Provides floating behaviour on small screens
+   * Provides floating behavior
    */
-  xsFloated?: boolean;
+  float?: boolean;
```
tolerants, line 8: We are removing a property of public interface which is also a component attribute. I'd say it's a a breaking change. Can we think of backward compatible solution?

libs/base/ui/search/typeahead/src/typeahead.component.ts
```diff
@@ -30,6 +30,7 @@ export class TypeaheadComponent
 {
   static styles = baseStyles;
 
+  @property({ reflect: true, type: Boolean }) open?: boolean;
```
tolerants, line 33: Looks like a `feat`.

libs/base/ui/search/typeahead/src/typeahead.component.ts
```diff
@@ -46,6 +47,7 @@ export class TypeaheadComponent
   @property() clearIcon?: string;
   @property({ reflect: true }) clearIconPosition?: ClearIconPosition;
   @property() clearIconAppearance?: ClearIconAppearance;
+  @property({ type: Boolean, reflect: true }) float?: boolean;
```
tolerants, line 50: Looks like a `feat`.

libs/template/presets/storefront/experience/header.ts
```diff
@@ -13,17 +14,29 @@ export const HeaderTemplate: ExperienceComponent = {
         {
           type: 'oryx-content-link',
           content: { data: { text: 'FREE DELIVERY & RETURNS' } },
-          options: { url: '/', icon: IconTypes.Check },
+          options: {
```
tolerants, line 17: All these changes should be done using feature flags.

tolerants (changes requested):

libs/base/ui/search/searchbox/src/searchbox.styles.ts
```diff
@@ -2,114 +2,234 @@ import {
...
-    inset-inline-start: var(--floating-padding-start, 10px);
-    inset-inline-end: var(
-      --floating-padding-end,
-      var(--floating-padding-start, 10px)
-    );
-    width: calc(
-      100vw -
-        calc(
-          var(--floating-padding-start, 10px) +
+  featureVersion >= '1.1'
```
tobi-or-not-tobi: hard to read. maybe split in 2 const and pull them in based on feature version condition?

libs/base/ui/search/searchbox/src/searchbox.model.ts
```diff
@@ -1,9 +1,9 @@
 export interface SearchAttributes {
   open?: boolean;
   /**
-   * Provides floating behaviour on small screens
+   * Provides floating behavior
```
tobi-or-not-tobi:
```suggestion
   * Provides floating behaviour of the search input element, so that the search box leaves more space to other elements. This is especially when there's a lack of space, eg. on small screens.
```

tobi-or-not-tobi (dismissed):
left 2 nitpicks

libs/template/presets/storefront/experience/header.ts
tobi-or-not-tobi: as discussed we make the exception here because of this is considered a fix.

libs/base/ui/search/searchbox/src/searchbox.controller.ts
```diff
@@ -186,8 +186,8 @@ export class SearchboxController implements ReactiveController {
   }
 
   protected onTriggerClick(): void {
-    const isOpened = this.host.hasAttribute('open');
-    this.host.toggleAttribute('open', !isOpened);
+    const isOpened = this.host.open;
```
tolerants, line 189: But do we really need this variable?

libs/domain/search/box/box.model.ts
```diff
@@ -12,4 +12,6 @@ export interface SearchBoxOptions extends SuggestionRendererOptions {
    * Minimum amount of characters required for querying the suggestion
    */
   minChars?: number;
+
```
tolerants: Suggest to add a description.

tolerants (changes requested):

libs/base/ui/search/searchbox/src/searchbox.controller.ts
tobi-or-not-tobi: I'd say yes, since we have conditions based on the true/false value. implementation detail.

tolerants (approved):

PR #757, opened 2023-09-11
fix: fix showing home page while loading routes with guards

supproduction:
closes: [HRZ-89725](https://spryker.atlassian.net/browse/HRZ-89725)

[HRZ-89725]: https://spryker.atlassian.net/browse/HRZ-89725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/product/src/services/routes.ts
+++ b/libs/domain/product/src/services/routes.ts
@@ -1,19 +1,17 @@
-import { resolve } from '@spryker-oryx/di';
 import { RouteType } from '@spryker-oryx/router';
 import { RouteConfig } from '@spryker-oryx/router/lit';
-import { map, take } from 'rxjs';
-import { ProductService } from './product.service';
 
 export const productRoutes: RouteConfig[] = [
   {
     path: '/product/:sku',
     type: RouteType.Product,
-    enter: ({ sku }) =>
-      resolve(ProductService)
-        .get({ sku })
-        .pipe(
-          take(1),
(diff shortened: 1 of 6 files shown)


libs/domain/product/src/services/routes.ts
```diff
@@ -1,19 +1,17 @@
...
     path: '/product/:sku',
     type: RouteType.Product,
-    enter: ({ sku }) =>
-      resolve(ProductService)
-        .get({ sku })
-        .pipe(
-          take(1),
-          map((product) => (product ? true : RouteType.NotFound))
-        ),
+    // TODO: uncomment when performance issue will be solved
```
tolerants: My suggestion is to register a Jira ticket already to have an action item and mention this ticket here, so we have a straightforward connnection.

libs/platform/experience/src/routes.ts
```diff
@@ -7,34 +7,6 @@ import { html, TemplateResult } from 'lit';
 import { map, take } from 'rxjs';
 import 'urlpattern-polyfill';
 
-export const experienceRoutes: RouteConfig[] = [
```
tolerants, line 10: Should we still deprecate sku and id routes in 1.1?

tolerants (dismissed):

tolerants (approved):

PR #765, opened 2023-09-13
test(e2e): improve search e2e stability

Sinucid:
no ticket yet

### DIFF
--- a/apps/storefront-e2e/src/integration/search.cy.ts
+++ b/apps/storefront-e2e/src/integration/search.cy.ts
@@ -52,24 +52,19 @@ describe('Search suite', () => {
 
       // clear sorting and check that it is default again
       searchPage.getProductSorting().clearSorting();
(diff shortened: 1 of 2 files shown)


apps/storefront-e2e/src/integration/search.cy.ts
```diff
@@ -52,24 +52,19 @@ describe('Search suite', () => {
 
       // clear sorting and check that it is default again
       searchPage.getProductSorting().clearSorting();
-      searchPage.waitForSearchRequest();
```
AndreyYevtukhov, line 55: Why we are removing this wait? 

Without it we might have false positives here

AndreyYevtukhov (changes requested):
left 1 comment, let's double check if that removal is required (and why it worked before)

tolerants:
It's not a `fix`, but `test` type.

AndreyYevtukhov (approved):

PR #766, opened 2023-09-14
revert: Price mode

tolerants:


nsilva-spryker (dismissed):

.github/workflows/test.yml
```diff
@@ -17,7 +17,7 @@ concurrency:
 env:
   GITHUB_REF: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref || github.ref }}
   IS_DEV: ${{ github.event_name != 'pull_request' && github.ref == 'refs/heads/development' }}
-  ORYX_FEATURE_VERSION: latest
```
AndreyYevtukhov, line 20: @tolerants should we remove this? I see now that there are no Breadcrumbs on pages in CI (that's why tests fail)

AndreyYevtukhov (changes requested):

.github/workflows/test.yml
tolerants: It depends on what we want to test, but you're probably right as long as for now we want to test the latest version.

AndreyYevtukhov (approved):

PR #768, opened 2023-09-15
feat: category link component

supproduction:
adds category link component
adjusts revealer for EB

closes: [HRZ-89824](https://spryker.atlassian.net/browse/HRZ-89824)

[HRZ-89824]: https://spryker.atlassian.net/browse/HRZ-89824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/product/category-link/category-link.model.ts
+++ b/libs/domain/product/category-link/category-link.model.ts
@@ -1,0 +1,6 @@
+export interface ProductCategoryLinkOptions {
+  /**
+   * Category identifier.
+   */
+  category?: string;
+}
--- a/libs/platform/experience/src/services/data-client/data-client.model.ts
+++ b/libs/platform/experience/src/services/data-client/data-client.model.ts
@@ -8,16 +8,26 @@ export const enum MessageType {
   Options = 'oryx.options',
   Products = 'oryx.products',
   Query = 'oryx.query',
+  Category = 'oryx.Category',
   Static = 'oryx.static',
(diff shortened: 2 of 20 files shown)


libs/domain/product/category-link/category-link.model.ts
```diff
@@ -0,0 +1,6 @@
+export interface ProductCategoryLinkOptions {
+  /**
+   * Category identifier.
```
tolerants, line 3: Maybe it's worth to explain why it's optional and what is the behaviour when it's missing.

libs/platform/experience/src/services/data-client/data-client.model.ts
```diff
@@ -8,28 +8,41 @@ export const enum MessageType {
...
   ComponentType = 'oryx.component-type',
   ComponentSchemas = 'oryx.component-schemas',
   ColorMode = 'oryx.color-mode',
   AppReady = 'oryx.app-ready',
   Icons = 'oryx.icons',
+  Categories = 'oryx.categories',
 }
 
+/**
+ * @deprecated Since version 1.1. Will be removed.
```
tolerants, line 22: Can we suggest to use another interface as a full or partial replacement?

tolerants (changes requested):

PR #771, opened 2023-09-18
fix(ui): button states styles

Sinucid:
- states-related styles of oryx-button were refactored a bit
- fixed variants story of navigation-button component

closes: [HRZ-89875](https://spryker.atlassian.net/browse/HRZ-89875)

[HRZ-89875]: https://spryker.atlassian.net/browse/HRZ-89875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/site/navigation-button/stories/Static/common.ts
+++ b/libs/domain/site/navigation-button/stories/Static/common.ts
@@ -1,6 +1,7 @@
 import { generateVariantsMatrix, Variant } from '@/tools/storybook';
 import { html, TemplateResult } from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
+import { NavigationButtonComponent } from '../../navigation-button.component';
 
 enum CategoryX {
   DEFAULT = 'Default',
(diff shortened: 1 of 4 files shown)


libs/domain/site/navigation-button/stories/Static/common.ts
```diff
@@ -51,3 +52,31 @@ export const renderVariants = (): TemplateResult => html`
...
+  obj.canvasElement
+    .querySelectorAll<HTMLElement>('.pseudo-parent')
+    .forEach((parent) => {
+      const pseudo = parent.dataset.pseudo;
+      const button = parent.querySelector<NavigationButtonComponent>(
+        'oryx-site-navigation-button'
+      );
+
+      if (pseudo) {
+        setTimeout(() => {
```
supproduction, line 73: Why do we need setTimeout here?

supproduction (changes requested):

libs/domain/site/navigation-button/stories/Static/common.ts
Sinucid: Just to be sure that shadowRoot is ready. Components renders children asynchronously, and usually it's enough just to add setTimeout(0). Not the ideal solution, but suitable for storybook

supproduction (approved):

PR #775, opened 2023-09-19
feat(content): resolve product, category and content links

supproduction:
Improve the content link component to resolve links to product, category or content pages. 
Reveal suggestions so that integrations can reuse the data while building content. 

closes: [HRZ-89824](https://spryker.atlassian.net/browse/HRZ-89824)

[HRZ-89824]: https://spryker.atlassian.net/browse/HRZ-89824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/content/link/link.component.ts
+++ b/libs/domain/content/link/link.component.ts
@@ -30,7 +30,7 @@ export class ContentLinkComponent extends ContentMixin<
 
   protected $text = computed(() => {
     const { type, id } = this.$options();
-    const { text } = this.$content();
+    const { text } = this.$content() ?? {};
 
     if (text) return text;
 
--- a/libs/domain/content/link/link.component.ts
+++ b/libs/domain/content/link/link.component.ts
@@ -1,10 +1,13 @@
 import { resolve } from '@spryker-oryx/di';
 import { ContentMixin } from '@spryker-oryx/experience';
+import { ProductCategoryService, ProductService } from '@spryker-oryx/product';
+import { RouteType } from '@spryker-oryx/router';
 import { LinkService } from '@spryker-oryx/site';
 import { computed, hydrate } from '@spryker-oryx/utilities';
 import { LitElement, TemplateResult, html } from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { when } from 'lit/directives/when.js';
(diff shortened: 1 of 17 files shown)


libs/domain/content/link/link.component.ts
```diff
@@ -21,6 +26,27 @@ export class ContentLinkComponent extends ContentMixin<
     return null;
   });
 
+  protected $text = computed(() => {
+    const { type, id } = this.$options();
```
tobi-or-not-tobi, line 47: we could move this down a bit

libs/domain/content/link/link.schema.ts
```diff
@@ -1,75 +1,86 @@
...
+        {
+          value: RouteType.Category,
+          text: 'Category',
+        },
+        {
+          value: RouteType.ProductList,
+          text: 'Search',
+        },
+      ],
+    },
```
tobi-or-not-tobi: We could introduce a feature to make those _hidden_ in EB options panel, as for now they're autofilled by the suggestion widget. Or just remove them for now.

libs/domain/content/link/link.schema.ts
```diff
@@ -1,75 +1,86 @@
...
+export const linkComponentSchema: ContentComponentSchema<
+  ContentLinkComponent,
+  { suggestion: unknown }
+> = {
+  name: 'Link',
+  group: 'Content',
+  icon: IconTypes.Link,
+  options: {
+    suggestion: {
+      type: 'suggestion',
```
tobi-or-not-tobi: Suggestion sounds to generic. What about ContentSuggestion?

tobi-or-not-tobi (changes requested):

libs/domain/content/link/link.schema.ts
supproduction: Dropped it now

libs/domain/content/link/link.component.ts
tobi-or-not-tobi: nitpick, but still open afaik

tobi-or-not-tobi (approved):

PR #777, opened 2023-09-19
fix(checkout): center checkout shipping icon

aleespryker:
closes: [HRZ-89879](https://spryker.atlassian.net/browse/HRZ-89879)

[HRZ-89879]: https://spryker.atlassian.net/browse/HRZ-89879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/checkout/shipping-method/shipping-method.component.ts
+++ b/libs/domain/checkout/shipping-method/shipping-method.component.ts
@@ -39,7 +39,7 @@ export class CheckoutShippingMethodComponent
   protected $hasMethods = computed(() => {
     const carriers = this.shipments()?.[0]?.carriers;
 
-    return carriers?.find((carrier) => !!carrier.shipmentMethods?.length);
+    return !!carriers?.find((carrier) => !!carrier.shipmentMethods?.length);
   });
 
   @elementEffect()
@@ -48,9 +48,9 @@ export class CheckoutShippingMethodComponent
   };
 
   protected override render(): TemplateResult | void {
-    const carriers = this.shipments()?.[0]?.carriers;
-
     if (!this.$hasMethods()) return this.renderEmpty();
+
+    const carriers = this.shipments()?.[0]?.carriers;
 
     return html`
       <oryx-checkout-header>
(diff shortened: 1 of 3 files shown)


libs/domain/checkout/shipping-method/shipping-method.component.ts
```diff
@@ -28,18 +36,28 @@ export class CheckoutShippingMethodComponent
     return !!this.form?.checkValidity();
   }
 
+  protected $hasMethods = computed(() => {
+    const carriers = this.shipments()?.[0]?.carriers;
+
+    return carriers?.find((carrier) => !!carrier.shipmentMethods?.length);
```
Sinucid:
```suggestion
    return !!carriers?.find((carrier) => !!carrier.shipmentMethods?.length);
```

libs/domain/checkout/shipping-method/shipping-method.component.ts
```diff
@@ -28,18 +36,28 @@ export class CheckoutShippingMethodComponent
...
+  protected $toggleNoMethods = (): void => {
+    this.toggleAttribute('no-methods', !this.$hasMethods());
+  };
+
   protected override render(): TemplateResult | void {
     const carriers = this.shipments()?.[0]?.carriers;
 
-    if (!carriers?.find((carrier) => !!carrier.shipmentMethods?.length))
-      return this.renderEmpty();
+    if (!this.$hasMethods()) return this.renderEmpty();
```
Sinucid, line 51:
```
    if (!this.$hasMethods()) return this.renderEmpty();

    const carriers = this.shipments()?.[0]?.carriers;
```

libs/domain/checkout/shipping-method/shipping-method.styles.ts
```diff
@@ -18,10 +18,11 @@ export const styles = css`
     display: contents;
   }
 
-  .no-methods {
-    --oryx-icon-size: 40px;
```
Sinucid, line 22: Why did you drop the icon size?

Sinucid (changes requested):

libs/domain/checkout/shipping-method/shipping-method.styles.ts
aleespryker: Current icon on netlify appears to be the wrong size. Fixing it here, according to designs from the address list component story.

aleespryker: Nevermind, Helen says it should be 24px. The size is correct.

Sinucid (approved):

tolerants (approved):

PR #778, opened 2023-09-19
chore: update fes theme with icons

nsilva-spryker:
closes: [HRZ-89914](https://spryker.atlassian.net/browse/HRZ-89914)

[HRZ-89874]: https://spryker.atlassian.net/browse/HRZ-89874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HRZ-89914]: https://spryker.atlassian.net/browse/HRZ-89914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/themes/src/fes.theme.ts
```diff
@@ -8,6 +8,13 @@ export const fesTheme: Theme = {
...
   icons: {
-    resource: materialDesignIcons,
+    resource: fontawesomeIcons,
+    resources: [
+      {
+        resource: backofficeIcons,
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        types: Object.keys(backofficeIcons.mapping!),
+      },
+    ],
```
tobi-or-not-tobi: Can we keep the "fes" theme inside the fes repo? It's not a secret, but will make it easier to change things going forward. For example, we use currently fontawesome, but we do like to move to material icons

tobi-or-not-tobi: I'd also deprecate the fes theme here, so we can drop it with upcoming major.

tobi-or-not-tobi (changes requested):

libs/template/themes/src/fes.theme.ts
nsilva-spryker: fes theme updates inside fes repo: https://github.com/spryker/fes/pull/1969

libs/template/themes/src/fes.theme.ts
```diff
@@ -2,6 +2,7 @@ import { Theme } from '@spryker-oryx/experience';
 import { materialDesignIcons } from '@spryker-oryx/resources';
 import { defaultBreakpoints } from '@spryker-oryx/themes/breakpoints';
 
+/** @deprecated fes theme will be deprecated in upcoming major versions */
```
supproduction: Why you don't use jsDoc. Also you can check the message in the already created code. Alsp suggest what user can use instead

/**
 * @deprecated Since version `which one` Use backofficeTheme instead.
*/

supproduction (changes requested):

tobi-or-not-tobi (approved):

supproduction (approved):

PR #781, opened 2023-09-20
feat: create-oryx utility

dunqan:
Support for `npm create oryx`. 
Precompiled wrapper for Oryx CLI, allowing to easily setup an Oryx application. 

Closes [HRZ-86543](https://spryker.atlassian.net/browse/HRZ-86543)

[HRZ-86543]: https://spryker.atlassian.net/browse/HRZ-86543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/cli/package.json
```diff
@@ -10,14 +10,15 @@
   "bin": {
     "oryx": "./bin/cli.js"
   },
-  "dependencies": {
+  "devDependencies": {
```
tolerants: Why do you think these are dev dependencies?

libs/template/cli/src/commands/create.ts
```diff
@@ -79,6 +80,7 @@ Possible values: ${Object.values(OryxPreset).join(', ')}`
   }
 
   async createApp(options: CreateAppOptions = {}): Promise<void> {
+    console.log(``);
```
tolerants, line 68: Leftover?

libs/template/cli/src/commands/create.ts
```diff
@@ -123,77 +134,58 @@ Please make sure to not use an existing directory name.`
...
-        message: `What is the name of your app? ${c.dim(`[--name, -n]`)}`,
-        validate: (value) => {
-          if (!value.trim()) {
-            return 'Please enter a name';
-          }
-          return undefined;
-        },
+        message: `What is the name of your app?`,
+        placeholder: 'oryx-app',
+        defaultValue: 'oryx-app',
```
tolerants, line 163: Maybe `initialValue` will be better?

libs/template/cli/src/commands/create.ts
```diff
@@ -239,13 +231,22 @@ interface CreateAppConfig extends Required<CreateAppOptions> {
...
   B2C = 'b2c',
+  B2B = 'b2b',
   Fulfillment = 'fulfillment',
 }
 
 export enum OryxTemplateRef {
   Latest = 'latest',
 }
+
+export enum OryxOption {
```
tolerants, line 192: Probably we can remove it as long as we are not using it yet.

libs/template/cli/src/commands/create.ts
```diff
@@ -239,13 +231,22 @@ interface CreateAppConfig extends Required<CreateAppOptions> {
 export interface CreateAppOptions {
   name?: string;
   preset?: OryxPreset;
+  options?: OryxOption[];
 }
 
 export enum OryxPreset {
```
tolerants: Probably we can remove it as long as we are not using it yet.

tolerants (changes requested):

libs/template/cli/src/commands/create.ts
dunqan: nope, empty line

libs/template/cli/src/commands/create.ts
dunqan: initial you'll have to edit if you want to use different name, while default will be used if no name is provided, or you can just write your own

libs/template/cli/src/commands/create.ts
dunqan: we can, up to you :)

libs/template/cli/package.json
dunqan: they are not yet, but I think we should build cli bin to make it work without all the oryx deps, like `create-oryx`. Then they will be real devDeps.

tolerants:
As long as we are not going to release it, probably it's a `chore`.

tolerants (dismissed):

supproduction (approved):

PR #783, opened 2023-09-20
chore(presets): add breadcrumb to the contact page

Sinucid:
closes: [HRZ-89853](https://spryker.atlassian.net/browse/HRZ-89853)

[HRZ-89853]: https://spryker.atlassian.net/browse/HRZ-89853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tobi-or-not-tobi (approved):
few breaking changes, but it's ok. We're not in the religion bizz. No one will use this page. We probably should drop it going forward and replace it with CMS page.

PR #784, opened 2023-09-20
fix(presets): hide empty cart message on ssr loading

tarashaliuk:
Hide empty cart message on SSR loading. Implemented OR conditional syntax for hideByRule (e.g. `rule1 || rule2`). Added `CORE.SERVER` resolver.

closes: https://spryker.atlassian.net/browse/HRZ-3147


Sinucid (changes requested):
Nice idea, but check the e2e's please, seems to be this changes broke one of them. 
And don't forget to update PR's title, it should contain related domain, ie `fix(presets):`

libs/domain/cart/src/services/resolver/cart.resolver.ts
```diff
@@ -6,12 +6,13 @@ import {
...
 import { Provider, inject, resolve } from '@spryker-oryx/di';
 import { featureVersion } from '@spryker-oryx/utilities';
-import { map } from 'rxjs';
+import { map, of } from 'rxjs';
 import { CartService } from '..';
 
 export type CartResolvers = {
   SUMMARY: Resolver;
   EMPTY: Resolver;
+  TEST: Resolver;
```
Sinucid: Do you need this?

libs/platform/experience/composition/composition-components.controller.ts
```diff
@@ -74,11 +74,16 @@ export class CompositionComponentsController implements ReactiveController {
         if (rules.hide) {
           return of(null);
         }
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        return this.tokenResolver.resolveToken(rules.hideByRule!).pipe(
-          //hidden by default
-          startWith(true),
-          map((value) => (value ? null : component))
+
+        const hideRules =
```
Sinucid: I believe this new `OR` logic should be a part of DefaultTokenService

Sinucid (changes requested):

libs/platform/experience/composition/composition-components.controller.ts
tarashaliuk: Why it should be part of DefaultTokenService? This service's responsibility is resolving tokens, not rules. Rules could have different "shapes" that's why I think the operations with rules should proceed in the controller layer and then pass prepared tokens to TokenService to resolve it.

Sinucid: Because not only composition uses token resolver, if some other component or service requires similar logic of resolving, should we duplicate the code then? Or add additional operator like AND, need to update each component then?

I believe that TokenService should be able to accept any string, even with conditions. In result it's purpose to receive string and return resolved value

libs/platform/core/src/services/token-resolver/default-token-resolver.service.ts
```diff
@@ -9,34 +9,63 @@ import {
...
+
+  protected resolveConditionalToken(token: string): ResolvedToken {
+    const tokensToResolve = token
+      .split('||')
+      .map((token) => token.trim())
+      .filter(
+        (token) => this.isToken(token) && this.getResolver(token.split('.')[0])
+      );
+
+    if (!tokensToResolve.length) {
```
Sinucid, line 41: This is a flaky behaviour, I believe better to throw an error instead, to make it clear that tokens are invalid. Because if we return false for invalid tokens will not be so easy to debug it. Or at least console.error, for each invalid token

libs/platform/experience/composition/composition-components.controller.spec.ts
```diff
@@ -75,6 +78,22 @@ const mockComponentWithVisibilityRule: Component = {
   ],
 };
 
+const mockComponentWithVisibilityConditionalRule: Component = {
```
Sinucid: Leftover?

Sinucid (changes requested):

Sinucid (approved):

dunqan (approved):

PR #785, opened 2023-09-20
feat(user): sign up

tarashaliuk:
The sign up feature includes a registration page (`/registration`) and a section on the login page with a link to the registration page. The registration page allows you to create a new user for the application. 

This feature is considered a breaking change as the existing login page is affected as well as a new route and page is added to the structure. To use this feature already in the 1.x version, you can use the [feature version](https://docs.spryker.com/docs/scos/dev/front-end-development/202307.0/oryx/getting-started/oryx-versioning.html#feature-flag-versioning) environment variable (`ORYX_FEATURE_VERSION=1.1`).

closes: [HRZ-332](https://spryker.atlassian.net/browse/HRZ-332)

[HRZ-332]: https://spryker.atlassian.net/browse/HRZ-332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/form/password/src/password-input.styles.ts
```diff
@@ -30,6 +30,32 @@ export const baseStyles = css`
...
+    margin-block-end: 5px;
+  }
+
+  .validation-message oryx-icon {
+    height: 16px;
+    width: 20px;
+  }
+
+  .validation-message.active {
+    color: var(--oryx-color-primary-11);
```
tobi-or-not-tobi, line 57: The primary 11 vs grey doesn't work great. I've left a detailed note for Helen to get this sorted. I would not block the feature because of this.

tobi-or-not-tobi (approved):

PR #787, opened 2023-09-20
fix(order): display delivery costs in order page

aleespryker:
closes: [HRZ-89892](https://spryker.atlassian.net/browse/HRZ-89892)

[HRZ-89892]: https://spryker.atlassian.net/browse/HRZ-89892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/order/src/services/totals/order-totals.resolver.ts
+++ b/libs/domain/order/src/services/totals/order-totals.resolver.ts
@@ -50,10 +50,12 @@ export class OrderTotalsResolver implements TotalsResolver {
           currencyIsoCode: currency,
           calculatedDiscounts,
           priceMode,
+          shipments,
         } = order;
 
         return {
           ...totals,
+          shipmentTotal: shipments?.[0].defaultGrossPrice,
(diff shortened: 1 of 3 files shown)


libs/domain/order/src/services/totals/order-totals.resolver.ts
```diff
@@ -50,10 +50,12 @@ export class OrderTotalsResolver implements TotalsResolver {
           currencyIsoCode: currency,
           calculatedDiscounts,
           priceMode,
+          shipments,
         } = order;
 
         return {
           ...totals,
+          shipmentTotal: shipments?.[0].defaultGrossPrice,
```
supproduction: But if we in the Net mode?

supproduction (changes requested):

Sinucid (approved):

supproduction (approved):

PR #788, opened 2023-09-21
fix(content): blank screen instead of 404 page

aleespryker:
closes: [HRZ-89845](https://spryker.atlassian.net/browse/HRZ-89845)

[HRZ-89845]: https://spryker.atlassian.net/browse/HRZ-89845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/content/src/services/default-content.service.ts
+++ b/libs/domain/content/src/services/default-content.service.ts
@@ -10,7 +10,7 @@ export class DefaultContentService implements ContentService {
   protected contents: Record<string, string[]> = {};
 
   constructor(
-    protected adapters = inject(ContentAdapter, null),
+    protected adapters = inject(ContentAdapter, [] as ContentAdapter[]),
     protected config = inject(ContentConfig, [] as ContentConfig[]),
     protected injector = inject(INJECTOR)
   ) {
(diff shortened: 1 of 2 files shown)


libs/domain/content/src/services/default-content.service.ts
```diff
@@ -18,36 +18,44 @@ export class DefaultContentService implements ContentService {
...
+              adapter.get(q).pipe(catchError(() => of(null)))
+            )
+          ).pipe(
+            map((contents) =>
+              contents.reduce(
+                (acc, curr) => (curr ? { ...acc, ...curr } : acc),
+                null
+              )
+            )
           )
```
supproduction:
```suggestion
    loader: (q: ContentQualifier) => 
      this.getAdapters(q).length
        ? combineLatest(
            adapters.map((adapter) =>
              adapter.get(q).pipe(catchError(() => of(null)))
            )
          ).pipe(
            map((contents) =>
              contents.reduce(
                (acc, curr) => (curr ? { ...acc, ...curr } : acc),
                null
              )
            )
          )
```

supproduction (changes requested):

supproduction (approved):

PR #790, opened 2023-09-22
feat: fulfillment app boilerplate improvements

tolerants:
- [x] Optimized FA boilerlate
- [x] Drop unused files in FA boilerlate
- [x] Renamed `StopEventsInitializer` into `CaptureEventsInitializer`
- [x] Changed `oryx-app` rendering, so now it's not rendering header and footer.
- [x] Added header and footer into experience pages. 
- [x] Improvements from https://github.com/spryker/oryx/pull/837
closes: [HRZ-89640](https://spryker.atlassian.net/browse/HRZ-89640), [HRZ-89641](https://spryker.atlassian.net/browse/HRZ-89641), [HRZ-89553](https://spryker.atlassian.net/browse/HRZ-89553)

[HRZ-89640]: https://spryker.atlassian.net/browse/HRZ-89640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-89641]: https://spryker.atlassian.net/browse/HRZ-89641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-89553]: https://spryker.atlassian.net/browse/HRZ-89553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tobi-or-not-tobi (changes requested):
There are a few important things related to the layout rendering I like to change in this PR. They're all related to the `type` we introduced to `OryxAppComponent`. I believe that's not the right thing to do, as we have everything we need in Oryx to manage layout for various screen sizes already. 

Here are the pieces:
- FA theme should only supports 1 breakpoint, we should add the small screen breakpoint rather than reuse the `defaultBreakpoints`
- The FA theme should not add all breakpoints for all sizes, but stick to sm only. Most likely it's cleaner to create an fa design token file, or bring them directly into the theme instead
- There's a bug in the currently used `backofficeNgTokens` that is not applying the container width correctly, as it's been wrapped in `layout` field.
- In order to leverage the small screen only, and apply the centered layout because of that, we should ideally leverage experience pages rather than connect components directly to the routes. If not easily possible, we could reuse layout styles in the various components to leverage the layout.

supproduction (dismissed):

libs/template/presets/storefront/experience/pages/registration-page.ts
```diff
@@ -58,69 +49,94 @@ export const registrationPage: ExperienceComponent = {
...
-            },
-            {
-              type: 'oryx-content-text',
+              type: 'oryx-composition',
               options: {
                 rules: [
                   {
-                    colSpan: 2,
-                    margin: '10px 0 0',
+                    layout: CompositionLayout.Grid,
```
supproduction: Please, don't use `CompositionLayout` we are going to deprecate it

supproduction (dismissed):

libs/template/presets/storefront/experience/pages/registration-page.ts
tolerants: Thanks, fixed.

apps/fulfillment/.env.template
```diff
@@ -1 +1,11 @@
-ORYX_FULFILLMENT_CLIENT_ID=fulfillment-client-local
+# The url where the backend is hosted.
+ORYX_FULFILLMENT_BACKEND_URL=
```
tobi-or-not-tobi, line 2: It would be nice if we could align this env var with the one we use in storefront. reason behind is that this would allow us to reuse launchpad logic to create env vars. If you agree, let's create a ticket for this.

libs/template/application/oryx-app/oryx-app.component.ts
```diff
@@ -11,9 +11,13 @@ export class OryxAppComponent extends LitElement {
 
   protected override render(): TemplateResult {
     return html`
-      <oryx-composition uid="header"></oryx-composition>
+      ${featureVersion < '1.2'
```
tobi-or-not-tobi: https://spryker.slack.com/archives/C029XPSBXC6/p1697630001287519

libs/template/presets/fulfillment/app.ts
```diff
@@ -43,13 +49,18 @@ export function fulfillmentFeatures(
     new BapiAuthFeature(),
     new BapiAuthComponentsFeature(),
     { resources: fulfillmentResources },
-    new FulfillmentRootFeature(config?.fulfillmentRoot),
+    featureVersion < '1.2'
```
tobi-or-not-tobi: `<= '1.2'

tobi-or-not-tobi (changes requested):

apps/fulfillment/.env.template
tolerants: Let's discuss it on catchup.

libs/template/application/oryx-app/oryx-app.component.ts
tolerants: Fixed.

libs/template/presets/fulfillment/app.ts
tolerants: Fixed.

apps/fulfillment/.env.template
tolerants: It's impossible as long as we gonna have the same variable for storefront and fulfillment on github.

tobi-or-not-tobi (dismissed):

apps/fulfillment/index.html
```diff
@@ -34,9 +34,14 @@
       href="https://fonts.googleapis.com/css2?family=Montserrat:wght@400;500;600;700&display=swap"
       rel="stylesheet"
     />
+    <style>
+      :root {
+        font-size: 16px;
+      }
+    </style>
```
tobi-or-not-tobi, line 41: For now this is OK. 

Capturing our conversation: 
The problem with the font-size is related to the `DefaultThemeMetaInitializer` we have. This is creating a hard-coded font-size of `14px` to the root of the document. This is not responsive currently - it must be `16px` on small screen.
While all styling typically belongs to the `<oryx-app>` element, the font-size must be set on the root, in order to make `rem` unit work. This is the only exception. Other stuff cascades down from the `oryx-app` element.

Going forward, I can think of 2 solutions:
1. make metaService responsive, observing the viewport and change the hardcoded font-size based on small vs large - bad idea, the meta service should not be bothered with this
2. introduce a additional section of document CSS variables (we only need font-size for now, but I've seen other cases where it could be useful). we then add the font-size here, and we can use that CSS var in the meta service.

tobi-or-not-tobi (approved):

PR #791, opened 2023-09-22
feat: add divider option attribute to navigation component

nsilva-spryker:
Add new divider option to show section division between option menu's
closes: [HRZ-89883](https://spryker.atlassian.net/browse/HRZ-89883)

[HRZ-89883]: https://spryker.atlassian.net/browse/HRZ-89883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/navigation/navigation-item/src/navigation-item.styles.ts
```diff
@@ -42,6 +49,24 @@ export const navigationItemStyles = css`
...
+
+  :host([divider]) {
+    margin-block-start: calc(var(--oryx-navigation-item-vertical-padding) * 2);
+  }
+
+  :host([divider]:host-context([collapsed]))::after {
+    transform: translateY(var(--oryx-navigation-item-divider-position))
+      translateX(-33%);
+  }
+
```
aleespryker, line 54: Does this work on Firefox? It's displaying with the wrong horizontal margins in at least storybook there.

nsilva-spryker: `:host-context` is not available on firefox. Code was refactor

libs/base/ui/navigation/navigation-item/src/navigation-item.component.ts
```diff
@@ -7,6 +7,25 @@ export class NavigationItemComponent extends LitElement {
...
+        this.removeAttribute('parent-collapsed');
+      }
+    });
+
+    mutationObserver.observe(parent, { attributes: true });
+
+    if (parent.hasAttribute('collapsed')) {
+      this.setAttribute('parent-collapsed', '');
+    }
+  }
```
aleespryker: Wouldn't it be possible to do this through css vars? So then we don't need the mutationObserver.

Something like 
```
:host([collapsed]) {
--transform: translateY(var(--oryx-navigation-item-divider-position)) translateX(-33%);
}
```

```
:host([divider]::after {
transform: var(--transform, none);
}
```

aleespryker (dismissed):

libs/base/ui/navigation/navigation-item/src/navigation-item.component.ts
nsilva-spryker: The collapsed attribute is from parent component that can be toggled every time, that's why we need the mutationObserver to update children component style according, if there is `collapsed` on parent or not.

libs/base/ui/navigation/navigation-item/src/navigation-item.styles.ts
```diff
@@ -42,6 +49,24 @@ export const navigationItemStyles = css`
...
+  :host([divider]) {
+    margin-block-start: calc(var(--oryx-navigation-item-vertical-padding) * 2);
+  }
+
+  :host([divider][parent-collapsed])::after {
+    transform: translateY(var(--oryx-navigation-item-divider-position))
+      translateX(-33%);
+  }
```
tobi-or-not-tobi: Seems overcomplicated, but I might miss something. I've simplified to:

```css

  :host([divider])::before {
    content: '';
    background-color: var(--oryx-color-neutral-5);
    width: calc(100% - 32px);
    height: 1px;
    position: absolute;
    inset-inline-start: 16px;
    margin-top: -64px;
  }
```

With this, it's not a breaking change and we do not need the MutationObserver

libs/base/ui/navigation/navigation-item/src/navigation-item.component.ts
tobi-or-not-tobi: I believwe don't need MutationObserver, see proposal at styles

libs/base/ui/navigation/navigation/src/stories/navigation.stories.ts
```diff
@@ -44,7 +44,12 @@ const Template: Story<unknown> = (): TemplateResult => {
             </svg>
           </oryx-icon>
 
-          test
+          Experience builder
+        </oryx-navigation-item>
+
+        <oryx-navigation-item divider>
+          <oryx-icon slot="icon" type="settings"> </oryx-icon>
+          Settings
```
tobi-or-not-tobi, line 52: can we add static stories for this one? With and without divider

tobi-or-not-tobi (changes requested):

libs/base/ui/navigation/navigation-item/src/navigation-item.styles.ts
nsilva-spryker: Yes, using the MutationObserver was overcomplicating what was simple, thks

tobi-or-not-tobi (approved):

PR #795, opened 2023-09-25
fix(ui): display password and input fields properly on SSR

aleespryker:
closes: [HRZ-89913](https://spryker.atlassian.net/browse/HRZ-89913)

[HRZ-89913]: https://spryker.atlassian.net/browse/HRZ-89913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/base/ui/form/password/src/password-input.component.ts
+++ b/libs/base/ui/form/password/src/password-input.component.ts
@@ -10,6 +10,7 @@ import {
   I18nMixin,
   I18nTranslationValue,
   Size,
+  hydrate,
   i18n,
 } from '@spryker-oryx/utilities';
 import { LitElement, PropertyValues, TemplateResult, html } from 'lit';
@@ -21,6 +22,7 @@ import {
 } from './password-input.model';
 import { baseStyles } from './password-input.styles';
 
+@hydrate()
 export class PasswordInputComponent
   extends I18nMixin(LitElement)
   implements FormControlOptions, AffixOptions, PasswordValidationOptions
@@ -62,15 +64,15 @@ export class PasswordInputComponent
   @property({ type: Boolean }) visible?: boolean;
 
   @state()
-  minLengthError = false;
+  minLengthError = true;
   @state()
-  maxLengthError = false;
+  maxLengthError = true;
   @state()
-  minUppercaseCharsError = false;
+  minUppercaseCharsError = true;
   @state()
-  minNumbersError = false;
+  minNumbersError = true;
   @state()
-  minSpecialCharsError = false;
+  minSpecialCharsError = true;
 
   connectedCallback(): void {
     super.connectedCallback();
--- a/libs/base/ui/form/password/src/password-input.component.ts
+++ b/libs/base/ui/form/password/src/password-input.component.ts
@@ -64,15 +64,15 @@ export class PasswordInputComponent
   @property({ type: Boolean }) visible?: boolean;
 
   @state()
-  minLengthError = true;
+  minLengthError = false;
   @state()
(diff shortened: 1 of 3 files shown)


libs/base/ui/form/password/src/password-input.component.ts
```diff
@@ -62,15 +64,15 @@ export class PasswordInputComponent
...
+  maxLengthError = true;
   @state()
-  minUppercaseCharsError = false;
+  minUppercaseCharsError = true;
   @state()
-  minNumbersError = false;
+  minNumbersError = true;
   @state()
-  minSpecialCharsError = false;
+  minSpecialCharsError = true;
```
supproduction: So we changed default state?

libs/platform/form/src/renderers/default-form.renderer.ts
```diff
@@ -208,6 +208,7 @@ export class DefaultFormRenderer implements FormRenderer {
         floatLabel=${ifDefined(field.floatLabel)}
         .style=${this.resolveStyles(field)}
         ?hasError=${field.attributes?.hasError}
+        ?required=${field.required}
```
supproduction, line 211: We have to add it only for text field?

supproduction (changes requested):

libs/base/ui/form/password/src/password-input.component.ts
aleespryker: Before hydration, these values are used to determine the active state and on the server, it showed up as `active` without running validation if you leave it `true`. But we wouldn't really have any input to test it on anyways since that all happens on the client and not the server side. I made with the assumption that the text field is empty from the server's perspective, so nothing is valid.

Unless there's a more preferable approach to this or if we want the overhead of a prehydration script to do the check.

libs/base/ui/form/password/src/password-input.component.ts
```diff
@@ -62,15 +64,15 @@ export class PasswordInputComponent
   @property({ type: Boolean }) visible?: boolean;
 
   @state()
-  minLengthError = false;
```
Sinucid: I believe that this is not a component issue. Maybe better to hydrate UserRegistrationComponent on window:load and add a TODO comment to find better solution?

Sinucid (changes requested):

libs/base/ui/form/password/src/password-input.component.ts
aleespryker: Prehydration script is another method. Not sure though if we are happy with that solution over just hydrating on load.

libs/platform/form/src/renderers/default-form.renderer.ts
aleespryker: There are a few components for which this does not apply like toggle, and doesn't show up even when hydrated.

libs/base/ui/form/password/src/password-input.component.ts
aleespryker: Should be addressed in a separate ticket. Hydrate on load for now.

libs/base/ui/form/password/src/password-input.component.ts
aleespryker: See Pavlo's comment below.

Sinucid (approved):

tobi-or-not-tobi:
@Sinucid can you merge this PR please?

supproduction (approved):

PR #796, opened 2023-09-26
feat(search): range facet component

Sinucid:
Basic range facet component with storybook integration and unit tests

closes: [HRZ-1954](https://spryker.atlassian.net/browse/HRZ-1954), [HRZ-2000](https://spryker.atlassian.net/browse/HRZ-2000), [HRZ-89934](https://spryker.atlassian.net/browse/HRZ-89934)

[HRZ-1954]: https://spryker.atlassian.net/browse/HRZ-1954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-2000]: https://spryker.atlassian.net/browse/HRZ-2000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-89934]: https://spryker.atlassian.net/browse/HRZ-89934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/search/facet-range/facet-range.component.ts
```diff
@@ -0,0 +1,153 @@
...
+    if (e.key === 'Enter') {
+      (e.target as HTMLInputElement).blur();
+    }
+  }
+
+  protected hasChangedValue({ min, max }: RangeFacetValue): boolean {
+    const facet = this.facet() as RangeFacet;
+    const {
+      values: { selected },
+    } = facet;
```
supproduction:
```suggestion
    const {
      values: { selected },
    } = this.facet() as RangeFacet;
    
    or should we check n `facet` first?

```

libs/domain/search/facet-range/facet-range.component.ts
```diff
@@ -0,0 +1,153 @@
...
+        this.controller.dispatchSelectEvent({ selected });
+      }
+    },
+    300
+  );
+
+  protected onBlur(e: InputEvent): void {
+    const { value, name } = e.target as HTMLInputElement;
+    this[name as keyof SearchFacetRangeComponentValues] = Number(value);
+  }
```
supproduction, line 79: But how it's possible to change value and has not been notified about it before blur?

libs/domain/search/facet-value-navigation/facet-value-navigation.component.ts
```diff
@@ -26,11 +26,12 @@ export class SearchFacetValueNavigationComponent
   @property({ type: Boolean }) enableSearch?: boolean;
   @property({ type: Boolean }) open?: boolean;
   @property({ type: Boolean }) enableClear?: boolean;
+  @property({ type: Boolean }) dirty?: boolean;
```
supproduction, line 29: mb better name `touched`?

libs/domain/search/facet/facet.component.ts
```diff
@@ -35,7 +36,7 @@ export class SearchFacetComponent extends I18nMixin(LitElement) {
   }
 
   protected override render(): TemplateResult | void {
-    const facet = this.facet();
+    const facet = this.facet() as SingleMultiFacet;
 
     if (!facet) return;
```
supproduction: Why we don't check on type here. And if you check on type you dont need typecasting

supproduction (changes requested):

libs/domain/search/facet/controllers/facet.controller.ts
```diff
@@ -48,46 +54,58 @@ export class FacetController implements ReactiveController {
...
 
-    return facet
-      ? facet.values.filter(({ name }) =>
-          (facet.selectedValues ?? []).includes(name as string)
-        )
-      : [];
+    if (!facet) return [];
+
+    if (facet.type === FacetType.Range) {
+      return (facet as RangeFacet).values.selected!;
```
supproduction: Do you need typecasting here?

supproduction (changes requested):

libs/domain/search/facet-value-navigation/facet-value-navigation.component.ts
Sinucid: `touched` usually means some interaction with element, first focus or click, even without changing the values. dirty is more about differences between current and initial values

libs/domain/search/facet-range/facet-range.component.ts
Sinucid: According the criteria for the story

libs/domain/search/facet/controllers/facet.controller.ts
Sinucid: Nope, dropped

libs/domain/search/facet-value-navigation/facet-value-navigation.component.ts
supproduction: and how it's possible to be dirty without interaction?

libs/domain/search/facet/facet.component.ts
Sinucid: refactored

libs/domain/search/facet-value-navigation/facet-value-navigation.component.ts
Sinucid: I meant that `touched` state means that user 'touched' the input, focused, clicked on it and etc, this 
does not guarantee that values were changed. `dirty` means that values were changed according the initial ones. In my case I need to make some emphasis that values were changed - thats why I believe that dirty is more clear and suitable

Sinucid: Or maybe do you have better ideas instead of `touched`?

libs/base/ui/form/multi-range/src/multi-range.model.ts
```diff
@@ -41,3 +41,8 @@ export interface MultiRangeProperties {
 
   disabled?: boolean;
 }
+
+export interface MultiRangeChangeEvent {
+  minValue: number;
+  maxValue: number;
```
tobi-or-not-tobi, line 47: what about just `min` and `max`?

libs/domain/product/src/services/adapter/normalizers/facet-range/facet-range.normalizer.ts
```diff
@@ -19,19 +19,18 @@ export function facetsRangeNormalizer(
...
     const normalizedFacet = {
+      type: FacetType.Range,
       name: localizedName,
       parameter: config.parameterName,
-      values: facetValues,
-      selectedValue: [],
-      ...(config.isMultiValued && { multiValued: config.isMultiValued }),
-    };
+      values,
+    } as RangeFacet;
```
tobi-or-not-tobi: `const normalizedFacet: RangeFacet = {` seems better, as the `as` could be used to force the type

libs/domain/search/facet-range/facet-range.component.ts
```diff
@@ -0,0 +1,153 @@
...
+      <oryx-input .label=${label}>
+        <input
+          aria-label=${ifDefined(label)}
+          name=${name}
+          type="number"
+          .value=${String(this[name])}
+          min=${min}
+          max=${max}
+          step=${this.step}
+          @blur=${this.onBlur}
```
tobi-or-not-tobi, line 125: The input is changing on blur or enter, but the slider instantly changes the inputs on drag'n drop. Wondering if this is deliberately? I can imagine that changing instantly is too heavy on the list results, but that's more because the backend might be slow. Any takes on the ux here?

tobi-or-not-tobi (changes requested):
I've changed the code locally a bit to render the range component for the facets:
- dropped price in bury
- changed the `default` provided facet component, and switched to the `oryx-search-range-facet` component when there's a facet value. The later was not convenient, as I had to write a condition based on the appearance of min/max.

libs/domain/search/facet-range/facet-range.component.ts
Sinucid: Slider is debounced by 300ms, check implementation of onRangeChange method

libs/base/ui/form/multi-range/src/multi-range.model.ts
Sinucid: It will be a BC. I've just provided missed public interface for the `CustomEvent`, these field names are already part of Alexander's implementation of multi-range component

tobi-or-not-tobi: ok, we'll keep it as-is

libs/domain/search/facet-range/facet-range.component.ts
Sinucid: In result component emit a select event, and how to handle it is out of component's scope, logic can be various

libs/domain/search/facet-range/facet-range.component.ts
```diff
@@ -0,0 +1,153 @@
...
+  extends LitElement
+  implements SearchFacetRangeComponentAttributes
+{
+  static styles = [searchRangeFacetStyles];
+
+  protected controller = new FacetController(this);
+
+  @signalProperty() name?: string;
+  @signalProperty({ type: Boolean }) open = false;
+  @signalProperty({ type: Boolean }) enableClear = true;
```
tarashaliuk: What about using `disableClear = false` instead? We should avoid using props with a `true` value as default.

libs/domain/search/facet-range/facet-range.component.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  protected renderControls(facet: RangeFacet): TemplateResult {
+    const {
+      values: { min, max },
+    } = facet;
+    return html`
+      ${this.renderInput('min', min, max - 1, this.labelMin)}
+
+      <span></span>
```
tarashaliuk: Why do we need an empty span?

libs/domain/search/facet-value-navigation/facet-value-navigation.component.spec.ts
```diff
@@ -164,7 +164,7 @@ describe('SearchFacetValueNavigationComponent', () => {
         element = await fixture(
           html`<oryx-search-facet-value-navigation
             enableClear
-            .selectedLength=${mockSelectedLength}
+            dirty
             @oryx.clear=${callback}
           ></oryx-search-facet-value-navigation>`
         );
```
tarashaliuk, line 170: description says `and selectedLength is provided`, but there is no `selectedLength` provided. What about testing `selectedLength` and `dirty` props in separate tests?

tarashaliuk (changes requested):

libs/domain/search/facet-range/facet-range.component.ts
Sinucid: Yup, it is used as separator between inputs (separator is a rectangle 8 x 2 pxx in designs)

libs/domain/search/facet-value-navigation/facet-value-navigation.component.spec.ts
```diff
@@ -144,7 +144,7 @@ describe('SearchFacetValueNavigationComponent', () => {
     });
   });
 
-  describe('when clear is enabled without selectedLength', () => {
+  describe('when clear is enabled without dirty', () => {
```
tarashaliuk:
```suggestion
  describe('when clear is enabled', () => {
```

libs/domain/search/facet-range/facet-range.component.ts
tarashaliuk: What about using `div` instead of span? I see you set `display: block` for this `span` in styles, but `div` is the block-level element by default.

tarashaliuk (changes requested):

supproduction (approved):

libs/domain/search/facet-range/facet-range.component.ts
tobi-or-not-tobi: Agreed to keep this as-is for now.

libs/domain/search/facet-range/facet-range.component.ts
tobi-or-not-tobi: If we stick to an element, why not use an `<hr />` then?

tarashaliuk (approved):

tobi-or-not-tobi (approved):

PR #799, opened 2023-09-30
feat: event repeater

supproduction:
adds event repeater 

closes [HRZ-3277](https://spryker.atlassian.net/browse/HRZ-3277)

[HRZ-3277]: https://spryker.atlassian.net/browse/HRZ-3277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/src/index.html
```diff
@@ -6,5 +6,72 @@
   <body>
     <oryx-app></oryx-app>
   </body>
+  <script>
+    const EVENTS_DATA = Symbol.for('EVENTS_DATA');
```
tobi-or-not-tobi: I assume this is POC for now, as we like to keep boilerplate clean. Will we be able to use a small module from packages instead of inline code here? Or is this done deliberatly?

apps/storefront/src/index.html
```diff
@@ -6,5 +6,72 @@
...
+    const replayableAttribute = 'replayable';
```
tobi-or-not-tobi: gpt suggests: 

```ts
const EVENTS_DATA = Symbol.for('EVENTS_DATA');
const replayableAttribute = 'replayable';

// Function to add events to elements with the replayable attribute
const addEventsAction = () => {
  const elements = document.querySelectorAll(`[${replayableAttribute}]`);

  elements.forEach(element => {
    element[EVENTS_DATA] = {
      events: [],
      listener: (event) => {
        event.stopPropagation();
        event.preventDefault();
        element[EVENTS_DATA]?.events.push(event);
      },
    };

    const events = element.getAttribute(replayableAttribute).split(':');

    events.forEach(eventType => {
      element.addEventListener(eventType, element[EVENTS_DATA].listener);
    });
  });
};

// Check if shadow DOM exists in oryx-app
const oryxApp = document.querySelector('oryx-app');
if (oryxApp && oryxApp.shadowRoot) {
  addEventsAction();
}

```

apps/storefront/src/index.html
```diff
@@ -6,5 +6,72 @@
...
+
+        const events = element.getAttribute(replayableAttribute);
+
+        for (const eventType of events.split(':')) {
+          element.addEventListener(eventType, element[EVENTS_DATA].listener);
+        }
+      }
+    };
+
+    const isShadow = !!document.querySelector?.('oryx-app').shadowRoot;
```
tobi-or-not-tobi: How do we go about custom root elements?

libs/base/utilities/src/misc/events-repeater.ts
```diff
@@ -0,0 +1,97 @@
...
+export type ElementWithEventsData = HTMLElement & {
+  [EVENTS_DATA]?: EventsData;
+};
+
+interface EventsData {
+  listener: (event: Event) => void;
+  events: Event[];
+}
+
+const treewalk = (
```
tobi-or-not-tobi: Wondering if this should be the (partially) same code that is used in the index.html

libs/base/utilities/src/misc/events-repeater.ts
```diff
@@ -0,0 +1,97 @@
+export const EVENTS_DATA = Symbol.for('EVENTS_DATA');
+export const replayableAttribute = 'replayable';
```
tobi-or-not-tobi: we use repeater (file) and replayable (var). Shoudl we standardize on one name?

libs/domain/cart/add/src/add.component.ts
```diff
@@ -65,6 +65,7 @@ export class CartAddComponent extends ProductMixin(
       ?disabled=${this.isInvalid || !this.$hasStock()}
       @click=${this.onSubmit}
       @mouseup=${this.onMouseUp}
+      replayable="click"
```
tobi-or-not-tobi: We probably need to support multiple events, e.g. focus and hover. Array? Comma/space separated?

tobi-or-not-tobi (changes requested):

apps/storefront/src/index.html
supproduction: Why we dropped treewalk here and also use forEach instead of `for of` which is slower?

apps/storefront/src/index.html
supproduction: It was my next step

apps/storefront/src/index.html
supproduction: When we move it to module, theoretically we should be able to drop it

libs/base/utilities/src/misc/events-repeater.ts
supproduction: when we move it to separate module I planned to reuse this code

apps/storefront/src/index.html
tobi-or-not-tobi: I wasn't aware of the performance decrease, also no opinions on the code so much. Was more coming from smaller code lines ;)

apps/storefront/src/index.html
tobi-or-not-tobi: sounds good, please update the comments when you've done that.

libs/base/utilities/src/misc/events-repeater.ts
```diff
@@ -0,0 +1,134 @@
...
+function eventsAction(
+  element: ElementWithEventsData,
+  shouldRemove?: true
+): void {
+  const events = element.getAttribute(repeatableAttribute) as string;
+
+  for (const eventType of events.split(',')) {
+    if (shouldRemove) {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      element.removeEventListener(eventType, element[EVENTS_DATA]!.listener);
```
tobi-or-not-tobi: We should probably trim the `eventType`, same a few lines below. And add a test for it.

libs/domain/cart/add/src/add.component.ts
tobi-or-not-tobi: trimming is missing

libs/base/utilities/src/misc/events-repeater.ts
```diff
@@ -0,0 +1,144 @@
+const eventsDataIdentifier = 'EVENTS_DATA';
```
dunqan: what about `ORYX.EVENT_DATA` or `ORYX.PREHYDRATION_EVENTS` ?

dunqan (dismissed):

libs/base/utilities/src/misc/hydrate-events.ts
```diff
@@ -0,0 +1,197 @@
...
+        if (!acc.includes(event.trim())) acc.push(event.trim());
+      }
+
+      return acc;
+    },
+    [...types, HYDRATE_EVENT]
+  );
+};
+
+export const stopEventsForHydrationInsertion = `
```
dunqan:
```suggestion
export const captureEventsForHydrationInsertion = `
```
?

libs/template/application/src/services/initializers/stop-events.initializer.ts
```diff
@@ -0,0 +1,21 @@
+import { AppInitializer, PageMetaService } from '@spryker-oryx/core';
+import { inject } from '@spryker-oryx/di';
+import { stopEventsForHydrationInsertion } from '@spryker-oryx/utilities';
+
+export const StopEventsInitializer = `${AppInitializer}StopEvents`;
+
+export class DefaultStopEventsInitializer implements AppInitializer {
```
dunqan:
```suggestion
export class DefaultCaptureEventsInitializer implements AppInitializer {
```
?

dunqan (approved):

tobi-or-not-tobi (approved):

PR #804, opened 2023-10-05
fix(cart): ensure handling of string-based cart entry quantities

tobi-or-not-tobi:
While connecting to a b2b backend in preparation of customer call, it turns out that the cart entry quantity is of type string. So far, we’ve been seeing the number type for quantity. 

We’re not entirely sure if quantity is always going to be numeric, hence we do not force the conversion of the response in the adapter layer. For now, we’ll assume it’s a quantity in the component layer, but over time we can change this when we have more clarity on the usage of string vs number.

closes: [HRZ-90046](https://spryker.atlassian.net/browse/HRZ-90046)

[HRZ-90046]: https://spryker.atlassian.net/browse/HRZ-90046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/cart/add/src/add.component.ts
+++ b/libs/domain/cart/add/src/add.component.ts
@@ -96,14 +96,29 @@ export class CartAddComponent extends ProductMixin(
     if (availability?.isNeverOutOfStock) return Infinity;
     if (availability?.quantity)
       return (
-        availability?.quantity -
+        availability.quantity -
         this.$entries()
           .filter((entry) => entry.sku === sku)
-          .map((entry) => entry.quantity)
-          .reduce((a: number, b) => a + b, 0)
+          .map((entry) => Number(entry.quantity))
+          .reduce((a: number, b) => a + Number(b), 0)
       );
     return 0;
   });
+
+  // protected $max = computed(() => {
+  //   const { availability, sku } = this.$product() ?? {};
+
+  //   if (availability?.isNeverOutOfStock) return Infinity;
+  //   if (availability?.quantity) {
+  //     const totalQuantity = this.$entries()
+  //       .filter((entry) => entry.sku === sku)
+  //       .map((entry) => entry.quantity as number) // Type assertion here
+  //       .reduce((a: number, b) => a + Number(b), 0);
+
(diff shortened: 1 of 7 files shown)


libs/domain/cart/add/src/add.component.ts
```diff
@@ -96,15 +96,30 @@ export class CartAddComponent extends ProductMixin(
...
           .filter((entry) => entry.sku === sku)
-          .map((entry) => entry.quantity)
-          .reduce((a: number, b) => a + b, 0)
+          .map((entry) => Number(entry.quantity))
+          .reduce((a: number, b) => a + Number(b), 0)
       );
     return 0;
   });
 
+  // protected $max = computed(() => {
```
Sinucid: Do we need this? If yes - leave a comment why

libs/domain/cart/src/services/resolver/cart.resolver.spec.ts
```diff
@@ -22,6 +22,14 @@ const bigCart = {
   ],
 };
 
+const bigB2bCart = {
```
Sinucid:
```suggestion
const variousValuesType = {
```

libs/domain/cart/add/src/add.component.ts
```diff
@@ -96,15 +96,30 @@ export class CartAddComponent extends ProductMixin(
...
     if (availability?.quantity)
       return (
-        availability?.quantity -
+        availability.quantity -
         this.$entries()
           .filter((entry) => entry.sku === sku)
-          .map((entry) => entry.quantity)
-          .reduce((a: number, b) => a + b, 0)
+          .map((entry) => Number(entry.quantity))
+          .reduce((a: number, b) => a + Number(b), 0)
```
Sinucid:
```suggestion
          .reduce((a, {quantity}) => a + Number(quantity), 0)
```

Sinucid (changes requested):

Sinucid (approved):

PR #807, opened 2023-10-06
chore: set client id as env variable

tarashaliuk:
Replaced hardcoded client_id with env variable. Env variables were set on Netlify as well.

closes: https://spryker.atlassian.net/browse/HRZ-90055


libs/platform/auth/src/bapi/feature.ts
```diff
@@ -34,7 +34,7 @@ export class BapiAuthFeature extends OauthFeature implements AppFeature {
       providers: [
         {
           id: 'spryker',
-          clientId: 'frontend',
```
AndreyYevtukhov, line 37: Should we set `frontend` as a default one (instead of an empty string) to make tests work?

apps/fulfillment/.env.template
```diff
@@ -1,2 +1,3 @@
 ORYX_FULFILLMENT_BACKEND_URL=http://localhost:3001
 ORYX_LABS=true
+ORYX_FULFILLMENT_CLIENT_ID=frontend
```
tolerants: Please add a line break.

AndreyYevtukhov (approved):

tolerants (approved):

PR #810, opened 2023-10-09
feat: show flag before to current locale

tobi-or-not-tobi:
Renders a flag emoji before the current locale, and update accordingly whenever the current locale is updated. 

The flag is an emoji that is calculated based on the given locale and country code. 

closes: [HRZ-90056](https://spryker.atlassian.net/browse/HRZ-90056)

[HRZ-90056]: https://spryker.atlassian.net/browse/HRZ-90056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/site/locale-selector/locale-selector.component.1.2.spec.ts
```diff
@@ -0,0 +1,100 @@
+import { fixture } from '@open-wc/testing-helpers';
```
tolerants: Do we really need a separate file per version? What are benefits?

libs/domain/site/locale-selector/locale-selector.component.ts
```diff
@@ -64,4 +76,22 @@ export class SiteLocaleSelectorComponent extends ContentMixin(LitElement) {
     });
     return languageNames.of(code) ?? code;
   }
+
+  /**
+   * @returns Flag Emoji for the locale country (e.g. 🇺🇸)
+   *
+   * @since 1.2
+   */
+  protected getFlag(locale: string): TemplateResult {
```
tolerants, line 85: Maybe `getFlagEmoji` so in case we want to implement an image flag in the future?

libs/domain/site/locale-selector/locale-selector.model.ts
```diff
@@ -0,0 +1,8 @@
+export interface LocaleSelectorOptions {
+  /**
+   * Shows a flag next to the selected locale.
+   *
+   * @since 1.2
+   */
+  enableFlag?: boolean;
```
tolerants, line 7: I think it's worth to update a demo-story.

tolerants (changes requested):

libs/domain/site/locale-selector/locale-selector.component.ts
tobi-or-not-tobi: I'd rather keep the API consistent and not tight to a specific media/format. Hence, the `TemplateResult` return type

libs/domain/site/locale-selector/locale-selector.component.1.2.spec.ts
tobi-or-not-tobi: We do not have a solution to do this in 1 file. 
@dunqan might come up with a solution going forward.

libs/domain/site/locale-selector/locale-selector.model.ts
tobi-or-not-tobi: Yes, let's discuss with @AndreyYevtukhov as we currently render stories in 1.0 version only

tolerants (approved):

PR #813, opened 2023-10-10
feat: price mode only used in b2b

tobi-or-not-tobi:
We no longer add the price mode selector to b2c, but only to the b2b experience data. 

closes: [HRZ-90073](https://spryker.atlassian.net/browse/HRZ-90073)

[HRZ-90073]: https://spryker.atlassian.net/browse/HRZ-90073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tolerants:
But it's a PR for the whole feature, not only for including it in our b2b preset. So it's a `feature`.

libs/template/presets/storefront/experience/header.ts
```diff
@@ -1,51 +1,65 @@
...
+
+  components.push({ type: 'oryx-site-currency-selector' });
+  components.push({ type: 'oryx-site-locale-selector' });
+
+  components[0].options = { rules: [{ style: 'margin-inline-start: auto' }] };
+
+  return components;
+};
+
+export const topHeader = (options?: {
```
tolerants, line 55: additional? secondary?

tolerants (dismissed):

PR #816, opened 2023-10-10
feat: impemented addditional requirements for facet rating component

tarashaliuk:
Implemented additional requirements that were described [here](https://spryker.atlassian.net/browse/HRZ-1947?focusedCommentId=428481)

closes: https://spryker.atlassian.net/browse/HRZ-90076


libs/domain/search/facet-rating/facet-rating.component.ts
```diff
@@ -1,26 +1,88 @@
...
 import { TemplateResult, html } from 'lit';
+import { repeat } from 'lit/directives/repeat.js';
 import { when } from 'lit/directives/when.js';
+import { SearchRatingFacetComponentOptions } from './facet-rating.model';
 import { searchFacetRatingStyles } from './facet-rating.styles';
 
-export class SearchRatingFacetComponent extends SearchFacetComponent {
+@defaultOptions({
+  min: 1,
+  max: 4,
```
tobi-or-not-tobi, line 16: should we keep min to 1 min? We can change it to `0` and it creates an additional non-star

libs/domain/search/facet-rating/facet-rating.component.ts
```diff
@@ -1,26 +1,88 @@
...
+      )}
+    </ul>`;
+  }
+
+  protected renderValueControlLabel(facetValue: FacetValue): TemplateResult {
+    const { scale } = this.$options();
+
+    if (!scale) {
+      return html``;
+    }
```
tobi-or-not-tobi:
```suggestion
  protected renderValueControlLabel(facetValue: FacetValue): TemplateResult | void {
    const { scale } = this.$options();

    if (!scale) return;
```

libs/domain/search/facet-rating/facet-rating.component.ts
```diff
@@ -1,26 +1,88 @@
...
         <oryx-rating
           readonly
           value=${facetValue.value}
-          scale=${5}
+          scale=${scale}
         ></oryx-rating>
-        ${when(Number(facetValue.value) < 5, () => html`<span>& up</span>`)}
+        ${when(
+          Number(facetValue.value) < Number(scale),
+          () => html`<span>& up</span>`
```
tobi-or-not-tobi: let's use i18n here. we do not need to mimic the `&` in the code, let's add a translation in labs package though.

libs/domain/search/facet-rating/facet-rating.model.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  /**
+   * Indicates the maximum rating value for a rating facet.
+   *
+   * @default 4
+   */
+  max?: number;
+
+  /**
+   * Indicates the scale for a rating facet.
```
tobi-or-not-tobi: What does scale mean? can we elaborate in the docs here?

tobi-or-not-tobi (changes requested):
Can we introduce the removal of this buried rating facet with featureVersion flag?

libs/domain/search/facet-rating/facet-rating.component.ts
tarashaliuk: Do you mean to create options validation that will block using the wrong values like `0`, `-1`  for `min` and other validations to rest options?

tobi-or-not-tobi: yes, or gracefully change to min 1. But i'm not sure if there are negative consequences tbh

libs/domain/search/facet-rating/facet-rating.component.ts
tarashaliuk: It is overridden method so we can't change the return type

libs/domain/search/facet-rating/facet-rating.component.ts
```diff
@@ -1,26 +1,82 @@
...
         <oryx-rating
           readonly
           value=${facetValue.value}
-          scale=${5}
+          scale=${scale}
         ></oryx-rating>
-        ${when(Number(facetValue.value) < 5, () => html`<span>& up</span>`)}
+        ${when(
+          Number(facetValue.value) < scale,
+          () => html`<span>${this.i18n('& up')}</span>`
```
tobi-or-not-tobi:
```suggestion
          () => html`<span>${this.i18n('search.facet.rating.up')}</span>`
```

We should add an entry at `libs/template/labs/src/i18n/translations/en.ts` with the translated `& up`.

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #820, opened 2023-10-12
feat(search): price facet

Sinucid:
- added new method `getCurrencySymbol` to the currency service that returns symbol of active currency according the selected locale

closes: [HRZ-90048](https://spryker.atlassian.net/browse/HRZ-90048)

[HRZ-90048]: https://spryker.atlassian.net/browse/HRZ-90048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/product/src/services/adapter/normalizers/facet-range/facet-range.normalizer.ts
```diff
@@ -7,8 +7,8 @@ export function facetsRangeNormalizer(
   rangeFacets: ApiProductListModel.RangeFacet[]
 ): RangeFacet[] {
   return rangeFacets.reduce((normalizedFacetList: RangeFacet[], facet) => {
-    //ignore the facet if there is no difference between min and max
-    if (facet.min === facet.max) {
+    //ignore the facet if difference between min and max is 1
+    if (facet.max - facet.min <= 1) {
```
supproduction, line 11: question, do we have step in response, if yes should we rely on it?

libs/domain/search/facet-price/facet-price.component.ts
```diff
@@ -0,0 +1,72 @@
...
+    const {
+      values: { min, max },
+    } = facet;
+
+    return html`
+      ${this.renderInput('min', min, max - 1, labelMin)}
+
+      <hr />
+
+      ${this.renderInput('max', min + 1, max, labelMax)}
```
supproduction, line 60: do we have `renderInput` method in this component?

libs/domain/search/facet-range/facet-range.component.ts
```diff
@@ -87,12 +97,32 @@ export class SearchRangeFacetComponent
...
   }
 
+  protected syncInputsValues(min: number, max: number): void {
+    const minInput = this.renderRoot.querySelector(
+      `input[name="min"]`
+    ) as HTMLInputElement;
+
+    const maxInput = this.renderRoot.querySelector(
+      `input[name="max"]`
+    ) as HTMLInputElement;
```
supproduction:
```suggestion
    const minInput = this.renderRoot.querySelector(
      `input[name="min"]`
    ) as HTMLInputElement;

    const maxInput = this.renderRoot.querySelector<HTMLInputElement>(
      `input[name="max"]`
    );
```

libs/domain/search/facet-range/stories/demo.stories.ts
```diff
@@ -10,6 +10,20 @@ import { SearchFacetRangeComponentAttributes } from '../facet-range.model';
 
 export default {
   title: `${storybookPrefix}/Facet Range`,
+  // disables Chromatic's snapshotting on a story level
+  parameters: { chromatic: { disableSnapshot: true } },
```
supproduction:
```suggestion
  parameters: { chromatic: { disableSnapshot: true } },
```

libs/domain/site/src/services/currency/default-currency.service.ts
```diff
@@ -11,7 +20,8 @@ export class DefaultCurrencyService implements CurrencyService {
 
   constructor(
     protected storeService = inject(StoreService),
-    protected queryService = inject(QueryService)
+    protected queryService = inject(QueryService),
+    protected localeService = inject(LocaleService)
```
supproduction: Add optional for backward compatibility with versions less then 1.2

```suggestion
    protected localeService = inject(LocaleService, null)
```

libs/domain/site/src/services/currency/default-currency.service.ts
```diff
@@ -28,6 +38,22 @@ export class DefaultCurrencyService implements CurrencyService {
...
+    return combineLatest([this.get(), this.localeService.get()]).pipe(
+      map(([currency, locale]) =>
+        (0)
+          .toLocaleString(locale, {
+            style: 'currency',
+            currency: currency,
+            minimumFractionDigits: 0,
+            maximumFractionDigits: 0,
+          })
+          .replace(/\d/g, '')
```
supproduction, line 57: Suggestion, just cut first char instead of using replace + Regex

supproduction (changes requested):

libs/domain/product/src/services/adapter/normalizers/facet-range/facet-range.normalizer.ts
Sinucid: Nope, BE does not care about step. The idea was to have at least the integer to prevent floating values. In result steps of the range slider controlled by the browser automatically depends on available screen width. For more accurate interaction user has min and max inputs

libs/domain/search/facet-price/facet-price.component.ts
Sinucid: yes, it is in super class

libs/domain/site/src/services/currency/default-currency.service.ts
Sinucid: depends on locale position of the currency symbol can be various

supproduction (approved):

PR #823, opened 2023-10-13
chore: upgrade packages

tarashaliuk:
Updated dependencies:
- @storybook/jest from 0.0.10 to 0.2.3 (done)
- @nrwl/storybook from 14.5.8 to 16.9.1 (done)
- stylelint from 14.16.1 to 15.10.3 (done)
- word-wrap from 1.2.3 to 1.2.4 (not exists)
- eslint-plugin-prettier from 3.4.1 to 5.0.0 (done)
- glob-parent  from 3.1.0 to 5.1.2 (not exists)
- webpack from 4.46.0 to 5.77.0 (done)
- @storybook/web-components 6.5.16 to 7.0.2 (done)

partially closes: https://spryker.atlassian.net/browse/HRZ-89933


libs/base/ui/overlays/modal/src/stories/interactive/close-modal.stories.ts
```diff
@@ -31,5 +31,5 @@ CloseModal.play = async (obj: {
   userEvent.click(closeBtn);
   await new Promise((r) => setTimeout(r, 1000));
 
-  expect(modal.hasAttribute('open')).toBeFalsy();
+  await expect(modal.hasAttribute('open')).toBeFalsy();
```
supproduction, line 31: Can we use waitFor here?

libs/base/ui/overlays/modal/src/stories/interactive/close-nested-modal.stories.ts
```diff
@@ -33,5 +33,5 @@ CloseNestedModal.play = async (obj: {
   userEvent.click(closeNestedModalBtn);
   await new Promise((r) => setTimeout(r, 1000));
 
-  expect(nestedModal.hasAttribute('open')).toBeFalsy();
+  await expect(nestedModal.hasAttribute('open')).toBeFalsy();
```
supproduction, line 34: Can we use waitFor here?

supproduction: Or can we omit it from other tests and use just await?

migrations.json
supproduction, line 1: Do we need it?

nx.json
```diff
@@ -23,17 +15,18 @@
           "stylelint",
           "paths"
         ],
-        "parallel": 1
+        "parallel": 1,
+        "accessToken": "OGI1YjE1ZTMtMDJhMi00MDVlLTkwNWItMWViY2ZjNmNmZTFjfHJlYWQtd3JpdGU="
```
supproduction: Do we need this token? Where ti comes from?

package.json
```diff
@@ -66,73 +66,80 @@
...
-    "@nrwl/workspace": "15.7.2",
+    "@nrwl/js": "^16.0.0",
+    "@nx/cypress": "^16.0.0",
+    "@nx/devkit": "^16.0.0",
+    "@nx/eslint-plugin": "^16.0.0",
+    "@nx/js": "^16.0.0",
+    "@nx/linter": "^16.0.0",
+    "@nx/storybook": "^16.0.0",
+    "@nx/web": "^16.0.0",
+    "@nx/workspace": "^16.0.0",
```
supproduction, line 82: Can we use the same approach for all other new libs?

package.json
```diff
@@ -66,73 +66,80 @@
...
+    "nx": "^16.0.0",
+    "nx-cloud": "16.4.0",
     "nx-stylelint": "^13.5.3",
     "postcss": "^8.4.16",
     "postcss-html": "^1.5.0",
     "postcss-syntax": "^0.36.2",
     "prettier": "2.7.1",
     "prettier-plugin-organize-imports": "^3.2.3",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
```
supproduction, line 134: Why do we need it?

package.json
```diff
@@ -66,73 +66,80 @@
...
     "express": "^4.17.1",
     "husky": "^7.0.2",
     "jest-useragent-mock": "^0.1.1",
     "jsdom": "^20.0.0",
     "lerna": "^6.0.0",
     "mockdate": "^3.0.5",
     "npm-run-all": "^4.1.5",
-    "nx": "14.5.8",
+    "nx": "^16.0.0",
+    "nx-cloud": "16.4.0",
```
supproduction: Do we need nx-cloud?

package.json
```diff
@@ -141,6 +148,7 @@
     "stylelint-no-unsupported-browser-features": "^5.0.3",
     "stylelint-order": "^5.0.0",
     "stylelint-use-logical-spec": "^3.2.2",
+    "tsconfig-paths-webpack-plugin": "^4.1.0",
```
supproduction: Why do we need it for webpack?

supproduction (changes requested):

package.json
tarashaliuk: because we use it Storybook. Will be removed when we will move to vite instead of webpack

package.json
tarashaliuk: it is recommended as a caching tool that could speedup CI builds, but it shouldn't be in the scope of the current story. Will remove it

package.json
tarashaliuk: Storybook 7 requires react

nx.json
tarashaliuk: It was added with nx-cloud. Will remove it

migrations.json
tarashaliuk: probably not. will remove it

libs/base/ui/overlays/modal/src/stories/interactive/close-modal.stories.ts
tarashaliuk: better to stay with await

PR #824, opened 2023-10-15
fix(ui): multi-range input values recalculation and event dispatching behaviour

Sinucid:
Refactoring of multi-range component. Backward compatibility has been maintained. New behaviour of the component: 

1. fixed bug with property calculation when a few props are changed at once
2. change event is emitted when user stop dragging and release mouse button
3. input become disabled when invalid set of properties are passed (min bigger than max and vice versa, step is bigger than max - min)

Also: 

1. according the changes the range facet component was simplified
2. fixed bug in `getFacet` method of facet list service, that it accidentally returned facet without `name` and `parameter` as the result of the method call (due to falsy-possitive comparison undefined === undefined)

closes: [HRZ-90091](https://spryker.atlassian.net/browse/HRZ-90091)

[HRZ-90091]: https://spryker.atlassian.net/browse/HRZ-90091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/form/multi-range/src/multi-range.component.ts
```diff
@@ -13,142 +12,123 @@ export class MultiRangeComponent
...
+    this.invalid =
+      this.observedKeys.some((key) => typeof this[key] === 'undefined') ||
+      this.min! > this.max! - this.step ||
+      this.max! < this.min! + this.step;
+
+    return this.invalid;
+  }
+
+  protected override render(): TemplateResult | void {
+    if (this.invalid) return;
```
tolerants: This is a change of behavior, so we can consider it as a fix... or wrap with feature flag. In first case we will need to change the type of commit to `fix`.

libs/base/ui/form/multi-range/src/multi-range.component.ts
```diff
@@ -13,142 +12,123 @@ export class MultiRangeComponent
...
-  }
-  set max(value: number) {
-    const oldValue = this._max;
-    this._max = value <= this.min ? this.min + this.step : value;
-    this.requestUpdate('max', oldValue);
+  @property({ type: Number }) min?: number;
+  @property({ type: Number }) max?: number;
+  @property({ type: Number }) minValue?: number;
+  @property({ type: Number }) maxValue?: number;
+  @property({ reflect: true, type: Boolean }) invalid = false;
```
tolerants: It's a new API, so most probably we will need to change PR title to `feat`.

libs/base/ui/form/multi-range/src/multi-range.styles.ts
```diff
@@ -31,7 +32,12 @@ export const multiRangeStyles = css`
     margin: 9px 0;
   }
 
-  :host > * {
+  :host([invalid]) {
+    display: none;
```
tolerants: So we are just hiding the whole component in case something is invalid? I'd say it's against common behaviour for html components.

Probably we will need to align on our own strategy for such cases. I'll discuss it on AA catchup.

libs/base/ui/form/multi-range/src/multi-range.styles.ts
```diff
@@ -86,13 +99,15 @@ export const multiRangeStyles = css`
     box-shadow: 0 0 3px var(--oryx-color-primary-9);
   }
 
-  .active {
```
tolerants: It's a breaking change.

libs/base/ui/form/multi-range/src/multi-range.component.ts
```diff
@@ -13,142 +12,123 @@ export class MultiRangeComponent
...
+      ${this.renderRangeInput(this._max!)}
     `;
   }
 
-  protected renderRangeInput(
-    value: number,
-    inputRef: Ref,
-    isFirst = false
-  ): TemplateResult {
+  protected renderRangeInput(value: number, isFirst = false): TemplateResult {
```
tolerants: It's a breaking change, as long as we changed method signature.

libs/base/ui/form/multi-range/src/multi-range.component.ts
```diff
@@ -13,142 +12,123 @@ export class MultiRangeComponent
...
     return html`
       <label aria-label=${isFirst ? 'min' : 'max'}>
         <input
-          ref="${ref(inputRef)}"
```
tolerants, line 143: It's a breaking change, as long as ref is not accessible anymore.

libs/base/ui/form/multi-range/src/multi-range.component.ts
```diff
@@ -13,142 +12,123 @@ export class MultiRangeComponent
...
+    properties: PropertyValues<MultiRangeProperties>
+  ): boolean {
+    return Array.from(properties.entries()).some(([_key, value]) => {
+      const key = _key as keyof MultiRangeComponent;
+      return this.observedKeys.includes(key) && this[key] !== value;
+    });
   }
 
-  protected override render(): TemplateResult {
+  protected hasInvalidRange(): boolean {
```
tolerants, line 201: It looks like this method can have a better name as long as we are not only checking is range is invalid but also updating `invalid` property.

tolerants (changes requested):

libs/base/ui/form/multi-range/src/multi-range.component.ts
tobi-or-not-tobi: I'd not make it a `feat`. The purpose of building a change log is to help users understand new features. This is a tiny detail for a component that has not been used yet.

libs/base/ui/form/multi-range/src/multi-range.component.ts
tobi-or-not-tobi: Refactor is fine in my pov, as this component has not been used so far.

libs/base/ui/form/multi-range/src/multi-range.styles.ts
Sinucid: refactored

libs/base/ui/form/multi-range/src/multi-range.styles.ts
Sinucid: refactored

libs/base/ui/form/multi-range/src/multi-range.component.ts
Sinucid: refactored

libs/base/ui/form/multi-range/src/multi-range.component.ts
Sinucid: refactored

libs/base/ui/form/multi-range/src/multi-range.component.ts
Sinucid: refactored

tobi-or-not-tobi (approved):

tolerants (approved):

PR #826, opened 2023-10-16
feat: scroll facet values into view on toggle

tobi-or-not-tobi:
When the facet is collapsed, and the values are not visible, we scroll the facet values into the (nearest) view.

closes: [HRZ-90098](https://spryker.atlassian.net/browse/HRZ-90098)

[HRZ-90098]: https://spryker.atlassian.net/browse/HRZ-90098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/base/ui/structure/collapsible/src/collapsible.component.ts
+++ b/libs/base/ui/structure/collapsible/src/collapsible.component.ts
@@ -55,12 +55,12 @@ export class CollapsibleComponent
 
   protected onToggle(): void {
     this.open = this.details?.open;
-    if (this.isManuallyOpened && this.open) {
-      if (featureVersion >= '1.2') {
+    if (featureVersion >= '1.2') {
+      if (this.isManuallyOpened && this.open) {
         this.scrollIntoView({ behavior: 'smooth', block: 'nearest' });
       }
+      this.isManuallyOpened = false;
     }
-    this.isManuallyOpened = false;
   }
 
   protected onClick(): void {
--- a/libs/base/ui/structure/collapsible/src/collapsible.component.ts
+++ b/libs/base/ui/structure/collapsible/src/collapsible.component.ts
@@ -1,6 +1,6 @@
 import { ButtonSize, ButtonType } from '@spryker-oryx/ui/button';
 import { IconTypes } from '@spryker-oryx/ui/icon';
-import { I18nMixin } from '@spryker-oryx/utilities';
+import { I18nMixin, featureVersion } from '@spryker-oryx/utilities';
 import { LitElement, TemplateResult, html } from 'lit';
 import { property, query } from 'lit/decorators.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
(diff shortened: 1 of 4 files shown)


libs/base/ui/structure/collapsible/src/collapsible.component.ts
```diff
@@ -23,12 +23,23 @@ export class CollapsibleComponent
 
   @query('details') protected details?: HTMLDetailsElement;
 
+  /**
+   * Indicates that the collapsible was opened by a user. We need this info since
+   * we need to scroll the collapsible into the view port whenever it is opened by
+   * the user. We do not like to do this when it's (initially) rendered `open`.
+   */
+  protected isManuallyOpened = false;
```
tolerants, line 31: Probably worth to add to the model.

libs/base/ui/structure/collapsible/src/collapsible.component.ts
```diff
@@ -44,6 +55,16 @@ export class CollapsibleComponent
...
     this.open = this.details?.open;
+    if (this.isManuallyOpened && this.open) {
+      if (featureVersion >= '1.2') {
+        this.scrollIntoView({ behavior: 'smooth', block: 'nearest' });
+      }
+    }
+    this.isManuallyOpened = false;
+  }
+
+  protected onClick(): void {
```
tolerants, line 66: What if customer, who started with `1.0`, extended this component and created a method `onClick` by himself. And now he wants to update to `1.2`.

libs/domain/search/facet-value-navigation/facet-value-navigation.styles.ts
```diff
@@ -25,7 +25,7 @@ export const facetValueNavigationStyles = css`
 
   slot:not([name]) + oryx-button {
     width: fit-content;
-    padding-block-end: 10px;
+    padding-block-start: 10px;
```
tolerants: Change of ui can be a breaking change :(

tolerants (changes requested):

libs/base/ui/structure/collapsible/src/collapsible.component.ts
tobi-or-not-tobi: A general concern, we could discuss during catchup. We do not have a strategy to prevent this. I cannot recall an industry standard to prepare for this case. I can imagine we move way from the binding we do in our handler names. In this case we could make it `onOpen`, but not too sure.

libs/domain/search/facet-value-navigation/facet-value-navigation.styles.ts
tobi-or-not-tobi: True. But this is a small bug, hence the change without warning.

tobi-or-not-tobi: but yeah shortcut, fixed

libs/base/ui/structure/collapsible/src/collapsible.component.ts
tolerants: Yeah, understand and probably worth to discuss and think how to deal with such cases.

tolerants (dismissed):

libs/base/ui/structure/collapsible/src/collapsible.component.ts
```diff
@@ -44,6 +55,16 @@ export class CollapsibleComponent
 
   protected onToggle(): void {
     this.open = this.details?.open;
+    if (this.isManuallyOpened && this.open) {
+      if (featureVersion >= '1.2') {
```
tolerants: Maybe it make sense to check version first? `isManuallyOpened` doesn't even exist in earlier versions.
```suggestion
    if (featureVersion >= '1.2') {
      if (this.isManuallyOpened && this.open) {
```

tolerants (changes requested):

tolerants (approved):

PR #835, opened 2023-10-20
chore: add var for FA e2e tests

tarashaliuk:
Added env variable for Fulfillment e2e tests according to BE changes


apps/fulfillment-e2e/.env.template
```diff
@@ -1 +1,2 @@
-ORYX_FULFILLMENT_BACKEND_URL=
\ No newline at end of file
+ORYX_FULFILLMENT_BACKEND_URL=
```
tolerants, line 1: Can you please add description for variables.

apps/fulfillment/.env.template
```diff
@@ -1,3 +1,3 @@
 ORYX_FULFILLMENT_BACKEND_URL=http://localhost:3001
 ORYX_LABS=true
-ORYX_FULFILLMENT_CLIENT_ID=frontend
+ORYX_FULFILLMENT_CLIENT_ID=fulfillment-client-local
```
tolerants, line 3: Can you please add description for variables.

tolerants (changes requested):

apps/fulfillment-e2e/.env.template
tarashaliuk: description is in readme file

tolerants (approved):

PR #837, opened 2023-10-23
refactor(presets): clean fa theme

tobi-or-not-tobi:
We like to improve the fulfilment app theming to leverage standard theming options for the max width of the application. 

- [x] Move `backoffice-ng` design-token content into `backoffice`.
- [x] Remove `backoffice-ng` design-token. 
- [x] Create small-backoffice design-token set.
- [x] FA theme should use 2 breakpoints and layout should behave differently for these two:
  - 530px - layout should take the whole width
  - 768px - layout is fixed (530px) and centred
- [x] Search box: fix float behaviour - for now the searchbox takes almost whole width of the view port
- [x] Modal: fix width for fulfillment
- [x] Add experience pages:
  - [x] Login Page 
  - [x] Picking lists Page
  - [x] Warehouse Selection

closes [HRZ-90261](https://spryker.atlassian.net/browse/HRZ-90261), [HRZ-90260](https://spryker.atlassian.net/browse/HRZ-90260)

[HRZ-90261]: https://spryker.atlassian.net/browse/HRZ-90261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-90260]: https://spryker.atlassian.net/browse/HRZ-90260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/themes/design-tokens/src/backoffice-ng/index.ts
```diff
@@ -22,30 +21,18 @@ export const backofficeNgTokens: DesignToken[] = [
       typographySmallTokens.typography
     ),
     ...layoutTokens,
-    layout: {
-      container: {
-        width: '414px',
-        bleed: '10px',
-      },
+    container: {
```
tolerants: I'm just thinking... is it ok we are using `backofficeNg` theme tokens in FA? Why not just `backoffice`?

libs/template/application/oryx-app/oryx-app.model.ts
```diff
@@ -1,11 +1,3 @@
-export interface OryxAppAttributes {
-  /**
-   * The visual type of Oryx application.
-   *
-   */
-  type?: OryxAppType;
-}
-
 /**
```
tolerants: We can remove the whole file, but I can do it.

libs/domain/picking/lists/lists.styles.ts
```diff
@@ -7,6 +7,16 @@ export const pickingListsComponentStyles = css`
     gap: 20px;
     min-height: 100vh;
     background-color: var(--oryx-color-neutral-3);
+
+    /* tmp, we rather use pages with layout */
```
tolerants: But not sure we can easily change it.

libs/template/themes/src/fulfillment.theme.ts
```diff
@@ -1,11 +1,14 @@
...
+import { Size } from '@spryker-oryx/utilities';
 
 export const fulfillmentTheme: Theme = {
   name: 'fulfillment',
-  breakpoints: defaultBreakpoints,
+  breakpoints: {
+    [Size.Sm]: {},
+  },
   designTokens: () =>
+    // TODO: consider renaming the tokens
```
tolerants: Let's sync regarding renaming :)

libs/domain/picking/picker/picker.styles.ts
```diff
@@ -3,6 +3,17 @@ import { css } from 'lit';
 export const pickingComponentStyles = css`
   :host {
     --sticky-offset: 66px;
+
+    /* tmp, we rather use pages with layout */
```
tobi-or-not-tobi:
```suggestion
    /* TODO: use pages with layout */
```

libs/template/presets/fulfillment/experience/pages/login-page.ts
```diff
@@ -0,0 +1,37 @@
...
+      {
+        query: { breakpoint: Size.Sm },
+        width: '100%',
+      },
+      {
+        query: { breakpoint: Size.Md },
+        width: '414px',
+      },
+    ],
+  },
```
tobi-or-not-tobi: we should not need to repeat the width here, the container size should already provide this. Happy to pair program on this one if needed.

libs/template/presets/fulfillment/experience/pages/login-page.ts
```diff
@@ -0,0 +1,37 @@
...
+  components: [
+    {
+      type: 'oryx-picking-login',
+      options: {
+        rules: [
+          {
+            width: '100%',
+          },
+        ],
+      },
```
tobi-or-not-tobi: hopefully we can avoid this.

libs/template/themes/design-tokens/src/mobile-backoffice/index.ts
```diff
@@ -22,30 +20,19 @@ export const backofficeNgTokens: DesignToken[] = [
...
-      screen: Size.Lg,
-    },
-    ...layoutTokens,
-  },
-  {
-    media: {
-      screen: Size.Md,
+    modal: {
+      width: '404px',
+      bleed: '5px',
```
tobi-or-not-tobi: I was thinking to do this inside the modal component, so that any layout wil benefit from this, regardless of the breakpoint / container. It means that the max-width of the modal will be limited to the container size (minus bleed).

libs/template/presets/fulfillment/experience/pages/login-page.ts
```diff
@@ -0,0 +1,26 @@
...
+    route: '/login',
+    description: 'Login Page Description',
+  },
+  options: {
+    rules: [
+      { layout: 'flex', justify: 'center' },
+      {
+        query: { breakpoint: Size.Sm },
+        width: '100%',
+      },
```
tobi-or-not-tobi:
```suggestion
      { layout: 'list'},
```

tobi-or-not-tobi: Here's btw the inspiration to drop the picking login component:

```ts
options: {
    rules: [
      {
        layout: 'list',
        height: '100vh',
        gap: '40px',
        style: 'align-content: center',
      },
    ],
  },
  components: [
    {
      type: 'oryx-content-image',
      id: 'site-logo',
      content: {
        data: {
          graphic: 'logo',
          label: 'Fulfillment App based on Oryx',
        },
      },
      options: {
        rules: [{ height: '82px' }],
      },
    },
    {
      type: 'oryx-content-text',
      content: {
        data: {
          text: `<h3>${i18n('welcome-please-log-in-to-start-picking')}</h3>`,
        },
      },
    },
    {
      type: 'oryx-auth-login',
      options: {
        enableRememberMe: false,
      },
    },
  ],
```

libs/template/themes/src/fulfillment.theme.ts
```diff
@@ -1,13 +1,20 @@
...
 export const fulfillmentTheme: Theme = {
   name: 'fulfillment',
-  breakpoints: defaultBreakpoints,
+  breakpoints: {
+    [Size.Sm]: {
+      max: 414,
+    },
+    [Size.Md]: {
+      min: 415,
+    },
```
tobi-or-not-tobi: we should use:

```ts
[Size.Sm]: {
    max: 767,
  },
  [Size.Md]: {
    min: 768,
  },
```

And add a container to the larger breakpoint of 414px. This supposed to ensure that if your on a small screen, you will not see bars next to the 414px container size.

libs/template/presets/fulfillment/experience/pages/login-page.ts
tolerants: Looks great, but we decided to finish it in separate ticket.

PR #838, opened 2023-10-23
chore(fulfillment): drop mockend

Sinucid:
closes: [HRZ-90200](https://spryker.atlassian.net/browse/HRZ-90200)

[HRZ-90200]: https://spryker.atlassian.net/browse/HRZ-90200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/fulfillment/.env.template
```diff
@@ -1,3 +1,2 @@
-ORYX_FULFILLMENT_BACKEND_URL=http://localhost:3001
 ORYX_LABS=true
```
tobi-or-not-tobi: I'd drop labs from the template, we don't use it and dont' have anything in labs for picking afaik.

tobi-or-not-tobi (dismissed):

libs/template/presets/fulfillment/app.ts
```diff
@@ -24,6 +27,16 @@ export function fulfillmentFeatures(
...
+    siteFeature,
+    formFeature,
+    {
+      //drop PageMetaResolver from experienceFeature
+      //to exclude unnecessary functionality from SPA
+      providers: experienceFeature.providers?.filter(
+        (feature) => ![PageMetaResolver].includes(feature.provide)
+      ),
+      components: experienceFeature.components,
+    },
```
supproduction, line 39: Is it still relevant?

supproduction (dismissed):

libs/template/presets/fulfillment/app.ts
Sinucid: Due to FA still not use static experience data for its pages (components instead) and PageMetaResolver is not satisfied with this

tobi-or-not-tobi: also we'll use it shortly for the work Alexander is doing

tobi-or-not-tobi (approved):

PR #847, opened 2023-10-26
feat(search): range facet component

Sinucid:
The multi-range facet component is used in the facet navigation to render range facets. Range facets can be controlled by a minimum and maximum value. The user can use either an input field for those values or a multi-slider navigation. 

When there are more specific facet components configured for the multi-range facet, those components are used instead. For example, the price or rating facet have more specific components as they have alternative requirements the standard facet range component.

A few details on the implementation:
- The `MultiRangeChangeEvent` interface is introduced for the select event of multi-range component 
- The facet's interfaces has a new type field ('single', 'multi', 'range'). Facet interface was converted to the type and split on ValueFacet and RangeFacet interfaces
- The `multiValued` field of the Facet interface has been deprecated since version 1.2. Use type === 'multi' check instead.
- The `SearchFacetValueNavigationComponent` has now a `dirty` property an indication that facet has selected values.
- The `FacetController` has now support for the range facet.
- The `enableClear` property of the `SearchFacetComponent` has been deprecated since version 1.2. Use similar `disableClear` property instead.

closes: [HRZ-1945](https://spryker.atlassian.net/browse/HRZ-1945)

[HRZ-1945]: https://spryker.atlassian.net/browse/HRZ-1945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/search/facet-value-navigation/facet-value-navigation.component.ts
```diff
@@ -26,11 +26,12 @@ export class SearchFacetValueNavigationComponent
   @property({ type: Boolean }) enableSearch?: boolean;
   @property({ type: Boolean }) open?: boolean;
   @property({ type: Boolean }) enableClear?: boolean;
+  @property({ type: Boolean }) dirty?: boolean;
 
   @state() protected expanded = false;
 
   protected override render(): TemplateResult {
-    const allowClear = this.enableClear && this.selectedLength;
+    const allowClear = this.enableClear && this.dirty;
```
tobi-or-not-tobi: This is a BC

libs/domain/search/src/renderers/facet.providers.ts
```diff
@@ -24,22 +25,31 @@ export const facetProviders = [
...
+
           return html`
             <oryx-search-facet
               @oryx.select=${selectListener}
               .name=${facet.name}
               .renderLimit=${options.renderLimit}
               .open=${options.open}
-              .enableClear="${options.enableClear}"
-              .multi=${facet.multiValued}
+              ?disableClear="${!options.enableClear}"
```
tobi-or-not-tobi, line 43: BC; we should also add the former field for as long as we're < 1.2.

`.enableClear=${ifDefined(featureVersion >= '1.2' ? undefined : options.enableClear}}`

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #849, opened 2023-10-27
feat(search): price facet component

Sinucid:
The price facet component brings a specialized UI for filtering products by price. It extends the RangeFacetComponent with a few small additions:
- the price currency symbol is rendered on the min/max labels
- the price value is converted to a localized price

The current site currency is used during the conversion of the price symbol and value.

Various changes where made to existing components and services:
- The currency service has a new method `getCurrencySymbol`
- Range facet support for the `SearchFacetNavigationComponent`
- Performance improvements for the `SearchFacetNavigationComponent`
- Consistent event behavior of the `MultiRangeComponent` so that the search results are only refreshed when the user stop dragging
- Fixed calculation of range values (`MultiRangeComponent`) when several properties are applied simultaneously
- Improved performance of the `MultiRangeComponent` by decreasing re-rendering
- `MultiRangeComponent` is emitting `drag` event that represents current min and max values

closes: [HRZ-2848](https://spryker.atlassian.net/browse/HRZ-2848)

[HRZ-2848]: https://spryker.atlassian.net/browse/HRZ-2848?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/form/multi-range/src/multi-range.component.ts
```diff
@@ -1,12 +1,17 @@
...
 import { createRef, Ref, ref } from 'lit/directives/ref.js';
+import { when } from 'lit/directives/when.js';
 import {
   MultiRangeChangeEvent,
   MultiRangeProperties,
 } from './multi-range.model';
 import { multiRangeStyles } from './multi-range.styles';
 
+const defaultMin = 0;
+const defaultMax = 100;
```
tobi-or-not-tobi, line 15: do we need those defaults? I can not easily see how they can be beneficial, as the values are so depending on the data.

libs/base/ui/form/multi-range/src/multi-range.component.ts
```diff
@@ -16,119 +21,214 @@ export class MultiRangeComponent
...
-  protected override render(): TemplateResult {
+  protected willUpdate(properties: PropertyValues<MultiRangeProperties>): void {
+    if (featureVersion >= '1.2') {
+      this.ensureValues(properties);
+    }
+
+    super.willUpdate(properties);
+  }
+
+  protected ensureValues(
```
tobi-or-not-tobi:
```suggestion
  protected updateValues(
```

libs/base/ui/form/multi-range/src/multi-range.component.ts
```diff
@@ -16,119 +21,214 @@ export class MultiRangeComponent
...
+  protected observedKeys: (keyof MultiRangeComponent)[] = [
+    'step',
+    'min',
+    'max',
+    'minValue',
+    'maxValue',
+  ];
+
+  protected _activeMin?: number;
+  protected _activeMax?: number;
```
tobi-or-not-tobi: we've never formalized the usage of the `_` prefix. I do not like it that much, and it's also more used for private members. Can we rename them? 

Maybe `minState`?

libs/base/ui/form/multi-range/src/multi-range.component.ts
```diff
@@ -157,17 +260,58 @@ export class MultiRangeComponent
...
     }
   }
+
+  protected onSelect(): void {
+    if (featureVersion < '1.2') return;
+
+    const minValue = (this.minValue = this._activeMin!);
+    const maxValue = (this.maxValue = this._activeMax!);
+
+    this.dispatchSelectEvent(minValue, maxValue);
```
tobi-or-not-tobi:
```suggestion
    this.minValue = this._activeMin!;
    this.maxValue = this._activeMax!;

    this.dispatchSelectEvent(this.minValue, this.maxValue);
```

tobi-or-not-tobi (changes requested):

libs/base/ui/form/multi-range/src/multi-range.component.ts
Sinucid: yes, they are used in several places and were created to increase transparency of the code

libs/base/ui/form/multi-range/src/multi-range.component.ts
Sinucid: lets use **activeMin** and **activeMax** without prefix

libs/base/ui/form/multi-range/src/multi-range.component.ts
Sinucid: This won't work, due to 'active' values just properties, they synchronised with inputs and are not reactive. They are just representing the intermediate state for the calculations that is not triggering the render and life-circle hooks. Creating the **let** minValue and maxValue is doing the same, decrease amount of renders and due to logic in this method contains some normalisation of the values, this allows to avoid unnecessary re-renders and greatly increase the performance

tobi-or-not-tobi (dismissed):

tobi-or-not-tobi (approved):

PR #853, opened 2023-10-31
chore: change plugins to async providing

supproduction:
closes [HRZ-90298](https://spryker.atlassian.net/browse/HRZ-90298)

[HRZ-90298]: https://spryker.atlassian.net/browse/HRZ-90298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/src/controllers/layout.controller.ts
```diff
@@ -5,7 +5,15 @@ import {
 } from '@spryker-oryx/experience/layout';
 import { Size, featureVersion, sizes } from '@spryker-oryx/utilities';
 import { LitElement, TemplateResult, html } from 'lit';
-import { Observable, map, withLatestFrom } from 'rxjs';
+import {
```
tobi-or-not-tobi, line 8: There's a featureVersion usage in this file that conflicts with https://spryker.slack.com/archives/C029XPSBXC6/p1697630001287519. Can you make it consistent?

libs/platform/experience/src/mixins/layout.mixin.ts
```diff
@@ -63,19 +69,22 @@ export declare class LayoutMixinInterface {
...
-  protected getLayoutRender(
-    place: keyof LayoutPluginRender,
-    data: LayoutControllerRender
-  ): TemplateResult;
+  protected $preLayoutRenderComposition: ConnectableSignal<
+    Record<string, TemplateResult> | undefined
+  >;
+  protected $postLayoutRenderComposition: ConnectableSignal<
+    Record<string, TemplateResult> | undefined
+  >;
```
tobi-or-not-tobi: Are these only used by compositions? Or also intended to be used inside other components, such as product list?

tobi-or-not-tobi: i'm not stoked by the names eitherway, any alternatives we could consider?

libs/platform/experience/src/services/layout/plugins/layout.plugin.ts
```diff
@@ -9,10 +9,38 @@ import {
...
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  export interface Layouts {}
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  export interface LayoutProperty {}
+
+  export interface LayoutStylesProperties {
+    layout?: LayoutStylesOptions | LayoutTypes;
+
+    // @deprecated since 1.2 will be removed.
```
tobi-or-not-tobi:
```suggestion
    // @deprecated since 1.2 will be removed, use the ... plugin instead
```

also in other notes below

libs/platform/experience/src/services/layout/plugins/properties/bleed/index.ts
```diff
@@ -1,2 +1 @@
 export * from './bleed-layout.model';
```
tobi-or-not-tobi, line 1:
```suggestion
/** 
 * Do NOT export any plugin files, as we'd break lazy loading otherwise. 
 * All plugin artefacts are publicly available in a static entry point, to allow for customisations.
 */
export * from './bleed-layout.model';
```

Or something like that. Can we do that in all those plugins, or if you think we can do it in a central place it might also work nice.

tobi-or-not-tobi (changes requested):

libs/platform/experience/src/mixins/layout.mixin.ts
supproduction: it's only for composition (we need it only for composition-component)

libs/platform/experience/src/services/layout/plugins/properties/bleed/index.ts
supproduction: added it here
`layout.static.ts`

libs/platform/experience/src/mixins/layout.mixin.ts
dunqan: it is for composition per se, or for cases with multiple elements in general? 
if it's only for composition, maybe we could extract it to another (sth like layout compostion) mixin, to separate concerns?

tobi-or-not-tobi: I was also thinking on additional mixin in this case

libs/platform/experience/src/services/layout/plugins/properties/bleed/index.ts
tobi-or-not-tobi: What about
- moving the note to `static.ts`
- introduce `static.ts` barrel files that we use to export includes to above layers. This will also make it more apparent on what we're doing

tobi-or-not-tobi (approved):

PR #855, opened 2023-10-31
feat: storage service strategies

dunqan:
- customizable storage service through storage strategies
- deprecate indexed DB storage service in core + refactor to drop dependency on 3rd-party library
- add IndexedDBStorage service to indexed-db library

Closes [HRZ-90051](https://spryker.atlassian.net/browse/HRZ-90051)

[HRZ-90051]: https://spryker.atlassian.net/browse/HRZ-90051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/core/src/services/storage/default-indexed-db-storage.service.ts
+++ b/libs/platform/core/src/services/storage/default-indexed-db-storage.service.ts
@@ -1,52 +1,73 @@
-import { Dexie, liveQuery } from 'dexie';
-import { Observable, shareReplay, switchMap } from 'rxjs';
+import { subscribeReplay } from '@spryker-oryx/utilities';
+import { map, Observable, shareReplay, Subscriber, switchMap } from 'rxjs';
 import {
   indexedDbStorageName,
   IndexedDBStorageService,
   indexedDbTableName,
 } from './indexed-db-storage.service';
-import { StoredValue } from './model';
 
+/** @deprecated since 1.2 */
 export class DefaultIndexedDBStorageService implements IndexedDBStorageService {
-  protected storage = new Observable<Dexie>((subscriber) => {
-    const db = new Dexie(indexedDbStorageName);
-    db.version(1).stores({
-      [indexedDbTableName]: '&key,value',
-    });
+  private db$: Observable<IDBDatabase> = new Observable(
+    (subscriber: Subscriber<IDBDatabase>) => {
+      const request = indexedDB.open(indexedDbStorageName, 1);
 
-    subscriber.next(db);
+      request.onerror = () => subscriber.error('DB: error');
+      request.onupgradeneeded = (evt: IDBVersionChangeEvent) => {
+        const db = (evt.target as IDBOpenDBRequest).result;
+        db.createObjectStore(indexedDbTableName, { keyPath: 'key' });
+      };
+      request.onsuccess = () => {
+        subscriber.next(request.result);
+        subscriber.complete();
+      };
+    }
+  ).pipe(shareReplay(1));
 
-    return () => {
-      db.close();
-    };
-  }).pipe(
-    switchMap(async (db) => {
-      await db.open();
+  private transactionRequest<T>(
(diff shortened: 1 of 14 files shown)


libs/platform/core/src/services/storage/default-indexed-db-storage.service.ts
```diff
@@ -1,52 +1,73 @@
...
-    this.storage.subscribe((storage) => storage.delete(key));
+  setItem(key: string, value: string): Observable<unknown> {
+    return subscribeReplay(
+      this.transactionRequest('readwrite', (store) => store.put({ key, value }))
+    );
   }
 
-  clear(): void {
-    this.storage.subscribe((storage) => storage.clear());
+  removeItem(key: string): Observable<unknown> {
```
Sinucid, line 62: Should we return something when delete or clear?  Prev interfaces were without return values

libs/platform/core/src/services/storage/default-storage.service.ts
```diff
@@ -1,15 +1,29 @@
-import { inject } from '@spryker-oryx/di';
+import { inject, INJECTOR } from '@spryker-oryx/di';
 import { isObservable, map, Observable, of } from 'rxjs';
 import { IndexedDBStorageService } from './indexed-db-storage.service';
-import { StorageType, StoredValue } from './model';
-import { StorageService } from './storage.service';
+import { StorageType } from './model';
+import { StorageService, StorageStrategy } from './storage.service';
+
+function toObservable<T>(value: Observable<T> | T): Observable<T> {
```
Sinucid: Maybe better to move it to utilities?

libs/platform/core/src/services/storage/default-indexed-db-storage.service.ts
dunqan: It's deprecated, but I wanted to make it more correct, if someone will use it anyway. I assumed, it's not a breaking change, as customer doesn't expect return anyway.
Also previous implementation was not so correct, as all methods (remove, clear, set) where asynchronous, so it was not possible to wait for them to finish properly.

libs/platform/core/src/services/storage/default-storage.service.ts
dunqan: 👍

dunqan: actually we have this utility already 🤦 ;)

Sinucid (dismissed):

libs/platform/core/src/services/storage/model.ts
```diff
@@ -1,7 +1,9 @@
 export const enum StorageType {
   Local = 'local',
   Session = 'session',
+  /** @deprecated since 1.2 */
```
tolerants: Suggest to add an alternative in deprecation.

libs/platform/core/src/services/storage/default-indexed-db-storage.service.ts
```diff
@@ -1,52 +1,73 @@
...
-      switchMap((storage) =>
-        liveQuery<StoredValue>(async () => {
-          return (await storage.get(key))?.value;
-        })
+  protected db$: Observable<IDBDatabase> = new Observable(
+    (subscriber: Subscriber<IDBDatabase>) => {
+      const request = indexedDB.open(indexedDbStorageName, 1);
+
+      request.onerror = () => subscriber.error('DB: error');
+      request.onupgradeneeded = (evt: IDBVersionChangeEvent) => {
```
tolerants: I think we don't need to save 2 letters :) Was not able to find in our code base such an abbreviation.
```suggestion
      request.onupgradeneeded = (event: IDBVersionChangeEvent) => {
```

tolerants (changes requested):

tolerants (approved):

PR #856, opened 2023-11-01
feat(experience): carousel navigation

tobi-or-not-tobi:
The carousel navigation can be used in combination with the carousel layout. The layout can be used with a combination of items from the light DOM or shadow DOM, either direct children or provided by nested components (e.g. product list component).

The carousel navigation supports a number of features: 
- Dynamic calculation of slides.
- Navigation by item or by slide.
- (optional) arrow controls to navigate to the previous or next item or slide. The controls disappear when they're not applicable at the carousel state, e.g. when the first slide is active, the previous control will disappear.
- (optional) slide indicator controls, allow you to navigate to individual slides.
- Highlight indicators (gradually) while scrolling the carousel. Both when manually scrolling as well as when navigating by the controls.
- (optional) allow for scrolling through the carousel by mouse.
- (optional) allow for scrolling through the carousel by touch.
- Instant vs smooth sliding behaviour.
- Indicator horizontal alignment: start, end, and center.
- Indicator vertical alignment: below, bottom.
- Vertical and horizontal carousels

closes: [HRZ-90099](https://spryker.atlassian.net/browse/HRZ-90099)

[HRZ-90099]: https://spryker.atlassian.net/browse/HRZ-90099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/product/images/src/images.component.ts
```diff
@@ -85,8 +89,18 @@ export class ProductImagesComponent extends ProductMixin(
...
     return html`<oryx-layout
       class="main"
-      layout=${ifDefined(imageLayout)}
       behavior=${ifDefined(scrollBehavior)}
+      layout=${ifDefined(imageLayout)}
+      .options=${{
+        rules: [
+          {
+            layout: {
+              showArrows: false,
```
Sinucid: should we pass `false` here?

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+    this.intersectionObserver.observe(this.hostElement);
+  }
+
+  /**
+   * Initializes the ResizeObserver to detect when the carousel's width changes.
+   * This is necessary to recalculate the carousel's width and build the navigation.
+   */
+  protected initializeResizeObserver(): void {
+    if (this.resizeObserver) return;
+    const throttleTime = this.hostElement.clientWidth === 0 ? 0 : 150;
```
Sinucid: I would store the default throttle in prop or const, ie:
`const throttleTime = this.hostElement.clientWidth ? defaultThrottleTime: 0;`

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+    this.resizeObserver = new ResizeObserver(
+      throttle(() => {
+        this.buildNavigation();
+      }, throttleTime)
+    );
+    this.resizeObserver.observe(this.hostElement);
+  }
+
+  protected initializeScrollListener(): void {
+    // if (this.scrollBehavior === CarouselScrollBehavior.Smooth) {
```
Sinucid: leftover?

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+    this.throttledScrollListener = throttle(() => {
+      this.updateState();
+    }, 50);
+    this.hostElement.addEventListener('scroll', this.throttledScrollListener);
+    // }
+    this.hostElement.addEventListener('scrollend', this.updateState.bind(this));
```
Sinucid: bind creates a new func entity, this won't work when you will try to removeEvent listener, the func should be the same

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+    this.intersectionObserver?.disconnect();
+    if (this.hostElement) {
+      this.hostElement.removeEventListener(
+        'scroll',
+        this.throttledScrollListener
+      );
+      this.hostElement.removeEventListener('scrollend', this.updateState);
+    }
+    super.disconnectedCallback();
+  }
```
Sinucid, line 76: seems to be that subscription logic can be moved to the controller, to decrease complexity

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+  @property({ reflect: true }) indicatorsAlignment?: CarouselIndicatorAlignment;
+  @property() scrollBehavior?: CarouselScrollBehavior;
+
+  protected resizeObserver?: ResizeObserver;
+  protected intersectionObserver?: IntersectionObserver;
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  protected throttledScrollListener: () => void = () => {};
+
+  @state() protected items: HTMLElement[] = [];
+  @state() protected slides: { index: number }[] = [];
```
Sinucid: Why not just `number[]`?

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+  /**
+   * Initializes the IntersectionObserver to detect when the carousel becomes visible.
+   * This is necessary to calculate the carousel's width and to build the navigation.
+   */
+  protected initializeIntersectionObserver(): void {
+    if (this.intersectionObserver) return;
+
+    this.intersectionObserver = new IntersectionObserver(
+      throttle((entries: IntersectionObserverEntry[]) => {
+        return entries.forEach((entry) => {
```
Sinucid, line 87: I didn't get, why do you need to cal `buildNavigation`,  `initializeResizeObserver` and the rest inside foreach? It will call this actions multiple times according the amount of entries

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+          if (entry.isIntersecting) {
+            if (this.resolveItems().length) {
+              this.buildNavigation();
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+              this.intersectionObserver!.disconnect();
```
Sinucid: Why do you need to disconnect and then re-initialize the observer?

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+      const percentage = distance / this.hostElement.clientWidth;
+      let opacity = percentage >= 0 ? 1 - percentage : 1 + percentage;
+      if (clientWidth + scrollLeft >= scrollWidth) {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        if (index === this.indicatorElements!.length - 1) {
+          // we could consider making the previous slide indicator to 0
+          // (opacity = index === this.indicatorElements!.length - 1 ? 1 : 0;)
+          opacity = 1;
+        }
+      }
```
Sinucid:
```suggestion
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
      if (clientWidth + scrollLeft >= scrollWidth && index === this.indicatorElements!.length - 1) opacity = 1;
```

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+   *
+   * Supports alt-key, so that users can navigate to the first slide in 1 click.
+   */
+  protected handleNext(e: Event): void {
+    let index;
+
+    if ((e as KeyboardEvent).altKey) {
+      index = this.items.length - 1;
+    } else {
+      const isRtl = window.getComputedStyle(this).direction === 'rtl';
```
Sinucid: I would even create a getter for this

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+
+      const { left, width } = this.hostElement.getBoundingClientRect();
+      if (isRtl) {
+        index = this.items.findIndex((el) => {
+          return el.getBoundingClientRect().left <= left;
+        });
+      } else {
+        const x =
+          this.arrowNavigationBehavior === ArrowNavigationBehavior.Item
+            ? left + 10
```
Sinucid: why 10?

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+    });
+  }
+
+  /**
+   * @override Renders the navigation arrow buttons (previous/next) and indicators.
+   */
+  protected override render(): TemplateResult[] {
+    const results: TemplateResult[] = [];
+
+    if (this.showArrows) {
```
Sinucid: I'm not sure that pushing template's to the array is a right and performanced way. Why not to use `when` directive?

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -0,0 +1,410 @@
...
+   * the source order, but it will come: https://developer.chrome.com/blog/reading-order/.
+   *
+   * @returns An array of HTMLElements representing the child elements of the host component.
+   */
+  protected resolveItems(): HTMLElement[] {
+    const query =
+      '*:not(:is(oryx-carousel-navigation,slot,style,.indicators,button,oryx-button, input)';
+
+    const items = [
+      ...(this.hostElement?.shadowRoot?.querySelectorAll(query) ?? []),
```
Sinucid: Are you just query all children in host's shadow dom? Can we narrow down the sample?

Sinucid (dismissed):

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: I do not expect performance issue, but I'll keep it aligned with all other render methods we implement.

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: no longer needed afaics

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: I've actually tried that, but it became a bit messy at the end. I/we could do another try.

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: fixed. call me lazy.

libs/domain/product/images/src/images.component.ts
tobi-or-not-tobi: yes. enabling the arrows on images takes another ticket as there are a few edge cases that did not work.

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: We're disconnecting the intersection observer, and then start observing other things, like scrolling and window resizing. Once the component has intersection, we initialize and kick of other observables.

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: That's how the intersection observer works. I've done a small optimisation to ensure that we only initialise once.

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: not sure. one case moved to external function.Please check again.

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: simplified. Slotted children are as simple as the element.children. Assigned elements are only direct elements.

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: won't do for now.

supproduction (dismissed):

dunqan (approved):

PR #859, opened 2023-11-02
refactor(search): rating facet component refactoring

Sinucid:
closes: [HRZ-90315](https://spryker.atlassian.net/browse/HRZ-90315)

[HRZ-90315]: https://spryker.atlassian.net/browse/HRZ-90315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/search/facet-rating/facet-rating.component.ts
```diff
@@ -1,84 +1,119 @@
...
+
+  protected override render(): TemplateResult | void {
+    const facet = this.$facet();
+
+    if (!facet) return;
+
+    return html`<oryx-search-facet-value-navigation
+      ?open=${this.open}
+      ?enableClear="${!this.disableClear}"
+      ?dirty=${!!this.$isDirty()}
```
tobi-or-not-tobi:
```suggestion
      ?dirty=${this.$isDirty()}
```

it's a boolean already afaics

libs/domain/search/src/renderers/value-renderer/rating.renderer.ts
```diff
@@ -14,12 +14,11 @@ export const ratingFacetRenderer = {
...
           @oryx.select=${selectListener}
           .name=${facet.name}
-          .renderLimit=${options.renderLimit}
           .open=${options.open}
           ?disableClear="${!options.enableClear}"
-          .multi=${facet.type === FacetType.Multi}
-        >
-        </oryx-search-facet-rating>
+          min="1"
+          max="4"
```
tobi-or-not-tobi: how come these are hardcoded?

tobi-or-not-tobi (changes requested):

libs/domain/search/src/renderers/value-renderer/rating.renderer.ts
Sinucid: I dropped default values for these props in component (to decrease amount of code and increase flexibility), facet's min and max are used by default, if values are not provided. The values are hardcoded in the renderer due to we don't have mechanism to pass per-component options on EB side for facet-navigation component, only the common set. 

For now the only way to override this hardcoded min and max - to override the renderer for the rating. But I agree, our current implementation of the facets at all are too limited

tobi-or-not-tobi: That's a bummer. In the former implementation, we could override the default values in the application configuration. The lack of EB is not so relevant as it's not a product that we ship anytime soon. Can we bring it back?

libs/domain/search/facet-price/facet-price.component.ts
```diff
@@ -41,6 +43,21 @@ export class SearchPriceFacetComponent extends I18nMixin(
...
+  protected onRangeChange(e: CustomEvent<MultiRangeChangeEvent>): void {
+    const { min, max } = this.$facet()!.values;
+    const { minValue, maxValue } = e.detail;
+    const selected = {
+      min: minValue !== min ? minValue : undefined,
+      max: maxValue !== max ? maxValue : undefined,
+    } as SelectRangeFacetValues;
+
+    if (this.hasChangedValue(selected)) {
+      console.log(this.$facet(), selected);
```
tobi-or-not-tobi: left over?

tobi-or-not-tobi (changes requested):

libs/domain/search/src/renderers/value-renderer/rating.renderer.ts
tobi-or-not-tobi: The idea was to allow customers to either come up with min=4 and max=4 (only showing 4-star, which is quit common) or allow to include 5

tobi-or-not-tobi (dismissed):

tobi-or-not-tobi (approved):

PR #864, opened 2023-11-06
test(storefront): search and category tests refactoring

Sinucid:
closes: [HRZ-90389](https://spryker.atlassian.net/browse/HRZ-90389)

[HRZ-90389]: https://spryker.atlassian.net/browse/HRZ-90389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront-e2e/src/integration/search.cy.ts
+++ b/apps/storefront-e2e/src/integration/search.cy.ts
@@ -6,76 +6,81 @@ import {
 import { SearchPage } from '../support/page-objects/search.page';
 import { sortingTestData } from '../support/test-data/search-products';
 
-let searchPage;
-
 describe('Search suite', () => {
+  let searchPage;
+
   describe('Products filtering', () => {
-    const query = 'Canon';
+    describe('by value facets', () => {
+      const query = 'Canon';
 
-    beforeEach(() => {
-      searchPage = new SearchPage({ q: query });
-      searchPage.visit();
+      beforeEach(() => {
+        searchPage = new SearchPage({ q: query });
+        searchPage.visit();
+      });
+
+      it('should update products and facets when filters are applied/cleared', () => {
+        searchPage.getFacets().setRating('4');
+        searchPage.waitForSearchRequest();
+        checkProductCardsFilteringByName(searchPage, 2, 1, query);
+
+        // we don't expect search request here because previous query is cached
+        searchPage.getFacets().resetRating();
+
+        // apply 1st filter
+        searchPage.getFacets().setLabel('New');
+        searchPage.waitForSearchRequest();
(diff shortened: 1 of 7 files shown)


apps/storefront-e2e/src/support/mixins/page-with-product-list.mixin.ts
```diff
@@ -24,11 +24,11 @@ export function WithProductList<TPage extends Constructor<E2EPage>>(
       cy.intercept('/catalog-search?**').as('catalogSearch');
     }
 
-    waitForSearchRequest(): void {
-      cy.wait('@catalogSearch');
+    waitForSearchRequest(skipRequest?: boolean): void {
+      if (!skipRequest) cy.wait('@catalogSearch');
```
AndreyYevtukhov: Let's create a separate method like `skipSearchRequest()` or `waitForProductsToLoad()` in this page object, because `waitForSearchRequest(true)` looks weird and is confusing

apps/storefront-e2e/src/integration/search.cy.ts
```diff
@@ -6,76 +6,81 @@ import {
...
+        cy.log('set minimum price');
+        searchPage.getFacets().setMinPrice(minPrice);
+        searchPage.waitForSearchRequest();
+        checkProductCardsFilteringByPrice(searchPage, minPrice);
+
+        cy.log('reset prices, set max price');
+        searchPage.getFacets().resetPrices();
+        searchPage.waitForSearchRequest(true);
+        searchPage.getFacets().setMaxPrice(maxPrice);
+        searchPage.waitForSearchRequest();
```
AndreyYevtukhov, line 67: I don't understand the logic behind skipping / not skipping the requests here - in some places we have it skipped, in other it still appears

Can you explain it, please? 

---
Screenshot 2023-11-07 at 08 28 03

AndreyYevtukhov: I also can't understand how skipping an API call wait increases test stability. 

In theory, it will make them unstable because API calls are random (visibility check might help, but it looks weird)

--
If it works - let's keep it and see where it goes

AndreyYevtukhov (dismissed):
Left some comments, please check

apps/storefront-e2e/src/integration/search.cy.ts
Sinucid: Because we are not making some requests (they have been already cached), but UI still need some time to re-build the template

supproduction (approved):

PR #869, opened 2023-11-07
chore(core): proper storage resolving for Service Worker

dunqan:


Sinucid (approved):

tolerants:
I believe it's a `chore` as long as we are fixing functionality introduced/adjusted which is not released yet.

tolerants (approved):

PR #870, opened 2023-11-07
fix(ui): modal maximum width to depend on container width

tolerants:
closes: [HRZ-90378](https://spryker.atlassian.net/browse/HRZ-90378)

[HRZ-90378]: https://spryker.atlassian.net/browse/HRZ-90378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/base/ui/overlays/modal/src/modal.styles.ts
+++ b/libs/base/ui/overlays/modal/src/modal.styles.ts
@@ -18,7 +18,7 @@ export const styles = css`
     );
     max-width: calc(
       var(--oryx-container-width, --oryx-modal-width) -
-      var(--oryx-modal-bleed, 0px) * 2
+        var(--oryx-modal-bleed, 0px) * 2
     );
     max-height: calc(100vh - var(--oryx-modal-bleed, 0px) * 2);
   }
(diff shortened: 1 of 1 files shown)


libs/base/ui/overlays/modal/src/modal.styles.ts
```diff
@@ -16,7 +16,10 @@ export const styles = css`
       calc(100vw - (var(--oryx-container-bleed, 0px) * 2)),
       var(--oryx-modal-width, 100%)
     );
-    max-width: calc(100vw - var(--oryx-modal-bleed, 0px) * 2);
+    max-width: calc(
+      var(--oryx-container-width, --oryx-modal-width) -
```
tobi-or-not-tobi:
```suggestion
      var(--oryx-container-width, var(--oryx-modal-width)) -
```

tobi-or-not-tobi: we might also need to add a hardcoded fallback maybe to not break the calc in case there's nothing

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #873, opened 2023-11-07
fix: layout callstack issue

supproduction:
fixes ssr callstack issue
fixes image centring on PDP


tobi-or-not-tobi:
closed, as it's already covered in carousel navigation

PR #874, opened 2023-11-08
feat: context serializers

dunqan:
Introducing mechanism of context serializers, that are useful for transforming complex contexts (like product qualifier) to readable an concise serialized string. 

Closes [HRZ-90222](https://spryker.atlassian.net/browse/HRZ-90222)

[HRZ-90222]: https://spryker.atlassian.net/browse/HRZ-90222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/product/src/mixins/product.mixin.ts
```diff
@@ -34,13 +36,19 @@ export const ProductMixin = <
...
+          map((sku) =>
+            featureVersion >= '1.3' ? sku : ({ sku } as ProductQualifier)
+          )
+        )
     );
 
     protected $product = computed(() => {
-      const sku = this.sku ?? this.$productContext();
-      return sku ? this.productService?.get({ sku }) : of(null);
+      const qualifier = this.sku ? { sku: this.sku } : this.$productQualifier();
```
tobi-or-not-tobi, line 49: BC

libs/domain/product/src/mixins/product.mixin.ts
```diff
@@ -19,6 +20,7 @@ export declare class ProductMixinInterface
 {
   sku?: string;
   protected $product: Signal<Product | null>;
+  protected $productQualifier: Signal<ProductQualifier | null>;
```
tobi-or-not-tobi:
```suggestion

  /**
   * A signal representing the product qualifier.
   * If the feature version is 1.3 or higher, it's an object containing SKU and offer.
   * Prior to version 1.3, it's a string representing SKU only.
   */
  protected $productQualifier: Signal<ProductQualifier | null>;
```

Also, the interface seems not aligned (`undefined` vs `null`)

tobi-or-not-tobi (changes requested):

libs/domain/product/src/mixins/product.mixin.ts
dunqan: actually it's aways qualifier, regardless the feature flag, as we are normalizing it inside, depending if context is a string (pre 1.3) or qualifier.

tobi-or-not-tobi (dismissed):

libs/domain/product/src/mixins/product.mixin.ts
dunqan: we can discuss, if we should not remove `protected $productContext`, but because it was not exposed in an mixin interface, I decided to treat it as a private API

libs/platform/core/src/services/context/default-context.service.ts
```diff
@@ -188,4 +188,37 @@ export class DefaultContextService implements ContextService {
...
   }
+
+  private getSerializer<T>(key: string): ContextSerializer<T> | null {
+    if (!this.serializers.has(key)) {
+      this.serializers.set(
+        key,
+        this.injector.inject(`${ContextSerializer}${key}`, null)
+      );
+    }
+    return this.serializers.get(key)! as ContextSerializer<T> | null;
```
supproduction: do we still need `!` for maps  with ts 5?

supproduction (dismissed):

libs/platform/core/src/services/context/default-context.service.ts
dunqan: 👍

tobi-or-not-tobi (dismissed):

supproduction (approved):

PR #880, opened 2023-11-09
fix: 404 simple pages

supproduction:
closes [HRZ-89741](https://spryker.atlassian.net/browse/HRZ-89741)

[HRZ-89741]: https://spryker.atlassian.net/browse/HRZ-89741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/experience/src/routes.ts
+++ b/libs/platform/experience/src/routes.ts
@@ -43,13 +43,17 @@ export const defaultExperienceRoutes: RouteConfig[] = [
   {
     path: '/:page',
     type: RouteType.Page,
-    afterEnter: ({ page }) =>
-      resolve(ExperienceService)
-        .getComponent({ route: `/${page}` })
-        .pipe(
-          take(1),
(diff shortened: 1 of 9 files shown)


tobi-or-not-tobi (dismissed):

libs/platform/experience/src/routes.ts
```diff
@@ -42,6 +43,13 @@ export const defaultExperienceRoutes: RouteConfig[] = [
   {
     path: '/:page',
     type: RouteType.Page,
+    afterEnter: ({ page }) =>
```
dunqan: Just wondering, do we need `afterEnter` tied to router configuration, if we can just provide router event listener instead. Feels more decoupled, and easier to override/extend.

dunqan (approved):

PR #882, opened 2023-11-09
feat(experience): header category navigation

tobi-or-not-tobi:
The header navigation contains some of the categories available in the assortment, which differ from b2c to b2b. The category is build with standard components, such compositions and link components, and the layout is provided by a new navigation layout. The navigation layout supports both horizontal or vertical layouts.

The navigation layouts adds a number of (opinionated) CSS variables that are used in the styles of the descendants content link components:
- `--oryx-link-icon-size`
- `--oryx-link-width`
- `--oryx-link-color`
- `--oryx-link-decoration`
- `--oryx-link-padding`
- `--oryx-link-hover-shadow`
- `--oryx-link-active-shadow`
- `--oryx-link-current-shadow`

closes: [HRZ-2519](https://spryker.atlassian.net/browse/HRZ-2519)

[HRZ-2519]: https://spryker.atlassian.net/browse/HRZ-2519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/search/box/box.component.spec.ts
```diff
@@ -93,7 +93,9 @@ describe('SearchBoxComponent', () => {
         },
       ],
     });
-    linkService = testInjector.inject(LinkService) as MockSemanticLinkService;
+    linkService = testInjector.inject(
+      LinkService
+    ) as unknown as MockSemanticLinkService;
```
Sinucid:
```suggestion
    linkService = testInjector.inject<MockSemanticLinkService>(
      LinkService
    );
```

Sinucid (changes requested):

libs/platform/experience/src/services/layout/plugins/types/navigation/navigation-layout.plugin.ts
```diff
@@ -0,0 +1,30 @@
...
+  LayoutPlugin,
+  LayoutPluginConfig,
+  LayoutPluginOptionsParams,
+} from '../../layout.plugin';
+
+export class NavigationLayoutPlugin implements LayoutPlugin {
+  getStyles(data: LayoutPluginOptionsParams): Observable<LayoutStyles> {
+    const { options } = data;
+
+    return ssrAwaiter(
```
tolerants, line 14: Is it a right usage of `ssrAwaiter`? I mean, why should layout know something about ssr?

tolerants (changes requested):

libs/platform/experience/src/services/layout/plugins/types/navigation/navigation-layout.plugin.ts
tobi-or-not-tobi: yup, this is standard approach we took in the plugins to ensure we can lazy load styles.

tolerants (approved):

Sinucid (approved):

PR #886, opened 2023-11-10
chore(fulfillment): split offline features

Sinucid:
Splitting of offline-related features into client and service-worker related. This allowed better managing the dependancies and decrease size of  the chunks. As a part of the splitting were done:

- fulfillment application no more supports labs feature
- fulfillment app vite PWA configuration was adjusted to decrease amount of cached files by service-worker 
- service-worker related functionality was extracted from `fulfilment` preset into the separate `fulfilment-sw` preset
- platform offline synchronisation related functionality (dexie entities, models, services) was extracted into `@spryker-oryx/offline/sync` sub-package
- platform offline service-worker related functionality (feature, plugins) was extracted into `@spryker-oryx/offline/service-worker` sub-package
- picking api related functionality (feature, models, adapters, services) was extracted into `@spryker-oryx/picking/api` sub-package
- picking service-worker related functionality (feature, services) was extracted into `@spryker-oryx/picking/service-worker` sub-package
- separate limited app builder was added to the `@spryker-oryx/application/service-worker` that is not including template related parts
- quantity input component was moved from cart domain to the ui library with changing the tag (`oryx-quantity-input`). `oryx-cart-quantity-input` component was remained for backward compatibility. Currently it just export resources from ui one
- dropped definition files exports from main entry points of some ui components (toggle, toggle-icon, checkbox, error-message, input-list, option, tag, rating, spinner, navigation-item, navigation, page-navigation-item, page-navigation, tab, tabs, drawer, dropdown, modal, popover, tile ). Currently component's definitions can be imported from the main library entry point (`@spryker-oryx/ui`)

closes: [HRZ-90277](https://spryker.atlassian.net/browse/HRZ-90277)

[HRZ-90277]: https://spryker.atlassian.net/browse/HRZ-90277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/cart/quantity-input/index.ts
```diff
@@ -1 +1,2 @@
+/** @deprecated since 1.2. Use quantity input from @spryker-oryx/ui/quantity-input*/
```
tolerants: Please check everywhere.
```suggestion
/** @deprecated since 1.2. Use oryx-quantity-input from @spryker-oryx/ui/quantity-input */
```

apps/fulfillment/vite.config.ts
```diff
@@ -13,6 +13,19 @@ export default defineConfig({
     outDir: '../../dist/apps/fulfillment',
     emptyOutDir: true,
     sourcemap: true,
+    rollupOptions: {
+      input: {
+        'service-worker': './dev-dist/sw/app.js',
```
tolerants, line 18: Do we really need a separate `dev-dist` folder? From developer perspective it's probably better to have just one dist folder with subfolders and not several of them.

libs/base/ui/form/quantity-input/quantity-input.component.ts
```diff
@@ -0,0 +1,217 @@
...
+   *
+   * Triggers validity reporting so that native browser validation appears.
+   */
+  protected onInput(e: InputEvent): void {
+    this.updateValue((e.target as HTMLInputElement).value);
+    this.dispatch(UPDATE_EVENT);
+  }
+
+  /**
+   * Dispatches an update event
```
tolerants:
```suggestion
   * Dispatches an update event.
```

libs/base/ui/form/quantity-input/quantity-input.component.ts
```diff
@@ -0,0 +1,217 @@
...
+  @property({ type: Number }) step = 1;
+  @property({ type: Boolean }) submitOnChange = false;
+  @property({ type: Boolean }) disabled = false;
+  @property({ type: Boolean }) hasError = false;
+  @property() decreaseIcon?: IconTypes;
+  @property() increaseIcon?: IconTypes;
+
+  protected inputRef = createRef<HTMLInputElement>();
+
+  reset(): void {
```
tolerants, line 48: It would be great to have description for every public method I'd say. But it can done in a separate ticket for sure.

libs/base/ui/form/quantity-input/quantity-input.styles.ts
```diff
@@ -20,7 +21,11 @@ export const styles = css`
   oryx-input {
     --oryx-form-control-border-radius: 0;
 
-    width: var(--oryx-cart-quantity-input-width, 71px);
+    ${unsafeCSS(
+      featureVersion >= '1.3'
+        ? 'width: var(--oryx-quantity-input-width, 71px);'
+        : 'width: var(--oryx-cart-quantity-input-width, 71px);'
+    )}
```
tolerants: In this case we avoid some code duplication. But at the same time we decrease readability a bit. So up to you.
```suggestion
      const inputWidth = featureVersion >= '1.3' ? '--oryx-quantity-input-width' : '--oryx-cart-quantity-input-width';
...
    ${unsafeCSS(
      width: var(${inputWidth}, 71px);
    )}
```

libs/domain/picking/service-worker/auth.feature.ts
```diff
@@ -0,0 +1,5 @@
+import { BapiAuthFeature } from '@spryker-oryx/auth';
+
+export class AuthFeature extends BapiAuthFeature {
```
tolerants: Something is wrong with the naming from my perspective, `AuthFeature` shouldn't extend `BapiAuthFeature` as long as common entity shouldn't extend more specific entity. Maybe `ServiceWorkerAuthFeature`?

libs/domain/picking/service-worker/services/push-initializer.service.ts
```diff
@@ -1,16 +1,16 @@
...
   PickingSyncAction,
   PushSyncPayload,
-} from './picking-sync-action-handler.service';
+} from '@spryker-oryx/picking/offline';
+import { Observable, firstValueFrom } from 'rxjs';
 
 declare let self: ServiceWorkerGlobalScope;
 
-export class SwPushInitializerService implements AppInitializer {
+export class PushInitializerService implements AppInitializer {
```
tolerants: Why not just `PushInitializer`?

libs/domain/picking/src/feature.ts
```diff
@@ -66,39 +48,33 @@ export interface PickingFeatureConfig extends PickingConfig {
...
-      { provide: PickingListAdapter, useClass: PickingListDefaultAdapter },
-      { provide: PickingHttpService, useClass: PickingHttpDefaultService },
-      { provide: PickingHeaderService, useClass: PickingHeaderDefaultService },
-      {
-        provide: WarehouseUserAssignmentsAdapter,
-        useClass: WarehouseUserAssignmentsDefaultAdapter,
+        provide: PickingHeaderService,
+        useClass: PickingHeaderDefaultService,
       },
+      //override SapiLocaleAdapter that is provided by siteFeature with default one
```
tolerants, line 71: Can you please add a TODO with the ticket number for it please?

libs/domain/picking/src/feature.ts
```diff
@@ -66,39 +48,33 @@ export interface PickingFeatureConfig extends PickingConfig {
   noDefaultRoutes?: boolean;
 }
 
-export class PickingFeature implements AppFeature {
+export class PickingFeature extends PickingApiFeature {
```
tolerants: Again, something wrong with the naming. We can think together about it.

libs/platform/offline/src/feature.ts
```diff
@@ -1,16 +1,11 @@
...
-  SyncActionRegistryDefaultService,
-  SyncActionRegistryService,
-  SyncExecutorDefaultService,
-  SyncExecutorService,
+  SyncEntity,
   SyncSchedulerDefaultService,
   SyncSchedulerService,
-} from './services';
-import { OfflineServiceWorkerPlugin } from './sw-plugin';
+} from '../sync';
```
tolerants: Isn't at a separate entry-point?

libs/platform/offline/src/mocks/src/mock-offline.providers.ts
```diff
@@ -1,5 +1,5 @@
 import { Provider } from '@spryker-oryx/di';
-import { SyncSchedulerService } from '@spryker-oryx/offline';
+import { SyncSchedulerService } from '../../../sync';
```
tolerants: Isn't is a separate entry-point?

libs/platform/offline/src/mocks/src/mock-sync.ts
```diff
@@ -1,6 +1,5 @@
-import { SyncStatus } from '@spryker-oryx/offline';
 import { of } from 'rxjs';
-import { vi } from 'vitest';
+import { SyncStatus } from '../../../sync';
```
tolerants: Isn't is a separate entry-point?

libs/platform/offline/src/mocks/src/mock-sync-scheduler-service.ts
```diff
@@ -1,5 +1,5 @@
-import { SyncSchedulerService } from '@spryker-oryx/offline';
 import { Observable, of } from 'rxjs';
+import { SyncSchedulerService } from '../../../sync';
```
tolerants: Isn't is a separate entry-point?

libs/template/application/service-worker/app/app-builder.ts
```diff
@@ -0,0 +1,6 @@
+import { AppBuilderWithModules } from './app.model';
+import { ModularAppBuilder } from './modular-app-builder';
+
+export function appBuilder(): AppBuilderWithModules {
```
tolerants, line 4: Why do we need a copy-paste from `src/app/app-builder.ts`?

libs/domain/picking/api/feature.ts
```diff
@@ -0,0 +1,46 @@
...
+  PickingListService,
+  WarehouseUserAssignmentsAdapter,
+  WarehouseUserAssignmentsDefaultAdapter,
+  WarehouseUserAssignmentsDefaultService,
+  WarehouseUserAssignmentsService,
+  warehouseUserAssignmentNormalizer,
+  warehouseUserAssignmentsNormalizer,
+} from './services';
+
+export class PickingApiFeature implements AppFeature {
```
tolerants: I don't really like the name `api` for this sub-package, it can be confusing for end-users as for me. Any other suggestions? Maybe `services` as we already using for other packages?

tolerants (changes requested):

apps/fulfillment/vite.config.ts
Sinucid: Yup, but FA building process divided into 2 steps: 

1. building service worker with related resources (**dev-dist** folder)
2. building a main bundle and injecting sw stuff into template

dev-dist folder is a temporary folder, just for temporary storing bundled resources before the 2 step. And both steps are accompanied with cleaning up of the build folders before staring the build process, that does not allow us to build sw into dist folder, the main bundle will drop it then

libs/base/ui/form/quantity-input/quantity-input.component.ts
Sinucid: Hmmm, tbh I'm not sure that we really need this public method, need to check the component, maybe even make some refactoring.

libs/base/ui/form/quantity-input/quantity-input.styles.ts
Sinucid: refactored

libs/domain/picking/service-worker/services/push-initializer.service.ts
Sinucid: Refactored

libs/domain/picking/src/feature.ts
Sinucid: This hacky fix was implemented previously, I've just moved the service overriding to another place (previously it was in preset, if I'm not mistaken). Added link to the original ticket in the comment

libs/template/application/service-worker/app/app-builder.ts
Sinucid: This is required to drop lit-releated stuff and not including it to the sw-bundle. I was thinking about creation of something like `BasicAppBuilder`, but we are already have the `SimpleAppBuilder` super class, that can confuse. In result I've just extracted minified version to the '@spryker-oryx/application/service-worker' package to keep it in one place with workbox bootstrap

libs/domain/picking/src/feature.ts
Sinucid: Renamed to `AuthServiceWorkerFeature`

libs/domain/picking/api/feature.ts
Sinucid: Renamed to `picking/services`

libs/template/application/service-worker/app/app-builder.ts
tolerants: Ok, let's keep it as is for now.

apps/fulfillment/vite.config.ts
tolerants: So there is no quick solution, right?

libs/base/ui/form/quantity-input/quantity-input.component.ts
tolerants: So what is the outcome?

tolerants (approved):

PR #888, opened 2023-11-12
fix(ui): icons are rendered outside the clickable area of a link

tobi-or-not-tobi:
When an icon is used in a link, the icon is moved visually inside the clickable real estate of the `<oryx-link>` component, so that users can click on all the real estate to open a link.

closes: [HRZ-90413](https://spryker.atlassian.net/browse/HRZ-90413)

[HRZ-90413]: https://spryker.atlassian.net/browse/HRZ-90413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/base/ui/action/link/src/styles/storefront.styles.ts
+++ b/libs/base/ui/action/link/src/styles/storefront.styles.ts
@@ -4,10 +4,11 @@ export const storefrontLinkStyles = css`
   :host {
     position: relative;
     display: inline-flex;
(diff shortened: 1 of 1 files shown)


Sinucid (dismissed):
I believe you we need to describe the new css vars in the PR description

Sinucid (approved):

PR #889, opened 2023-11-12
fix(experience): top header layout is broken on medium screens

tobi-or-not-tobi:
We've added an option to the flex layout to specify whether items should warp to the next line. On the top header, we do not want this, which is the default behaviour. 

closes: [HRZ-90414](https://spryker.atlassian.net/browse/HRZ-90414)

[HRZ-90414]: https://spryker.atlassian.net/browse/HRZ-90414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/experience/src/services/layout/plugins/types/flex/flex-layout.model.ts
+++ b/libs/platform/experience/src/services/layout/plugins/types/flex/flex-layout.model.ts
@@ -14,7 +14,8 @@ export interface FlexLayoutProperties {
   /**
    * If true, items will wrap to the next line.
    *
-   * @since 1.3, 1.2 had hardcoded wrap for all flex items.
+   * @since 1.3, previously wrap was enabled by default and there was
+   * no ability to disable it.
    */
   wrap?: boolean;
 }
--- a/libs/platform/experience/src/services/layout/plugins/types/flex/flex-layout.model.ts
+++ b/libs/platform/experience/src/services/layout/plugins/types/flex/flex-layout.model.ts
@@ -6,4 +6,15 @@ declare global {
   export interface Layouts {
     flex: undefined;
   }
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  export interface LayoutProperty extends FlexLayoutProperties {}
+}
+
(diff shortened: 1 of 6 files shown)


libs/platform/experience/src/services/layout/plugins/types/flex/flex-layout.plugin.ts
```diff
@@ -1,13 +1,28 @@
...
+  getStyleProperties(
+    data: LayoutPluginPropertiesParams
+  ): Observable<LayoutStyleProperties> {
+    const options = { ...this.$defaultOptions(), ...data.options };
+    const props: LayoutStyleProperties = {};
+    if (options.wrap) {
+      props['flex-wrap'] = 'wrap';
+    }
+
+    return of(props);
```
Sinucid:
```suggestion
    return of(options.wrap ? {'flex-wrap': 'wrap'}: {});
```

Sinucid (changes requested):

libs/platform/experience/src/services/layout/plugins/types/flex/flex-layout.model.ts
```diff
@@ -6,4 +6,15 @@ declare global {
...
   }
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  export interface LayoutProperty extends FlexLayoutProperties {}
+}
+
+export interface FlexLayoutProperties {
+  /**
+   * If true, items will wrap to the next line.
+   *
+   * @since 1.3, 1.2 had hardcoded wrap for all flex items.
```
tolerants:
```suggestion
   * @since 1.3, previously wrap was enabled by default and there was no ability to disable it.
```

tolerants:
Please add package name `fix` -> `fix(experience)`.

tolerants (dismissed):

Sinucid (dismissed):

tolerants (approved):

PR #892, opened 2023-11-13
chore: category navigation e2e

tobi-or-not-tobi:
Test whether the category navigation contains a link to all products and whether the navigation links work.

closes: [HRZ-90392](https://spryker.atlassian.net/browse/HRZ-90392)

[HRZ-90392]: https://spryker.atlassian.net/browse/HRZ-90392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront-e2e/src/integration/category-navigation.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+      const navigation = new CategoryNavigationFragment().get();
+      navigation.should('be.visible');
+    });
+
+    it('should have all products link in the category navigation bar', () => {
+      const link = new CategoryNavigationFragment().getLink(0);
+      link.should('contain.text', 'All products');
+    });
+
+    describe('and a navigation item is clicked', () => {
```
AndreyYevtukhov: I'm not quite sure that we have to test links in this way (it might be too expensive, because we pay for each e2e test) 

We should have a unit test that proves that the <a> element with the correct URL/text is rendered and that's enough)

apps/storefront-e2e/src/integration/category-navigation.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+describe('Category navigation suite', () => {
+  let landingPage;
+
+  describe('when the landing page is visited', () => {
+    beforeEach(() => {
+      landingPage = new LandingPage();
+      landingPage.visit();
+    });
+
+    it('should render the category navigation in the header ', () => {
```
AndreyYevtukhov: We can combine these tests in one to make them more cost-efficient ---
And the second check covers the first one, because if the navigation is not rendered - the link text will not be found and both tests will fail. 

So, the first check can be removed.

apps/storefront-e2e/src/support/page-fragments/category-navigation.fragment.ts
```diff
@@ -0,0 +1,5 @@
+export class CategoryNavigationFragment {
+  get = () => cy.get(`[uid='category-navigation']`);
```
AndreyYevtukhov, line 2: Usually, we call this method `getWrapper` to make it more readable, instead of `this.get()` which can be interpreted differently

AndreyYevtukhov (dismissed):
Left some comments

libs/base/ui/action/link/src/styles/storefront.styles.ts
```diff
@@ -4,10 +4,11 @@ export const storefrontLinkStyles = css`
   :host {
     position: relative;
     display: inline-flex;
+    width: var(--oryx-link-width);
```
Sinucid: Same as before, I would recommend you to describe each new css var in PR's description

Sinucid (changes requested):

libs/base/ui/action/link/src/styles/storefront.styles.ts
tobi-or-not-tobi: @Sinucid  I have done this in the parent PR which will be the base for release docs

tobi-or-not-tobi:
just discussed with @AndreyYevtukhov and we decided to drop e2e completely. We will have very soon visual regression on the storefront, which means we do not need to build this UI e2e manually.

libs/domain/search/box/box.component.spec.ts
```diff
@@ -96,9 +96,9 @@ describe('SearchBoxComponent', () => {
     linkService = testInjector.inject<MockLinkService>(
       LinkService
     ) as MockLinkService;
-    routerService = testInjector.inject(
+    routerService = testInjector.inject<MockRouterService>(
```
Sinucid: Should be:
```suggestion
    routerService = testInjector.inject<MockRouterService>(RouterService);
```
Without type casting. And check the other injections behind this one

Sinucid (changes requested):

Sinucid (approved):

PR #914, opened 2023-11-17
feat(fulfillment): notify user about push notification support and permission 

Sinucid:
Added browser's permissions and supports for push notification API that is required by offline mode.

- created compositions for picking and customer note pages
- fulfillment app integrated with notification center
- fixed word break behavior of notification component's text content
with long words
- `refresh data` button in FA profile menu was renamed to `sync data`,
and the button became disabled in offline mode to prevent unnecessary
requests errors
- NetworkState service was added to platform offline package to observe
browser's network state
- `getList` method was added to picking list service to get picking list
by id
- `PickingListContextFallback` was added to provide fallback picking
list context by `pickingListId` route param

closes: [HRZ-90415](https://spryker.atlassian.net/browse/HRZ-90415)

[HRZ-90415]: https://spryker.atlassian.net/browse/HRZ-90415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/offline/src/services/network-state-default.service.ts
+++ b/libs/platform/offline/src/services/network-state-default.service.ts
@@ -6,16 +6,16 @@ import {
   merge,
   of,
 } from 'rxjs';
-import { NetworkState, NetworkStateService } from './network-state.service';
+import { NetworkStateService } from './network-state.service';
 
 export class NetworkStateDefaultService implements NetworkStateService {
-  get(): Observable<NetworkState> {
+  online(): Observable<boolean> {
     return merge(
       of(null),
       fromEvent(window, 'online'),
       fromEvent(window, 'offline')
     ).pipe(
-      map(() => (navigator.onLine ? 'online' : 'offline')),
+      map(() => navigator.onLine),
       distinctUntilChanged()
     );
   }
--- a/libs/platform/offline/src/services/network-state-default.service.ts
+++ b/libs/platform/offline/src/services/network-state-default.service.ts
@@ -4,17 +4,17 @@ import {
   fromEvent,
   map,
   merge,
-  of,
+  startWith,
 } from 'rxjs';
 import { NetworkStateService } from './network-state.service';
 
 export class NetworkStateDefaultService implements NetworkStateService {
   online(): Observable<boolean> {
     return merge(
-      of(null),
       fromEvent(window, 'online'),
       fromEvent(window, 'offline')
     ).pipe(
+      startWith(navigator.onLine),
       map(() => navigator.onLine),
       distinctUntilChanged()
     );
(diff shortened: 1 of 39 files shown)


libs/platform/offline/src/services/network-state-default.service.ts
```diff
@@ -0,0 +1,22 @@
...
+  map,
+  merge,
+  of,
+} from 'rxjs';
+import { NetworkStateService } from './network-state.service';
+
+export class NetworkStateDefaultService implements NetworkStateService {
+  online(): Observable<boolean> {
+    return merge(
+      of(null),
```
supproduction: why do we need it?

libs/platform/push-notification/web/web-push.ts
```diff
@@ -65,6 +60,46 @@ export class WebPushProvider implements PushProvider<PushSubscriptionJSON> {
...
+      error = 'Browser does not support service-worker API';
+    } else if (!('SyncManager' in window)) {
+      error = 'Browser does not support background sync API';
+    } else if (
+      (await navigator.permissions.query({ name: 'notifications' })).state ===
+      'denied'
+    ) {
+      error =
+        'Permission to accept push notifications is not granted. Check the browser configuration or reset the permission';
+    } else if (
```
supproduction: can we refactor it in order that it will be readable ? (else if is bad pattern)

supproduction (changes requested):

libs/platform/offline/src/services/network-state-default.service.ts
Sinucid: we need initial value for the first emit, due to window is emitting 'offline' | 'online' events only when state really changes

supproduction: use startWith instead then

libs/platform/push-notification/web/web-push.ts
```diff
@@ -30,11 +30,15 @@ export class WebPushProvider implements PushProvider<PushSubscriptionJSON> {
...
+    return this.precondition().pipe(
+      switchMap(() =>
+        this.getExistingSubscription().pipe(
+          switchMap((subscription) =>
+            subscription ? of(subscription) : this.createSubscription()
+          ),
+          map((subscription) => subscription.toJSON())
+        )
+      )
     );
```
supproduction:
```suggestion
    return this.precondition().pipe(
      switchMap(() =>
        this.getExistingSubscription()
      )
       switchMap((subscription) =>
            subscription ? of(subscription) : this.createSubscription()
          ),
          map((subscription) => subscription.toJSON()
    );
```

supproduction (changes requested):

supproduction (approved):

PR #920, opened 2023-11-20
fix: remove offlineFulfillmentFeatures for feature sets

tobi-or-not-tobi:
we deprecated `offlineFulfillmentFeatures` and only expose `fulfillmentFeatures` in our presets. It does not mean we're no longer supporting online fulfilment feauters, but it's no longer a standard feature set. 

closes: [HRZ-90524](https://spryker.atlassian.net/browse/HRZ-90524)

[HRZ-90524]: https://spryker.atlassian.net/browse/HRZ-90524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/presets/fulfillment/app.ts
```diff
@@ -46,6 +52,21 @@ export function fulfillmentFeatures(
...
+      ...config,
+    };
+    return [
+      ...onlineFeatures,
+      new IndexedDbFeature(config?.indexedDb),
+      new OfflineFeature(),
+      new OfflinePickingFeature(),
+    ];
+  } else {
+    return onlineFeatures;
```
tolerants: I don't really like we can't early return this easy case. But not critical.

tolerants (dismissed):

tolerants (approved):

PR #924, opened 2023-11-21
feat(picking): drop headers components 

Sinucid:
Since FA uses static experience data for building of the pages, some of the components are no longer relevant and can be replaced with more flexible and customisable compositions.

- `oryx-picking-header`, `oryx-picking-list-header` and `oryx-picking-picker-header` components were dropped and replaced with compositions on corresponding pages
- `oryx-picking-reference` component was added to render order id 
- `oryx-picking-search` component was added to perform search by orders reference id
- `PickingHeaderService` was dropped and replaced with `PickingGuardService`, that can be mediator for any routes and protect it from unexpected living from the page
- `PickingItemComponent` add its id to the payload to 'oryx.note' event, instead of cart note message
-  `setSortingQualifier` and  `getSortingQualifier` methods of `PickingListService` were replaced with more generic methods `setQualifier` and  `getQualifier`, that can work with any type of qualifiers
- new methods `isActiveSearch` and toggleActiveSearch were added to `PickingListService`, to return and toggle active search state of the service
- `get` method of `PickingListService` is no more accepting params. Qualifier can be specified via `setQualifier`
  
closes: [HRZ-90510](https://spryker.atlassian.net/browse/HRZ-90510)

[HRZ-90510]: https://spryker.atlassian.net/browse/HRZ-90510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/presets/fulfillment/experience/templates/header-picking-lists.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const HeaderPickingListsTemplate: ExperienceComponent = {
+  ...HeaderTemplate,
+  id: 'header-picking-lists',
+  components: [
+    {
+      type: 'oryx-content-text',
+      content: {
+        data: {
+          text: `<h4>${i18n('picking.pick-lists')}</h4>`,
```
tobi-or-not-tobi:
```suggestion
          text: `<h4>PICK LISTS</h4>`,
```

i18n is not supported, we introduced it lately but that's been a mistaken

libs/template/presets/fulfillment/experience/templates/header-picking-lists.ts
```diff
@@ -0,0 +1,23 @@
...
+        },
+      },
+      options: {
+        rules: [{ style: 'flex: 1 0 auto; text-transform: uppercase;' }],
+      },
```
tobi-or-not-tobi:
```suggestion
      options: {
        rules: [{ margin: '0 auto 0 0' }],
      },
```

alternative

libs/template/presets/fulfillment/experience/templates/header.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const HeaderTemplate: ExperienceComponent = {
+  type: 'oryx-composition',
+  id: 'header',
+  components: [
+    {
+      type: 'oryx-site-navigation-item',
+      options: {
+        contentBehavior: 'modal',
+        label: i18n('picking.account'),
```
tobi-or-not-tobi:
```suggestion
        label: 'Account'
```

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #926, opened 2023-11-21
fix: update state error after log out

nsilva-spryker:
Closes: [HRZ-90534](https://spryker.atlassian.net/browse/HRZ-90534)

[HRZ-90534]: https://spryker.atlassian.net/browse/HRZ-90534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/auth/src/oauth/providers/code-grant.ts
+++ b/libs/platform/auth/src/oauth/providers/code-grant.ts
@@ -88,7 +88,9 @@ export class OauthCodeGrantProvider implements OauthProvider {
 
   revoke(): Observable<void> {
     if (!this.config.revocationUrl) {
-      return this.updateState();
+      this.state$.next(undefined);
+
+      return of();
     }
 
     return this.revocationRequest().pipe(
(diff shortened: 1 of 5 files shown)


libs/platform/auth/src/oauth/providers/code-grant.ts
```diff
@@ -88,7 +88,9 @@ export class OauthCodeGrantProvider implements OauthProvider {
 
   revoke(): Observable<void> {
     if (!this.config.revocationUrl) {
-      return this.updateState();
+      this.state$.next(undefined);
```
dunqan: With previous solution we were both removing token and nexting undefined to the state as a result 🤔 I'm wondering why not removing helps, also is this behavior is correct, as we are just changing local app state (so full refresh may still show, that you are logged in, as token would be picked up)?

nsilva-spryker: After nexting undefined we were setting state again(basically removes from local storage to set it again). Maybe some async problem with rxjs was not adding the token again on local storage with new data and with that the login was block.
Token data  on local storage is update depending if you are logged or not.

tobi-or-not-tobi (approved):

dunqan (approved):

PR #928, opened 2023-11-22
refactor(picking): use shadow style for header

tobi-or-not-tobi:
Since we moved the header to experience data, we can leverage the shadow style property.

### DIFF
--- a/libs/base/ui/navigation/tabs/src/tabs.styles.ts
+++ b/libs/base/ui/navigation/tabs/src/tabs.styles.ts
@@ -27,7 +27,7 @@ export const baseStyles = css`
   }
 
   :host([shadow]) slot:not([name]) {
-    box-shadow: 0 4px 8px var(--oryx-elevation-color-2);
+    box-shadow: 0 4px 8px var(--oryx-shadow-color);
   }
 
   input[type='range'] {
--- a/libs/base/ui/navigation/tabs/src/tabs.styles.ts
+++ b/libs/base/ui/navigation/tabs/src/tabs.styles.ts
@@ -5,8 +5,8 @@ import { TabsAppearance } from './tabs.model';
 const secondary = unsafeCSS(TabsAppearance.Secondary);
 const shadowCSSVar =
   featureVersion >= '1.3'
-    ? unsafeCSS('--oryx-shadow-color')
-    : unsafeCSS('--oryx-elevation-color-2');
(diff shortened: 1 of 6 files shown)


libs/base/ui/navigation/tabs/src/tabs.styles.ts
```diff
@@ -27,7 +27,7 @@ export const baseStyles = css`
   }
 
   :host([shadow]) slot:not([name]) {
-    box-shadow: 0 4px 8px var(--oryx-elevation-color-2);
+    box-shadow: 0 4px 8px var(--oryx-shadow-color);
```
Sinucid: BC

libs/template/themes/design-tokens/src/common-tokens.ts
```diff
@@ -1,6 +1,21 @@
 import { ThemeToken } from '@spryker-oryx/experience';
 
 export const commonTokens: ThemeToken = {
+  /** @deprecated use oryx-shadow instead */
+  elevation: {
```
Sinucid, line 5: BC?

Sinucid (changes requested):

libs/template/themes/design-tokens/src/common-tokens.ts
Sinucid: And why deprecated?

tobi-or-not-tobi: no BC until it's removed. We moved to `shadow` prefix rather than `elevation`.

libs/base/ui/navigation/tabs/src/tabs.styles.ts
```diff
@@ -27,7 +27,11 @@ export const baseStyles = css`
   }
 
   :host([shadow]) slot:not([name]) {
-    box-shadow: 0 4px 8px var(--oryx-elevation-color-2);
```
Sinucid, line 30: I would make it a bit more readable:

```
const shadowCSSVar = featureVersion >= '1.3' ? 
                                         unsafeCSS( '--oryx-shadow-color'): 
                                         unsafeCSS( '--oryx-elevation-color-2'): 

...

box-shadow: 0 4px 8px var(${shadowCSSVar});
```

Sinucid (changes requested):

Sinucid (dismissed):

Sinucid (approved):

PR #931, opened 2023-11-22
feat: guest cart loading improvement

dunqan:
- drops AnonAuthTokenService and AnonTokenInterceptors
- handles guest user in IdentityService with optional requireGuest flag
- add GuestIdentityInterceptor
- guest user Id is not created and guest cart is not fetched before first add to cart

Closes [HRZ-90591](https://spryker.atlassian.net/browse/HRZ-90591)

[HRZ-90591]: https://spryker.atlassian.net/browse/HRZ-90591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/auth/src/sapi/sapi-identity.service.ts
```diff
@@ -2,31 +2,81 @@ import {
...
 import { parseToken } from './utils';
 
 export class SapiIdentityService implements IdentityService {
+  protected ANONYMOUS_USER_IDENTIFIER = 'oryx.anonymous-user';
+
+  protected generateGuest$ = new BehaviorSubject<boolean>(false);
+
   protected identity$ = this.authService.getToken().pipe(
-    map((token) => this.getFromToken(token)),
+    featureVersion >= '1.3'
```
tolerants:
```suggestion
    featureVersion >= '1.4'
```

libs/platform/auth/src/services/anon-auth-token.service.ts
```diff
@@ -41,7 +43,11 @@ export class AnonAuthTokenService implements AuthTokenService {
       .get<string>(this.ANONYMOUS_USER_IDENTIFIER, StorageType.Session)
       .pipe(
         switchMap((userId) => {
-          return userId ? of(userId) : this.createAnonymousId();
+          return featureVersion >= '1.3'
```
tolerants:
```suggestion
          return featureVersion >= '1.4'
```

tolerants (changes requested):

libs/platform/auth/src/services/anon-auth-token.service.ts
```diff
@@ -41,7 +43,11 @@ export class AnonAuthTokenService implements AuthTokenService {
       .get<string>(this.ANONYMOUS_USER_IDENTIFIER, StorageType.Session)
       .pipe(
         switchMap((userId) => {
-          return userId ? of(userId) : this.createAnonymousId();
+          return featureVersion >= '1.4'
+            ? of(userId)
+            : userId
+            ? of(userId)
+            : this.createAnonymousId();
```
tobi-or-not-tobi:
```suggestion
          return featureVersion >= '1.4' || userId
            ? of(userId)
            : this.createAnonymousId();
```

tobi-or-not-tobi: but seems that we do not need this, as we're no longer using this service in >= 1.4

tobi-or-not-tobi (changes requested):

libs/platform/auth/src/services/anon-auth-token.service.ts
dunqan: 👍

tobi-or-not-tobi (dismissed):

libs/platform/auth/src/sapi/feature.ts
```diff
@@ -92,7 +109,9 @@ export class SapiAuthFeature extends OauthFeature implements AppFeature {
 
 export interface SapiAuthFeatureConfig extends OauthFeatureConfig {
   skipRoutes?: boolean;
+  /** deprecated since 1.4 */
```
tolerants: Maybe let's add a note that `guestIdentityInterceptor` should be used instead, so it's clear without taking a look into this interface.

libs/platform/auth/src/services/anon-auth-token.service.ts
```diff
@@ -13,6 +13,7 @@ import { AuthTokenData, AuthTokenService } from './auth-token.service';
 import { AuthService } from './auth.service';
 import { generateID } from './utils';
 
+/** deprecated since 1.4 */
```
tolerants: Suggest to add a replacement.

tolerants (approved):

PR #938, opened 2023-11-23
fix: move light mode color mode out of boilerplate

tobi-or-not-tobi:
We move the light mode configuration to the feature set of fulfillment app.

closes: [HRZ-90569](https://spryker.atlassian.net/browse/HRZ-90569)

[HRZ-90569]: https://spryker.atlassian.net/browse/HRZ-90569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/template/presets/fulfillment/app.ts
+++ b/libs/template/presets/fulfillment/app.ts
@@ -4,7 +4,7 @@ import {
 } from '@spryker-oryx/application';
 import { BapiAuthComponentsFeature, BapiAuthFeature } from '@spryker-oryx/auth';
 import { contentFeature } from '@spryker-oryx/content';
-import { AppFeature, coreFeature } from '@spryker-oryx/core';
+import { AppFeature, coreFeature, FeatureOptions } from '@spryker-oryx/core';
 import { experienceFeature, Resources } from '@spryker-oryx/experience';
 import { formFeature } from '@spryker-oryx/form';
 import { I18nFeature, I18nFeatureOptions } from '@spryker-oryx/i18n';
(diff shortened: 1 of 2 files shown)


supproduction (approved):

libs/template/presets/fulfillment/app.ts
```diff
@@ -62,6 +62,10 @@ export function fulfillmentFeatures(
               provide: ThemeMetaInitializer,
               useClass: PWAThemeMetaInitializer,
             },
+            {
+              provide: FeatureOptions,
+              useValue: { 'oryx-app': { colorMode: ColorMode.Light } },
```
tolerants, line 67: But can we change experience in patch? For me it looks like a minor and for me it looks like we need to wrap it with featureVersion.

tolerants (changes requested):

libs/template/presets/fulfillment/app.ts
tobi-or-not-tobi: The experience does not change. It's a fix to me. We initially forced this behaviour with boilerplate, but realized that we should do it with feature set. I'd not overcomplicate it.

tolerants: Well, as long as it's in our module - it can change an experience as we're forcing some specific mode. But ok, it's not critical to me.

tolerants (approved):

PR #944, opened 2023-11-26
feat(product): product category list component

tobi-or-not-tobi:
Product Category list component can be used inside navigations. The component contains an `exclude` option, that can be used to exclude one or multiple categories from the list. Excluding can be done by providing IDs (array or comma separated), smaller or greater than an ID (e.g. `<5`) or a range of IDs (e.g. `5..8`).

closes: [HRZ-90382](https://spryker.atlassian.net/browse/HRZ-90382)

[HRZ-90382]: https://spryker.atlassian.net/browse/HRZ-90382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/product/category/list/list.component.ts
+++ b/libs/domain/product/category/list/list.component.ts
@@ -1,0 +1,52 @@
+import { resolve } from '@spryker-oryx/di';
+import { ContentMixin, LayoutMixin } from '@spryker-oryx/experience';
+import { ProductCategoryService, ProductMixin } from '@spryker-oryx/product';
+import { computed, hydrate } from '@spryker-oryx/utilities';
+import { LitElement, TemplateResult, html } from 'lit';
+import { property } from 'lit/decorators.js';
+import { repeat } from 'lit/directives/repeat.js';
+import { unsafeHTML } from 'lit/directives/unsafe-html.js';
+import {
+  ProductCategoryListAttributes,
+  ProductCategoryListOptions,
+} from './list.model';
+import { categoryListStyles } from './list.styles';
+
+@hydrate()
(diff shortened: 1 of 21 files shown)


libs/domain/product/category/list/list.component.ts
```diff
@@ -0,0 +1,51 @@
...
+import { repeat } from 'lit/directives/repeat.js';
+import { unsafeHTML } from 'lit/directives/unsafe-html.js';
+import {
+  ProductCategoryListAttributes,
+  ProductCategoryListOptions,
+} from './list.model';
+
+@hydrate()
+export class ProductCategoryListComponent
+  extends ProductMixin(
```
dunqan: do we need `ProductMixin` for this component?

libs/domain/product/src/services/category/adapter/default-product-category.adapter.ts
```diff
@@ -12,7 +16,7 @@ export class DefaultProductCategoryAdapter implements ProductCategoryAdapter {
     protected transformer = inject(JsonAPITransformerService)
   ) {}
 
-  get(categoryId: string): Observable<ProductCategory> {
+  get(qualifier: ProductCategoryQualifier): Observable<ProductCategory> {
```
dunqan: this should not be a string in a first place agreed, but is it a breaking change?

ah, I guess we are missing string handling for <1.4?

dunqan (approved):

PR #947, opened 2023-11-27
fix(resources): PWA manifest icons for dark mode

tobi-or-not-tobi:
We've made a few changes:
- Single base64 SVG based favicon that handles dark mode inline
- Single SVG based public file for the manifest, as base64 won't work. 

closes: [HRZ-90568](https://spryker.atlassian.net/browse/HRZ-90568)

[HRZ-90568]: https://spryker.atlassian.net/browse/HRZ-90568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/resources/src/graphics/common/logo.ts
```diff
@@ -1,5 +1,11 @@
-const symbol = `
-<g part="symbol">
+import { featureVersion } from '@spryker-oryx/utilities';
+
+const darkMode =
+  featureVersion >= '1.4'
```
tolerants, line 4: In case we are relying on the `1.4` version then it's a `feat` probably. But it looks like a `fix` to me, so then we probably should backport it, and rely on `1.3` version.

libs/template/resources/src/graphics/common/logo.ts
```diff
@@ -24,11 +29,20 @@ const wordmark = `
...
 `;
-export const logoSymbol = `<svg viewBox="0 0 28.3 41.2">${symbol}</svg>`;
-export const logoWordmark = `<svg viewBox="36 16 73 18">${wordmark}</svg>`;
+
+export const logoSymbol = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 28.3 41.2">
+  ${darkMode}
+  ${symbol}
+</svg>`;
+
+export const logoWordmark = `<svg xmlns="http://www.w3.org/2000/svg"  viewBox="36 16 73 18">
```
tolerants:
```suggestion
export const logoWordmark = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="36 16 73 18">
```

apps/fulfillment/vite.config.ts
```diff
@@ -40,26 +40,10 @@ export default defineConfig({
...
-            sizes: '192x192',
-            type: 'image/png',
-          },
-          {
-            purpose: 'maskable',
-            src: '/icons/android-chrome-maskable-512x512.png',
-            sizes: '512x512',
-            type: 'image/png',
+            src: '/icons/logo.svg',
+            sizes: '192x192 512x512',
```
tolerants, line 44: But what about png for chrome?

tolerants (changes requested):

apps/fulfillment/vite.config.ts
tobi-or-not-tobi: Why do we need PNG for chrome?

tolerants: Maybe I got it wrong, but I think you said during the daily svg is not working for android (not chrome, sorry).

tolerants (dismissed):

tolerants (approved):

PR #949, opened 2023-11-27
chore: create oryx updates

dunqan:
Updates `create-oryx` utility:
- use new url for downloading template
- check for latest version and warn
- always download latest template

Closes [HRZ-90610](https://spryker.atlassian.net/browse/HRZ-90610)

[HRZ-90610]: https://spryker.atlassian.net/browse/HRZ-90610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/template/create-oryx/package.json
+++ b/libs/template/create-oryx/package.json
@@ -1,13 +1,14 @@
 {
   "name": "create-oryx",
-  "version": "0.21.0",
+  "version": "0.33.0",
   "type": "module",
   "author": "Spryker Systems GmbH",
   "license": "See license in LICENSE",
   "description": "Create Spryker Oryx Apps",
   "bin": "./bin/main.cjs",
   "devDependencies": {
-    "@spryker-oryx/cli": "*"
+    "@spryker-oryx/cli": "*",
+    "latest-version": "^7.0.0"
   },
   "publishConfig": {
     "access": "public"
--- a/libs/template/create-oryx/package.json
+++ b/libs/template/create-oryx/package.json
@@ -7,8 +7,7 @@
   "description": "Create Spryker Oryx Apps",
   "bin": "./bin/main.cjs",
   "devDependencies": {
-    "@spryker-oryx/cli": "*",
-    "latest-version": "^7.0.0"
+    "@spryker-oryx/cli": "*"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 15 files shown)


libs/template/create-oryx/package.json
```diff
@@ -1,13 +1,14 @@
 {
   "name": "create-oryx",
-  "version": "0.21.0",
+  "version": "0.33.0",
```
tolerants, line 3: Wow... 12 versions. Maybe it make sense to use rc for testing? :)

libs/template/create-oryx/src/check-version.ts
```diff
@@ -0,0 +1,18 @@
+import latestVersion from 'latest-version';
```
tolerants: Do we really need a dependency for it?

libs/template/create-oryx/package.json
```diff
@@ -1,13 +1,14 @@
...
   "type": "module",
   "author": "Spryker Systems GmbH",
   "license": "See license in LICENSE",
   "description": "Create Spryker Oryx Apps",
   "bin": "./bin/main.cjs",
   "devDependencies": {
-    "@spryker-oryx/cli": "*"
+    "@spryker-oryx/cli": "*",
```
tolerants: Why not `latest`?

tolerants (changes requested):

libs/template/create-oryx/package.json
dunqan: Not 12, as new one started with `0.30`.
It was not about testing, as testing was done with verdaccio, but troubleshoting deployment issues.
Not sure if there is a point for rc either, as nothing/noone is relying on this version.

libs/template/create-oryx/package.json
dunqan: It doesn't make a big difference, as we build it internally anyway. But techically "*" is more correct, as we don't use latest stable release, but what we have in the monorepo at the time of the build.

libs/template/create-oryx/src/check-version.ts
dunqan: Define "really". The really,  "really" we do not :)
It's only a dev-dependency, it's not affecting anything else.

tolerants (approved):

PR #950, opened 2023-11-27
feat(cart): coupon component

nsilva-spryker:
Coupon can be applied to the cart to get discounts for some specific product. More than one coupon can be applied to a cart, each coupon have a validation date that's it presented to the user.

There is a validation on the coupon input to give feedback to the user. Coupons are both used for guest users and authenticated users.

Closes [HRZ-2852](https://spryker.atlassian.net/browse/HRZ-2852)

[HRZ-2852]: https://spryker.atlassian.net/browse/HRZ-2852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


supproduction (dismissed):

tobi-or-not-tobi (changes requested):
We're missing storybook stories
- demo story
- static story
  - state with no added coupons
  - state with one added coupons (success)  
  - state with two added coupons (success)
  - state with two added coupons (one success, one failure)

tobi-or-not-tobi (approved):

PR #951, opened 2023-11-27
feat: add strapi for storefront

supproduction:
closes [HRZ-3254](https://spryker.atlassian.net/browse/HRZ-3254)

[HRZ-3254]: https://spryker.atlassian.net/browse/HRZ-3254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/content/src/models/content.model.ts
```diff
@@ -1,10 +1,22 @@
+declare global {
+  interface DynamicContentFields {
+    component: undefined;
+  }
+}
+
+/**
+ * @deprecated since 1.3.1 will be removed. Use DynamicContentFields interface instead.
```
tobi-or-not-tobi:
```suggestion
 * @deprecated since 1.4 will be removed. Use DynamicContentFields interface instead.
```

libs/domain/content/src/models/content.model.ts
```diff
@@ -1,10 +1,22 @@
+declare global {
+  interface DynamicContentFields {
```
tobi-or-not-tobi: Why `Dynamic`? Isn't all dynamic? I tend to drop prefixes like these (remember `SemanticLinkService`?)

libs/template/labs/src/articles/strapi/strapi-content.adapter.ts
```diff
@@ -0,0 +1,155 @@
...
+      {}
+    ).strapi.defaultType;
+  }
+
+  /**
+   * @deprecated Since version 1.1. Will be removed.
+   */
+  getKey(qualifier: ContentQualifier): string {
+    return qualifier.id ?? qualifier.query ?? '';
+  }
```
tobi-or-not-tobi, line 51: Let's not introduce it

tobi-or-not-tobi: same for Contentful and Storyblok

libs/template/presets/storefront/experience/footer.ts
```diff
@@ -86,7 +86,7 @@ const aboutUs = {
   content: {
     data: {
       text: `<b>About us</b><br/>
-        <a href="/">Company</a><br/>
+        <a href="/about/company">Company</a><br/>
```
tobi-or-not-tobi: bc

tobi-or-not-tobi: I'd leave it out for now. Whenever our CMS integration is ready, we'll create the footer navigations by CMS, e.g. getting a content list where we query CMS. We'll have different lists from different CMS.

tobi-or-not-tobi (changes requested):

libs/template/labs/src/articles/strapi/strapi-content.adapter.ts
supproduction: I can'r cause it's required

libs/domain/content/src/models/content.model.ts
supproduction: We. already have `ContentFields` which is deprecated, I need another name then

tobi-or-not-tobi (approved):

PR #953, opened 2023-11-29
feat(cart): manage carts crud and context provider

Sinucid:
- `create` and `delete` methods were added to the cart adapter to create or delete the cart
- `createCart` and `deleteCart` methods were added to the cart service
- `setActiveCart` method added to cart service to toggle default cart

closes: [HRZ-90079](https://spryker.atlassian.net/browse/HRZ-90079), [HRZ-90089](https://spryker.atlassian.net/browse/HRZ-90089)

[HRZ-90079]: https://spryker.atlassian.net/browse/HRZ-90079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-90089]: https://spryker.atlassian.net/browse/HRZ-90089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/cart/services/src/adapter/default-cart.adapter.ts
```diff
@@ -231,11 +250,18 @@ export class DefaultCartAdapter implements CartAdapter {
...
       : [ApiCartModel.Includes.Items];
 
     return `${this.SCOS_BASE_URL}/${path}${`?include=${includes.join(',')}`}`;
   }
+
+  //TODO: adjust setting of the cart name for multi vs single cart
+  protected generateCartName(
+    qualifier?: CreateCartQualifier
+  ): Observable<string | void> {
+    return of(qualifier?.name ?? 'My cart');
```
supproduction, line 265: Why observable?

supproduction (changes requested):

libs/domain/cart/services/src/adapter/default-cart.adapter.ts
Sinucid: This is a temporary placeholder, that will be replaced with real logic during implementation of cart creation (separate story). There is a nuance in cart creation for single vs multi cart envs. **name** field is required for multi and forbidden for single, that is required some logic to figure out the type the env. I don't think that final solution will be synchronous

supproduction (approved):

PR #956, opened 2023-11-29
test: fix fa e2e tests after mock removal

AndreyYevtukhov:
* refactored 85% of tests and supporting files
* removed useless tests from suites
* split suites in E2E (integration folder) and components tests (they should not be a part of the E2E check, but should be covered by unit/component tests)

---
There are 2 sources of flakiness which I can't handle:

* when you log in - in 50% of cases you are stuck on the loading screen because some backend requests never occur. I'm 99% sure this is a front-end issue
* when you create a new order and go to back office ui -> in 50% of cases 3 minutes (!!!) is not enough to process the order and change the status so that it can be moved to ready for picking. It is purely a BE-related topic, we can't fix it. From what I know, a job that changes statuses runs each minute and I have no idea why 3 minutes on a test server (which no one uses) is insufficient. 

---
If checks are fine, and CR is approved -> I'll merge development in this branch to check if it works after all refactorings we made (don't want to mess the code with possible merge conflict resolutions)

closes: https://spryker.atlassian.net/browse/HRZ-89888


apps/fulfillment-e2e/src/component/customer-note.cy.ts
```diff
@@ -4,12 +4,15 @@ import { ListsFragment } from '../support/page_fragments/lists.fragment';
...
 
-describe('Customer note suite', () => {
+// TODO: this is not E2E, these are component tests
+xdescribe('Customer note suite', () => {
   beforeEach(() => {
     cy.clearIndexedDB();
     cy.login();
   });
 
+  // TODO: ????
```
tolerants: Don't you want to replace question marks with some real todo?

apps/fulfillment-e2e/src/support/page_objects/abstract-fa.page.ts
```diff
@@ -1,4 +1,4 @@
-export class AbstractFAPage {
```
tolerants, line 1: Please revert, we shouldn't use such abbreviations. I'd say we shouldn't even use `FA` abbreviation.

tolerants: btw, why not just `Page`?

apps/fulfillment-e2e/src/component/start-picking-list-with-customer-note.cy.ts
```diff
@@ -53,6 +55,7 @@ describe('Start picking a picklist with customer note', () => {
       customerNoteFragment.getProceedToPickingButton().click();
     });
 
+    // TODO: this is a mess
```
tolerants: Can I ask you to review all the TODO's? So there are clear descriptions.

tolerants (changes requested):

AndreyYevtukhov:
Pushed selectors updates, but some errors in the application should be fixed:

1. Error in the console while logging in 
Screenshot 2023-12-08 at 08 53 06

2. When the "Sync data" button is clicked -> modal is not closed (it should close)
Screenshot 2023-12-08 at 08 53 17

3. There are some UI issues as well: on the login page and on the Loading screen (see this thread for details - https://spryker.slack.com/archives/C029XPSBXC6/p1701252092755179)

---
I've also improved picking-lists tests a bit, now it should run faster if there is a visible picking already

tolerants (approved):

PR #960, opened 2023-11-30
my account navigation

tobi-or-not-tobi:


libs/domain/user/navigation-trigger/navigation-trigger.component.ts
```diff
@@ -0,0 +1,47 @@
...
+import { RouteType, RouterService } from '@spryker-oryx/router';
+import { LinkService } from '@spryker-oryx/site';
+import { ButtonType } from '@spryker-oryx/ui/button';
+import { IconTypes } from '@spryker-oryx/ui/icon';
+import { UserService } from '@spryker-oryx/user';
+import { hydrate, signal } from '@spryker-oryx/utilities';
+import { LitElement, TemplateResult, html } from 'lit';
+
+@hydrate()
+export class UserNavigationTriggerComponent extends ContentMixin(LitElement) {
```
dunqan: UserAuthNavigationComponent? 🤔 
Not sure, but this "navigation trigger" sounds not too clear for me...

libs/template/presets/storefront/experience/header.ts
```diff
@@ -82,6 +83,47 @@ export const topHeader = (options?: {
...
+      options: {
+        type: RouteType.AccountOverviewPage,
+        id: 'overview',
+        icon: IconTypes.User,
+      },
+      content: { data: { text: 'Overview' } },
+    },
+    {
+      type: 'oryx-content-link',
+      options: {
```
dunqan: This one will also have to be in labs only for now

tobi-or-not-tobi: agreed. Moved over to labs.

PR #961, opened 2023-12-01
refactor(ui): improve heading utility to support flexible margins margin

tobi-or-not-tobi:
Improved the heading margin so that we can provide custom block margin. The block margin defaults to `0`.

### DIFF
--- a/libs/base/ui/form/input-list/src/input-list.styles.ts
+++ b/libs/base/ui/form/input-list/src/input-list.styles.ts
@@ -41,7 +41,7 @@ export const styles = css`
 export const screenStyles = screenCss({
   sm: css`
     legend {
-      ${headingUtil(HeadingTag.Subtitle, '0 20px')}
+      ${headingUtil(HeadingTag.Subtitle, { margin: '0 0 20px' })}
     }
 
     .content {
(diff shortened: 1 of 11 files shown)


libs/base/ui/form/input-list/src/input-list.styles.ts
```diff
@@ -41,9 +41,7 @@ export const styles = css`
 export const screenStyles = screenCss({
   sm: css`
     legend {
-      ${headingUtil(HeadingTag.Subtitle)}
-
-      margin-block-end: 20px;
+      ${headingUtil(HeadingTag.Subtitle, '0 20px')}
```
Sinucid: maybe better to use an object with styles rules as the second param instead of just a value for margin-block? WiIll be much more flexible

Sinucid: Otherwise, do we really need to specify this additional options for **headingUtil**?

Sinucid: Or at least not to focus on the margin-block only, and allow to pass value just for `marging` rule, ie: '0 20px 0 0'

tobi-or-not-tobi: agreed. will go with your first suggestion. margin: `0 0 20px` conflicts a bit with our approach to logical properties, although it's not wrong

Sinucid (dismissed):

Sinucid (approved):

PR #962, opened 2023-12-01
feat(experience): Layout Transition Property Plugin

tobi-or-not-tobi:
The layout transition property plugin will conditionally apply a transition to the layout. Previously, this affect was applied to all layouts, which caused performance issues during the re-render of the layout. 

closes: [HRZ-90695](https://spryker.atlassian.net/browse/HRZ-90695)

[HRZ-90695]: https://spryker.atlassian.net/browse/HRZ-90695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/experience/src/services/layout/plugins/properties/transition/index.ts
+++ b/libs/platform/experience/src/services/layout/plugins/properties/transition/index.ts
@@ -1,0 +1,1 @@
+export * from './transition.model';
--- a/libs/platform/experience/src/services/layout/plugins/types/list/index.ts
+++ b/libs/platform/experience/src/services/layout/plugins/types/list/index.ts
@@ -1,0 +1,1 @@
+export * from './list.model';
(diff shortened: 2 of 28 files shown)


libs/platform/experience/src/services/layout/plugins/types/list/index.ts
```diff
@@ -0,0 +1 @@
+export * from './list.model';
```
supproduction, line 1: Please, add all needed exports

libs/platform/experience/src/services/layout/plugins/properties/transition/index.ts
```diff
@@ -0,0 +1 @@
+export * from './transition.model';
```
supproduction, line 1: Please, adjust exports

supproduction (dismissed):

supproduction (approved):

PR #964, opened 2023-12-04
feat(experience): added before and after divider

tobi-or-not-tobi:
We've added a before and after divider styles to the layout property. Additionally, the dividers (color, width and style) are now configurable per divider. 

We've also fixed a small issue with the footer styles. 

closes: [HRZ-90698](https://spryker.atlassian.net/browse/HRZ-90698)
fixes: [HRZ-90697](https://spryker.atlassian.net/browse/HRZ-90697)

[HRZ-90698]: https://spryker.atlassian.net/browse/HRZ-90698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-90697]: https://spryker.atlassian.net/browse/HRZ-90697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/themes/design-tokens/src/layout.tokens.ts
```diff
@@ -15,7 +16,7 @@ export const layoutTokens: ThemeToken = {
     split: { equal: '6', aside: '3', main: '8' },
   },
   row: { gap: '20px' },
-  divider: { width: '1px' },
+  ...(featureVersion >= '1.4' ? { divider: { width: '1px' } } : {}),
```
tolerants: Is it correct? It looks like we are removing previous behaviour.

libs/platform/experience/src/services/layout/plugins/properties/divider/stories/demo.stories.ts
```diff
@@ -0,0 +1,57 @@
+import { Meta, Story } from '@storybook/web-components';
+import { TemplateResult, html } from 'lit';
+import { storybookPrefix } from '../../../../../../../.constants';
+import { items } from './util';
+
+export default {
```
tolerants, line 6: Does it make sense to disable snapshot for demo story?

tolerants (changes requested):

tolerants:
I think we can specify this feat is related to experience package, so we can change `feat` to `feat(experience)`.

libs/template/presets/storefront/experience/footer.ts
```diff
@@ -279,12 +280,7 @@ export const FooterTemplate: ExperienceComponent = {
       {
         layout:
           featureVersion >= '1.2'
-            ? {
-                type: 'flex',
-                divider: true,
-                bleed: true,
-                sticky: true,
-              }
+            ? { type: 'flex', bleed: true, sticky: true }
```
tolerants, line 283: Shouldn't we add a new condition?

tolerants (changes requested):

libs/template/presets/storefront/experience/footer.ts
tobi-or-not-tobi: This was a bug and was also not having any effect. Hence i took the freedom to drop it.

tolerants (dismissed):

tolerants (approved):

PR #965, opened 2023-12-05
chore: use inline-block when there's no tag

tobi-or-not-tobi:
Heading elements without a tag, no longer causing a block display. 

closes: [HRZ-90720](https://spryker.atlassian.net/browse/HRZ-90720)

[HRZ-90720]: https://spryker.atlassian.net/browse/HRZ-90720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/base/ui/structure/heading/src/typography.controller.ts
+++ b/libs/base/ui/structure/heading/src/typography.controller.ts
@@ -46,9 +46,7 @@ export class TypographyController implements ReactiveController {
    * Sets the font properties for the heading, based on the provided tag and size.
    */
   protected setFont(tag?: HeadingTag | HeadingVisibility, size?: Size): void {
-    if (!tag) {
-      return;
-    }
+    if (!tag) return;
 
     this.setStyleProperty('--_s', size, `var(--oryx-typography-${tag}-size)`);
     this.setStyleProperty('--_w', size, `var(--oryx-typography-${tag}-weight)`);
@@ -63,7 +61,7 @@ export class TypographyController implements ReactiveController {
     tag?: HeadingTag | HeadingVisibility,
     size?: Size
   ): void {
-    if ((!tag && !size) || (!size && this.isInlineStyle(tag))) {
+    if (!size && (!tag || this.isInlineStyle(tag))) {
       this.setStyleProperty('--_d', undefined, 'inline-block');
     } else {
       this.setStyleProperty(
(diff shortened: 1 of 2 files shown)


libs/base/ui/structure/heading/src/typography.controller.ts
```diff
@@ -37,28 +37,52 @@ export class TypographyController implements ReactiveController {
...
+   * Sets the display property for the heading, based on the provided tag and size.
+   * When no tag is provided, the default display is `inline`.
+   * When no size is provided, the default display is `inline-block`.
+   *
+   */
+  protected setDisplay(
+    tag?: HeadingTag | HeadingVisibility,
+    size?: Size
+  ): void {
+    if (!tag || (!size && this.isInlineStyle(tag))) {
```
Sinucid: Is the description for this method correct? Description says when no tag -> display: 'inline', but according the implementation - it is `inline-block`

Sinucid (changes requested):

Sinucid (dismissed):

libs/base/ui/structure/heading/src/typography.controller.ts
```diff
@@ -46,7 +46,9 @@ export class TypographyController implements ReactiveController {
    * Sets the font properties for the heading, based on the provided tag and size.
    */
   protected setFont(tag?: HeadingTag | HeadingVisibility, size?: Size): void {
-    if (!tag) return;
+    if (!tag) {
```
Sinucid: What was wrong with prev version?

libs/base/ui/structure/heading/src/typography.controller.ts
```diff
@@ -61,7 +63,7 @@ export class TypographyController implements ReactiveController {
     tag?: HeadingTag | HeadingVisibility,
     size?: Size
   ): void {
-    if (!tag || (!size && this.isInlineStyle(tag))) {
+    if ((!tag && !size) || (!size && this.isInlineStyle(tag))) {
```
Sinucid: `if (!size && (!tag || this.isInlineStyle(tag))) {`

Sinucid (changes requested):

libs/base/ui/structure/heading/src/typography.controller.ts
tobi-or-not-tobi: nothing, left this as part of refactor. rolling back

Sinucid (approved):

PR #966, opened 2023-12-05
feat(cart): cart list components

Sinucid:
Cart list and cart list item components implementations, with storybook stories and unit tests

closes: [HRZ-90080](https://spryker.atlassian.net/browse/HRZ-90080)

[HRZ-90080]: https://spryker.atlassian.net/browse/HRZ-90080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/cart/list-item/list-item.component.ts
```diff
@@ -0,0 +1,117 @@
...
+  implements CartListItemProperties
+{
+  static styles = cartListItemStyles;
+
+  protected cartService = resolve(CartService);
+
+  @property({ type: Boolean, reflect: true }) open?: boolean;
+
+  protected setDefault(): void {
+    this.cartService.setDefaultCart({ cartId: this.$cart()!.id }).subscribe();
```
tobi-or-not-tobi:
```suggestion
    this.cartService.setDefaultCart({ cartId: this.$cart()?.id }).subscribe();
```

libs/domain/cart/list-item/list-item.component.ts
```diff
@@ -0,0 +1,117 @@
...
+            ${cart.name}
+            ${this.i18n('cart.cart.totals.<count>-items', {
+              count: totalQuantity,
+            })}
+          </a>
+        </oryx-link>
+        ${when(
+          cart.isDefault,
+          () =>
+            html`<oryx-chip slot="heading" appearance="success">
```
tobi-or-not-tobi:
```suggestion
            html`<oryx-chip slot="heading" .appearance=${AlertType.Success}>
```

libs/domain/cart/list-item/list-item.component.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  protected cartService = resolve(CartService);
+
+  @property({ type: Boolean, reflect: true }) open?: boolean;
+
+  protected setDefault(): void {
+    this.cartService.setDefaultCart({ cartId: this.$cart()!.id }).subscribe();
+  }
+
+  protected override render(): TemplateResult | void {
```
tobi-or-not-tobi, line 30: let's break the method open for better maintainability and extensibility:
- entries
- meta (or footer)
- actions

libs/domain/cart/list-item/list-item.component.ts
```diff
@@ -0,0 +1,117 @@
...
+          )</span
+        >
+
+        ${when(
+          !totalQuantity,
+          () => html`<p>${this.i18n('cart.list.no-cart-entries')}</p>`,
+          () => html`
+            <oryx-cart-entries cartId=${cart.id}></oryx-cart-entries>
+          `
+        )}
```
tobi-or-not-tobi: Can we only render this when the collapsible is open? It means that we won't load entries while they're not needed, which saves us from initial loading.

libs/domain/cart/list-item/list-item.component.ts
tobi-or-not-tobi: we can also consider to move actions out in a component. That will have similar effect.

libs/domain/cart/list/list.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  static styles = cartListStyles;
+  protected cartService = resolve(CartService);
+  protected $carts = signal(this.cartService.getCarts());
+
+  protected override render(): TemplateResult | void {
+    const carts = this.$carts();
+
+    return html`
+      <section>
+        <oryx-heading tag=${HeadingTag.H1} typography=${HeadingTag.H3}>
```
tobi-or-not-tobi:
```suggestion
        <oryx-heading .tag=${HeadingTag.H1} .typography=${HeadingTag.H3}>
```

libs/domain/cart/list/list.component.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class CartListComponent extends ContentMixin(LitElement) {
+  static styles = cartListStyles;
+  protected cartService = resolve(CartService);
+  protected $carts = signal(this.cartService.getCarts());
+
+  protected override render(): TemplateResult | void {
+    const carts = this.$carts();
+
+    return html`
```
tobi-or-not-tobi: What about keep this component only focussing on the list and move out heading and text:
- CartListTitleComponent
- standard text component

libs/domain/cart/src/models/cart-qualifier.ts
```diff
@@ -24,6 +24,7 @@ export interface CreateCartQualifier {
 }
 
 export interface UpdateCartQualifier extends CartEntryQualifier {
+  name?: string;
   priceMode?: string;
   version?: string;
   isDefault?: true;
```
tobi-or-not-tobi:
```suggestion
  isDefault?: boolean;
```

libs/domain/cart/services/src/default-cart.service.ts
```diff
@@ -291,7 +291,7 @@ export class DefaultCartService implements CartService {
     return this.executeWithOptionalCart(qualifier, this.updateCartCommand$);
   }
 
-  setActiveCart(qualifier: UpdateCartQualifier): Observable<unknown> {
+  setDefaultCart(qualifier: UpdateCartQualifier): Observable<unknown> {
```
tobi-or-not-tobi: do we need this method as the qualifier for update already allows to do this?

tobi-or-not-tobi (changes requested):

libs/domain/cart/list-item/list-item.component.ts
Sinucid: Broke render into few methods. Not sure, that we need create separate components for action buttons

libs/domain/cart/list-item/list-item.component.ts
Sinucid: Currently is not possible, collapsible does not emit any event to check changing of the `open` state. But we can provide oryx.toggle event that will be emitted on collapsible toggle, up to you

libs/domain/cart/services/src/default-cart.service.ts
Sinucid: I was thinking about it, and made a conclusion to make an additional method for this. But now, I'm not sure. Lets drop it

libs/domain/cart/list-item/list-item.component.ts
tobi-or-not-tobi: 1. It's not so flexible if you only want to change the content of the text (say add a component in between title and list)
2. Our list component contains layout styles (e.g. flex, gap)
3. We have this pattern already established else where (cart list title, order list title)
Hence I prefer to end up with the same.

libs/domain/cart/list-item/list-item.component.ts
tobi-or-not-tobi: but we have the state in this component

libs/domain/cart/list-item/list-item.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export class CartListItemComponent
+  extends CartComponentMixin(I18nMixin(LitElement))
+  implements CartListItemProperties
+{
+  static styles = cartListItemStyles;
+
+  protected cartService = resolve(CartService);
+
+  @property({ type: Boolean, reflect: true }) open?: boolean;
```
tobi-or-not-tobi:
```suggestion
  @property({ type: Boolean }) open?: boolean;
```

tobi-or-not-tobi (changes requested):

libs/domain/cart/list-item/list-item.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  protected setDefault(): void {
+    this.cartService
+      .updateCart({
+        cartId: this.$cart()?.id,
+        isDefault: true,
+      })
+      .subscribe();
```
dunqan: subscribe is not needed (unless you want to react on error, or wait for the successuf execution)

libs/domain/cart/list-item/list-item.component.ts
Sinucid: refactored list component and carts page

libs/domain/cart/list-item/list-item.component.ts
Sinucid: yes, but we don't know when collapsible (who control the state) is changing it

libs/domain/cart/list/list.component.ts
Sinucid: refactored

libs/domain/cart/list/list.component.ts
```diff
@@ -0,0 +1,32 @@
...
+export class CartListComponent extends LayoutMixin(ContentMixin(LitElement)) {
+  protected cartService = resolve(CartService);
+  protected $carts = signal(this.cartService.getCarts());
+
+  protected override render(): TemplateResult | void {
+    const carts = this.$carts();
+
+    // return html`
+
+    // `;
```
tobi-or-not-tobi: leftover

libs/domain/cart/list/list.component.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    return this.renderLayout({
+      template: html`
+        ${repeat(
+          carts ?? [],
+          ({ id }) => id,
+          ({ id }) =>
+            html`<oryx-cart-list-item cartId=${id}></oryx-cart-list-item>`
+        )}
+        ${unsafeHTML(`<style>${this.layoutStyles()}</style>`)}
```
tobi-or-not-tobi: Styles are rendered as part of renderLayout. I see the confusion, you can deprecate `layoutStyles` in the mixin interface if you like, so we won't do this again. Will talk to Alexey about it though.

libs/template/presets/b2b-storefront/experience/pages/my-account/carts-page.ts
```diff
@@ -0,0 +1,84 @@
...
+        },
+        {
+          type: 'oryx-content-text',
+          content: {
+            data: {
+              text: `
+            <p style="text-wrap:initial; margin:0;">
+              Manage your carts with ease. Create, track, and shop effortlessly. 
+              Need help? Our support team is here for you. Enjoy your shopping!
+            </p>
```
tobi-or-not-tobi: what about just dropping the p tag?

libs/template/presets/b2b-storefront/app.ts
```diff
@@ -1,9 +1,9 @@
 import { AppFeature } from '@spryker-oryx/core';
 import { provideExperienceData } from '@spryker-oryx/experience';
 import { storefrontFeatures } from '@spryker-oryx/presets/storefront';
-import { cartsPage } from './experience/carts-page';
 import { contactPage } from './experience/contact-page';
 import { HeaderTemplate } from './experience/header';
+import { cartsPage } from './experience/pages';
```
tobi-or-not-tobi, line 6: Wonder if it should be part of generic pages as from oryx pov multi-cart is not b2b only

tobi-or-not-tobi (changes requested):

libs/template/presets/b2b-storefront/app.ts
Sinucid: Might be. Then we have to introduce project configuration first that Stan suggested, ie: 
```
.withOptions({
    ['cart.multi']: true,
  })
```

Lets merge this as is, and I will create additional subtask to organise pages and make them visible according the project config

libs/domain/cart/list/list.component.ts
Sinucid: Styles was dropped. I checked other components, and some of them are using the same pattern with layoutStyles. If there is something obsolete, better to deprecate it and make some cleanup in the separate ticket, because I'm not sure when this story will be ready to be merged to dev, there ads too many missed things

tobi-or-not-tobi (approved):

PR #968, opened 2023-12-06
fix: fix broken carousel navigation on Safari

tobi-or-not-tobi:
- The navigation is now visible again in Safari.
- The `focusin` event was changed to `input` to fix issues in Safari.
- The scrollbar invisible on Firefox.
- The `margin-block-bottom` was changed to `padding-block-bottom` for the carousel indicators at the bottom to ensure proper alignment of other carousel navigation elements.
- Repainting of the carousel and its navigation was done at the end of the resizing. 

closes: [HRZ-90723](https://spryker.atlassian.net/browse/HRZ-90723)

[HRZ-90723]: https://spryker.atlassian.net/browse/HRZ-90723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -123,13 +127,16 @@ export class CarouselNavigationComponent
...
     this.resizeObserver = new ResizeObserver(
-      throttle(() => {
-        this.buildNavigation();
-      }, throttleTime)
+      throttle((entries: ResizeObserverEntry[]) => {
+        window.requestAnimationFrame((): void | undefined => {
+          if (!Array.isArray(entries) || !entries.length) return;
+          setTimeout(
+            () => this.buildNavigation(),
+            this.intersectionThrottleTime
```
Sinucid: Why do we need to set timeout for such long period? The resize is already throttled

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -199,7 +207,7 @@ export class CarouselNavigationComponent
     this.toggleAttribute(
```
Sinucid, line 202: Just for better readability, and replace, please, **a** and **b** with proper names for the consts

```
const a = getScrollDimensions(this.hostElement, this.isVertical).position +
                 this.hostElement.getBoundingClientRect().width;
const b = getScrollDimensions(this.hostElement, this.isVertical).size -
        getDimensions(this.items?.[0], this.isVertical).size;

this.toggleAttribute('has-next', a <= b);

```

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -304,15 +313,10 @@ export class CarouselNavigationComponent
       ${when(
         this.showIndicators,
         () => html`
-          <div class="indicators">
+          <div class="indicators" @input=${this.handleIndicator}>
```
Sinucid: why is input event handled by div but not by input?

libs/platform/experience/carousel-navigation/util.ts
```diff
@@ -2,8 +2,9 @@ export function getDimensions(
   el: HTMLElement,
   isVertical = false
 ): { position: number; size: number } {
-  const { left, top, width, height } = el.getBoundingClientRect();
-  const size = isVertical ? height : width;
+  const { left, top } = el.getBoundingClientRect();
+  const style = window.getComputedStyle(el);
+  const size = isVertical ? parseFloat(style.height) : parseFloat(style.width);
```
Sinucid, line 7: What is wrong with width and height from DomRect?

Sinucid (changes requested):
Navigation arrows are no more centred relatively content, and don't forget to update unit-test according this changes =)

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: We throttle to avoid too many repaints, however, with this approach we're often missing the latest resize. To ensure we're always taken the latest resized version, we're adding a timeout before rendering the latest time. 

I'm doing a small refactor though, to align the various throttle times, as it's overly complex now.

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: Input events bubble up, hence we only need a single listener on the parent element. We can also handle it on the host, but it's more clear here.

libs/platform/experience/carousel-navigation/util.ts
tobi-or-not-tobi: `getBoundingClientRect()` includes the padding and border when the element is using `box-sizing: border-box`. That's not working.

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
Sinucid: Sounds not very stable

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
Sinucid: Yup, but inputs list is built in the map, you don't need to assign event listener on each of them manually, just move event listener assigning to the input. The same amount of code, but much more clear from developer's perspective

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: I think it is, since the throttle and the timeout are the same. If 1ms before the last throttle time we trigger the repaint, we will still have the latest repaint.

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
tobi-or-not-tobi: the input event handler is on the div, which is outside the iteration

Sinucid (dismissed):

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -123,36 +117,41 @@ export class CarouselNavigationComponent
...
-    this.style.setProperty('--width', `${this.hostElement.clientWidth}px`);
-    this.style.setProperty('--height', `${this.hostElement.clientHeight}px`);
-    if (this.showIndicators)
-      this.hostElement.style.setProperty(
-        'margin-block-end',
-        'var(--indicator-area-height, 50px)'
-      );
+    this.style.setProperty(
+      '--width',
+      `${getDimensions(this.hostElement, false).size}px`
```
Sinucid:
```suggestion
      `${getDimensions(this.hostElement).size}px`
```

libs/platform/experience/carousel-navigation/carousel-navigation.component.ts
```diff
@@ -377,9 +375,17 @@ export class CarouselNavigationComponent
...
+   * @param @deprecated use handleIndicator instead
    */
   protected handleIndicatorClick(e: Event): void {
+    this.handleIndicator(e);
+  }
+
+  /**
+   * Handles the click on an indicator.
+   */
+  protected handleIndicator(e: Event): void {
```
Sinucid: **handleIndicatorInput?** or **onIndicatorInput**, or any other more specific name for this method. And this can be handled not by click only, keyboard navigation is doing the same, check the description of it

Sinucid (changes requested):

Sinucid (approved):

PR #970, opened 2023-12-06
feat: ref saving

supproduction:
closes [HRZ-90673](https://spryker.atlassian.net/browse/HRZ-90673)

[HRZ-90673]: https://spryker.atlassian.net/browse/HRZ-90673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/platform/experience/src/services/experience/default-experience.service.ts
```diff
@@ -104,14 +104,22 @@ export class DefaultExperienceService implements ExperienceService {
   }
 
   protected reloadComponent(uid: string): void {
+    /**
+     * @deprecated Since version 1.4. Use provided `ExperienceAdapter.get` method.
+     */
```
tobi-or-not-tobi: implementation detail, hidden in the method, why do we deprecate?

libs/platform/experience/composition/composition-components.controller.ts
```diff
@@ -46,17 +46,54 @@ export class CompositionComponentsController implements ReactiveController {
...
-      .pipe(
-        switchMap((uid) =>
-          uid
-            ? this.experienceService
-                .getComponent({ uid })
-                .pipe(map((component) => component.components ?? null))
-            : of(null)
-        )
-      );
+    return this.observe.get('uid').pipe(
```
tobi-or-not-tobi, line 49: want to add some tests around it? Maybe inside service

tobi-or-not-tobi (changes requested):

libs/platform/experience/src/services/experience/default-experience.service.ts
supproduction: It's protected method so available for user. And also user might not have injected `ExperienceAdapter` which we introduced on 1.1 version so the have fallback for old http server url

tobi-or-not-tobi: But than i'd expect it to be on the header of the method

tobi-or-not-tobi (approved):

PR #980, opened 2023-12-08
feat(site): Added day component

tobi-or-not-tobi:
The day component takes the english weekday as an input and creates the localised variation for the current locale. 

closes: [HRZ-90756](https://spryker.atlassian.net/browse/HRZ-90756)

[HRZ-90756]: https://spryker.atlassian.net/browse/HRZ-90756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/domain/site/day/day.def.ts
+++ b/libs/domain/site/day/day.def.ts
@@ -1,0 +1,7 @@
+import { componentDef } from '@spryker-oryx/utilities';
+
+export const siteDayComponent = componentDef({
+  name: 'oryx-day',
+  impl: () => import('./day.component').then((m) => m.SiteDayComponent),
+  schema: import('./day.schema').then((m) => m.siteDaySchema),
+});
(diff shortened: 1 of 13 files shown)


libs/domain/site/day/day.def.ts
```diff
@@ -0,0 +1,7 @@
+import { componentDef } from '@spryker-oryx/utilities';
+
+export const siteDayComponent = componentDef({
+  name: 'oryx-day',
```
dunqan: shouldn't be `oryx-site-day`?

libs/platform/i18n/src/lib/locale/default-locale.service.ts
```diff
@@ -112,4 +123,28 @@ export class DefaultLocaleService implements LocaleService {
       )
     );
   }
+
+  formatDay(dayName: string): Observable<string> {
```
dunqan, line 127: would be good to cover it with unit tests

tobi-or-not-tobi: done, but poorly. Writing a test for it is almost rewriting the implementation in the test.

dunqan (approved):

PR #987, opened 2023-12-12
feat: content list component

supproduction:
closes [HRZ-90753](https://spryker.atlassian.net/browse/HRZ-90753)

[HRZ-90753]: https://spryker.atlassian.net/browse/HRZ-90753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/template/presets/storefront/experience/footer.ts
```diff
@@ -34,11 +34,30 @@ const legalLinks: ExperienceComponent = {
...
+            },
+          },
+        ]
+      : [
+          link('Imprint', '/faq/imprint'),
+          link('Terms & conditions', '/article/terms-and-conditions'),
+          link('Privacy Notice', '/article/privacy'),
+          link('Data preference'),
+          link('Condition of use'),
+        ]),
```
tobi-or-not-tobi: We should keep the footer as-is (at least for now), and then merge the CMS driven content in using labs.

tobi-or-not-tobi (changes requested):

libs/domain/content/src/models/content.model.ts
```diff
@@ -19,6 +19,7 @@ export interface ContentQualifier {
 
 export interface ContentField {
   [key: string]: unknown;
+  heading?: string;
```
tobi-or-not-tobi, line 22: we won't need this with entity fields, right? can we leave it out already?

tobi-or-not-tobi (changes requested):

libs/domain/content/src/models/content.model.ts
supproduction: I will fix it in model parsing

tobi-or-not-tobi (approved):

PR #990, opened 2023-12-12
feat(cart): carts pages

Sinucid:
- carts page (**/my-account/carts**) was added to the storefront preset
- TokenResolve's `resolveToken` method is accepting options object now instead string token `({ token: string, contextElement?: HTMLElement | LitElement})`, with token and optional context element that can be used as context provider
- Cart token resolver's `SUMMARY` and `EMPTY` are able to resolve values within context now
- Cart and Order totals resolvers are also resolving the totals depends on the context
- `CompositionComponentsController` is providing host element now as a `contextElement` for dynamic visibility rules, to be able to resolve resources within context

closes: [HRZ-90718](https://spryker.atlassian.net/browse/HRZ-90718),  [HRZ-90086](https://spryker.atlassian.net/browse/HRZ-90086)

[HRZ-90718]: https://spryker.atlassian.net/browse/HRZ-90718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HRZ-90086]: https://spryker.atlassian.net/browse/HRZ-90086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/src/app.ts
```diff
@@ -17,4 +17,7 @@ export const app = appBuilder()
   .withFeature(features)
   .withTheme([storefrontTheme])
   .withEnvironment(env)
+  .withOptions({
+    cart: { multi: env.ORYX_PRESET === 'b2b' || env.ORYX_MULTI_CART },
```
tolerants: Shouldn't it be on a `presets` feature level? Especially keeping in mind enabling `ORYX_MULTI_CART` can't enable this feature by itself.

libs/domain/cart/services/src/resolver/cart.resolver.ts
```diff
@@ -13,32 +20,52 @@ export class CartResolver extends BaseResolver<CartResolvers> {
   /** @deprecated since 1.1 use cartService instead*/
   protected cartService$ = resolve(CartService);
   protected cartService = inject(CartService);
+  protected contextService = inject(ContextService);
+
+  protected getContext(
+    options?: TokenResolverOptions
+  ): Observable<CartQualifier | undefined> {
+    return featureVersion < '1.4' || !options?.contextElement
```
supproduction: We use `featureVersion >= '1.4'`

libs/domain/cart/services/src/totals/cart-totals.resolver.ts
```diff
@@ -1,16 +1,38 @@
...
+    protected cartService = inject(CartService),
+    protected contextService = inject(ContextService)
+  ) {}
 
-  getTotals(): Observable<NormalizedTotals | null> {
-    return this.cartService.getCart().pipe(
+  protected getContext(
+    options?: TotalsResolverOptions
+  ): Observable<CartQualifier | undefined> {
+    return featureVersion < '1.4' || !options?.contextElement
```
supproduction: same rule for `featureVersion`

supproduction (changes requested):

libs/domain/cart/services/src/resolver/cart.resolver.ts
dunqan: Do we have to consider it as a braking change?
Is it affecting anything in versions <1.4?

libs/domain/cart/services/src/totals/cart-totals.resolver.ts
dunqan: same as before, is the result different if default context is current cart anyway?

libs/domain/cart/src/controllers/totals.controller.ts
```diff
@@ -30,7 +30,7 @@ export class TotalsController implements ReactiveController {
     return defer(() =>
       this.context.get<string>(TotalsContext.Reference).pipe(
         switchMap((context) => {
-          return this.totalsService.get(context);
+          return this.totalsService.get(context, { contextElement: this.host });
```
dunqan: 1. What about having TotalService.get({ context,, element, ... }) for better upgradability in the future. 
2. What's better, passing element or cart id to the totals service? Just question. TotalService is mostly used inside component to solve totals from context, so passing element may be better indeed

libs/platform/core/src/services/token-resolver/token-resolver.service.ts
```diff
@@ -7,14 +8,26 @@ export type ResolvedToken = Observable<
...
 
-export type Resolver = () => Observable<ResolvedResult>;
+export type Resolver = (
+  options?: TokenResolverOptions
+) => Observable<ResolvedResult>;
 
 export interface TokenResolver {
-  resolveToken(token: string): Observable<ResolvedResult>;
+  resolveToken(
+    token: string,
```
dunqan: again, looks like better approach is to go with object instead of multiple parameters

apps/storefront/src/app.ts
Sinucid: Dropped for current implementation

libs/domain/cart/services/src/resolver/cart.resolver.ts
Sinucid: dropped

libs/domain/cart/services/src/totals/cart-totals.resolver.ts
Sinucid: Dropped

libs/domain/cart/src/controllers/totals.controller.ts
Sinucid: 1. fixed
2. yup, I don't think that we will use this service outside totals components

libs/template/presets/b2b-storefront/app.ts
```diff
@@ -3,10 +3,9 @@ import { provideExperienceData } from '@spryker-oryx/experience';
 import { storefrontFeatures } from '@spryker-oryx/presets/storefront';
 import { contactPage } from './experience/contact-page';
 import { HeaderTemplate } from './experience/header';
-import { cartsPage } from './experience/pages';
 
 const StaticB2BExperienceFeature: AppFeature = {
-  providers: [provideExperienceData([contactPage, HeaderTemplate, cartsPage])],
+  providers: [provideExperienceData([contactPage, HeaderTemplate])],
```
tolerants, line 8: But why it's not under feature version?

supproduction (approved):

libs/template/presets/b2b-storefront/app.ts
Sinucid: It's not a BC. Page was added as a part of another sub-task of this feature, but in this one I moved it to the main storefront preset

tolerants (approved):

libs/domain/cart/services/src/totals/default-totals.service.ts
```diff
@@ -13,14 +14,18 @@ export class DefaultTotalsService implements TotalsService {
     return `${TotalsResolver}${ref ?? 'CART'}`;
   }
 
-  get(context: string): Observable<NormalizedTotals | null> {
+  get(
+    contextOptions?: TotalsOptions | string
```
dunqan, line 21: let's use overloading, and deprecate signature with `string` only

libs/domain/cart/src/services/totals/totals.service.ts
```diff
@@ -10,8 +11,19 @@ export const TotalsResolver = 'oryx.TotalsResolver*';
...
+    options?: TotalsResolverOptions
+  ): Observable<NormalizedTotals | null>;
+}
+
+export interface TotalsOptions extends TotalsResolverOptions {
+  totalsContext?: string;
+}
+
+export interface TotalsResolverOptions {
+  contextElement?: HTMLElement | LitElement;
```
dunqan: just `element`?

libs/domain/cart/src/services/totals/totals.service.ts
```diff
@@ -10,8 +11,19 @@ export const TotalsResolver = 'oryx.TotalsResolver*';
...
 
 export interface TotalsResolver {
-  getTotals(): Observable<NormalizedTotals | null>;
+  getTotals(
+    options?: TotalsResolverOptions
+  ): Observable<NormalizedTotals | null>;
+}
+
+export interface TotalsOptions extends TotalsResolverOptions {
+  totalsContext?: string;
```
dunqan: just `context`? as it's part of (Totals)Options already?

dunqan (approved):

PR #994, opened 2023-12-13
feat(content): move cms adapters

supproduction:
closes [HRZ-90774](https://spryker.atlassian.net/browse/HRZ-90774)

[HRZ-90774]: https://spryker.atlassian.net/browse/HRZ-90774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/content/package.json
```diff
@@ -23,6 +23,8 @@
     "@spryker-oryx/utilities": "1.3.1"
   },
   "peerDependencies": {
+    "marked": "^5.1.0",
+    "@types/marked": "^5.0.0",
```
tobi-or-not-tobi: can we bump to latest version. `11.0.0`. Also, let's use major versions only as we're using caret, unless the package does not respect semver.

libs/domain/content/src/services/cms/storyblok/storyblok-content.adapter.ts
tobi-or-not-tobi, line 1: We use `version=draft` which seems wrong.

libs/domain/content/src/services/cms/strapi/strapi-content.adapter.ts
```diff
@@ -29,7 +25,7 @@ export interface TypeData {
 export class DefaultStrapiContentAdapter implements ContentAdapter {
   constructor(
     protected token = inject(StrapiToken),
-    protected config = inject(ContentConfig),
+    protected config = inject(ContentConfig, [] as ContentConfig[]),
```
tobi-or-not-tobi: let's move constructor arguments out already as discussed

libs/domain/search/package.json
```diff
@@ -29,5 +29,10 @@
     "@spryker-oryx/site": "1.3.1",
     "@spryker-oryx/ui": "1.3.1",
     "@spryker-oryx/utilities": "1.3.1"
+  },
+  "peerDependencies": {
+    "lit": "^2.4.0",
+    "rxjs": "^7.5.0",
```
tobi-or-not-tobi, line 35:
```suggestion
    "lit": "^2.0.0",
    "rxjs": "^7.0.0",
```

tobi-or-not-tobi (changes requested):

libs/domain/content/list/list.component.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+
+describe('ContentListComponent', () => {
+  let element: ContentListComponent;
+
+  beforeAll(async () => {
+    await useComponent(contentListComponent);
+  });
+
+  beforeEach(async () => {
+    mockFeatureVersion('1.4');
```
Sinucid: Why do you need this?

libs/domain/content/list/list.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  protected $data = computed(() => {
+    const _options = this.$options();
+    const options =
+      _options.type || _options.query || _options.tags ? _options : undefined;
+    const context = this.$context();
+    const props =
+      this.type || this.query || this.tags
+        ? { type: this.type, query: this.query, tags: this.tags }
+        : undefined;
+    const data = props ?? options ?? context;
```
Sinucid, line 41: To call the context signal as a fallback only when it will be necessary
```suggestion
    const props =
      this.type || this.query || this.tags
        ? { type: this.type, query: this.query, tags: this.tags }
        : undefined;
    const data = props ?? options ?? this.$context();
```

libs/domain/content/src/models/content.model.ts
```diff
@@ -13,17 +13,30 @@ export interface ContentQualifier {
...
 }
 
-export interface Content<T = Record<string, unknown>> {
-  fields: T & ContentField;
-  id: string;
-  type: string;
+/**
+ * @deprecated Since version 1.4. Will be removed.
+ */
+interface DeprecatedContent {
```
Sinucid, line 29: I didn't get. This interface has been just added and already deprecated?

libs/domain/content/src/services/adapter/content-experience.adapter.spec.ts
```diff
@@ -13,6 +13,7 @@ describe('ContentExperienceAdapter', () => {
   let contentService: ContentService;
 
   beforeEach(() => {
+    mockFeatureVersion('1.4');
```
Sinucid, line 16: Don't you need to test < 1.4 as well?

Sinucid (changes requested):

libs/domain/content/src/models/content.model.ts
supproduction: We use this interface in Content type, it's for readability purpose rather then I deprecate each fields in the main one

libs/domain/search/package.json
supproduction: I would say we should do it globally in separate ticket, cause might have problem with inconsistency

libs/domain/content/list/list.component.ts
supproduction: I am going to adjust the whole approach in the next ticket

libs/domain/content/src/services/adapter/content-experience.adapter.spec.ts
supproduction: I think it will be redundant tests, cause it's a bit artificial versioning

Sinucid: The idea is to ensure that the < 1.4 functionality is still working as was

tobi-or-not-tobi: Ideally we would test different versions, true. We're not doing it often though.

tobi-or-not-tobi (approved):

Sinucid (approved):

PR #995, opened 2023-12-13
feat(experience): Added image style plugin

tobi-or-not-tobi:
The image object-fit and object-position properties can be controlled with layout styles, so we no longer need them specifically in components. 

closes: [HRZ-90786](https://spryker.atlassian.net/browse/HRZ-90786)

[HRZ-90786]: https://spryker.atlassian.net/browse/HRZ-90786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/content/image/image.component.ts
```diff
@@ -1,8 +1,9 @@
 import { ContentMixin, defaultOptions } from '@spryker-oryx/experience';
-import { html, LitElement, TemplateResult } from 'lit';
+import { featureVersion } from '@spryker-oryx/utilities';
+import { LitElement, TemplateResult, html } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
 import { ContentImageContent, ContentImageOptions } from './image.model';
 import { contentImageStyles } from './image.styles';
-import { ifDefined } from 'lit/directives/if-defined.js';
 
 @defaultOptions({ fit: 'cover' })
```
Sinucid: I guess
```suggestion
@defaultOptions(featureVersion >= '1.4'? {}: { fit: 'cover' })
```

libs/domain/content/image/image.component.ts
```diff
@@ -39,7 +40,7 @@ export class ContentImageComponent extends ContentMixin<
   protected getStyles(): string | undefined {
     const { fit, position } = this.$options();
     let styles = '';
-    if (fit) styles += `--image-fit:${fit};`;
+    if (featureVersion < '1.4' && fit) styles += `--image-fit:${fit};`;
     if (position) styles += `--image-position:${position};`;
```
Sinucid: What about this one? According the Pr's desc `position` should be specified via layouts as well

libs/platform/experience/src/services/layout/plugins/layout-plugins.providers.ts
```diff
@@ -47,6 +47,13 @@ export const layoutPluginsProviders: Provider[] = [
         (m) => m.SpacingStylePlugin
       ),
   },
+  {
+    provide: LayoutStylesPlugin,
+    asyncClass: () =>
+      import('./styles/image/image-style.plugin').then(
```
Sinucid: Maybe better to choose another name for this plugin? `ObjectsStylesPlugin` for example, because this styles can be potentially applied for any 'object-like' elements: iframes, video, ...

libs/platform/experience/src/services/layout/plugins/styles/image/image-style.plugin.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    const properties: {
+      '--image-fit'?: string;
+      '--image-position'?: string;
+    } = {};
+
+    properties['--image-fit'] ??= styles.imageFit;
+    properties['--image-position'] ??= styles.imagePosition;
+
+    return of(properties);
```
Sinucid:
```suggestion
    const { imageFit, imagePosition  } = data.styles;

    const properties = {
      ...(imageFit ? {'--image-fit': imageFit}: {}),
      ...(imagePosition ? {'--image-position': imagePosition}: {}),
    };

    return of(properties);
```

libs/platform/experience/src/services/layout/plugins/styles/image/index.ts
```diff
@@ -0,0 +1 @@
+export * from './image-style.model';
```
Sinucid: Are we not exporting the schemas and plugin files from the plugins? How to extend them then?

Sinucid (changes requested):

libs/platform/experience/src/services/layout/plugins/styles/image/index.ts
tobi-or-not-tobi: I forgot. We need to add it to static exports (`static.ts`)

Sinucid (dismissed):

Sinucid (approved):

PR #997, opened 2023-12-14
fix: custom colors in dark mode for logo's

tobi-or-not-tobi:
In a previously PR we fixed a few things related to dark mode and logo's, but also introduced new bugs. Additionally, we add a few small improvements to some of the logo's in the footer for a better presentation.


libs/base/ui/graphical/image/src/image.styles.ts
```diff
@@ -32,9 +32,9 @@ export const styles = css`
     object-position: var(--image-position);
   }
 
-  @media (prefers-color-scheme: dark) {
+  /* @media (prefers-color-scheme: dark) {
```
tolerants: Can we just remove it?

libs/platform/experience/src/services/layout/plugins/styles/canvas/canvas-style.model.ts
```diff
@@ -6,6 +6,13 @@ declare global {
      */
     background?: string;
 
+    /**
+     * Sets the overflow behavior of an element's content. If an element's content is larger than
```
tolerants, line 10: Does it really make sense to explain css properties? In any case we are not providing possible values and not affecting default implementation in the browser. Maybe we can just give a link to MDN or W3?

tolerants (changes requested):

libs/platform/experience/src/services/layout/plugins/styles/canvas/canvas-style.model.ts
tobi-or-not-tobi: We had it on other elements and hence I stayed consistent with it. The longer term vision is that we'd have docs on those attributes for both devs and users in EB. Inline docs is not working for that, I know.

tobi-or-not-tobi: lets keep it for now

tolerants (approved):

PR #1006, opened 2023-12-15
feat(cart): carts list page

Sinucid:
Carts list page that presents all available user's carts to interact with them. Available by `/my-account/carts` route.
As a part of the story were made:

- **create** and **delete** methods were added to the cart adapter to create or delete the cart
- **createCart** and **deleteCart** methods were added to the cart service
- added cart context and fallback cart context 
- Multi-cart `@spryker-oryx/cart/multi` extension for cart feature. Provides **cart-list**, **cart-list-item** components and feature option `{cart: {multi: boolean}}`. Is applied by default to B2B preset, and can be configured for B2C with **ORYX_MULTI_CART** env var
- TokenResolve's resolveToken method is accepting options object now instead string token `({ token: string, contextElement?: HTMLElement | LitElement})`, with token and optional context element that can be used as context provider
- Cart token resolver's **SUMMARY** and **EMPTY** are able to resolve values within context now
- Cart and Order totals resolvers are also resolving the totals depends on the context
- **CompositionComponentsController** is providing host element now as a contextElement for dynamic visibility rules, to be able to resolve resources within context

closes: [HRZ-89986](https://spryker.atlassian.net/browse/HRZ-89986)

[HRZ-89986]: https://spryker.atlassian.net/browse/HRZ-89986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/cart/list-item/list-item.component.ts
```diff
@@ -0,0 +1,147 @@
...
+          ></oryx-button>
+          <oryx-button icon="share" type="icon" .size=${size}></oryx-button>
+          <oryx-button
+            icon="download"
+            type="icon"
+            .size=${size}
+            ?disabled=${!disabled}
+          ></oryx-button>
+          <oryx-button icon="delete" type="icon" .size=${size}></oryx-button>
+        </div>
```
tobi-or-not-tobi: Those actions don't do anything so far, so I'd rather make them a comment for now. Also, going forward, we could consider:
- moving them out to a separate component
- make each action a separate component (e.g. `CartDownloadActionComponent`, so that we can easily integrate them on the main cart page as well.

libs/domain/cart/list-item/list-item.component.ts
```diff
@@ -0,0 +1,147 @@
...
+const size = Size.Md;
+
+export class CartListItemComponent
+  extends CartComponentMixin(I18nMixin(LitElement))
+  implements CartListItemProperties
+{
+  static styles = cartListItemStyles;
+
+  protected cartService = resolve(CartService);
+  protected linkService = resolve(LinkService);
```
tobi-or-not-tobi, line 18: let's use link service from router as it's moved recently

libs/domain/cart/list-item/list-item.component.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  @property({ type: Boolean }) open?: boolean;
```
tobi-or-not-tobi: do we need a property? I'd rather use `@state` here and leave open out of the public interface

libs/domain/cart/list-item/list-item.component.ts
```diff
@@ -0,0 +1,147 @@
...
+      ></oryx-site-price>
+
+      <span slot="heading"
+        >(
+        ${this.i18n(
+          cart.priceMode === PriceMode.GrossMode
+            ? 'cart.mode.gross'
+            : 'cart.mode.net'
+        )}
+        )</span
```
tobi-or-not-tobi: Let's keep the `(` and `)` to localisation, so we have a key (e.g. `carts.cart.price-mode.gross`) that can be used to add the paranthese.

libs/domain/cart/list-item/list-item.styles.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  .actions {
+    display: flex;
+    gap: 10px;
+    margin-inline-start: auto;
+  }
+
+  p {
+    text-wrap: initial;
```
tobi-or-not-tobi: no longer needed

libs/domain/cart/list-item/list-item.component.ts
```diff
@@ -0,0 +1,147 @@
...
+      >
+    `;
+  }
+
+  protected renderControls(): TemplateResult {
+    const disabled = !this.$totalQuantity();
+
+    return html`
+      <div class="meta">
+        ${when(
```
tobi-or-not-tobi, line 81: i'm not a big fan of using `when` directive, rather create the template outside the html (e.g. `callToActionTemplate = this.$cart()?.isDefault ? ...` and use it inside the html template. can be an inline. Also an opinionated nitpick that you can ignore ;)

libs/domain/cart/list-item/list-item.model.ts
```diff
@@ -0,0 +1,6 @@
+export interface CartListItemProperties {
+  /**
+   * Open the collapsible
+   */
+  open?: boolean;
```
tobi-or-not-tobi: prob not needed see other comment

libs/domain/cart/list-item/stories/Static/variants.stories.ts
```diff
@@ -0,0 +1,37 @@
+import { Meta, Story } from '@storybook/web-components';
+import { TemplateResult, html } from 'lit';
+import { storybookPrefix } from '../../../.constants';
+
+export default {
+  title: `${storybookPrefix}/List Item/Static`,
```
tobi-or-not-tobi:
```suggestion
  title: `${storybookPrefix}/List Item`,
```

libs/domain/cart/list-item/stories/Static/variants.stories.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    <h4>Multiple entries</h4>
+    <oryx-cart-list-item cartId="multiple"></oryx-cart-list-item>
+
+    <h4>Multiple entries opened</h4>
+    <oryx-cart-list-item cartId="multiple" open></oryx-cart-list-item>
+  `;
+};
+
+export const Variants = Template.bind({});
```
tobi-or-not-tobi:
```suggestion
export const Static = Template.bind({});
```

libs/domain/cart/list-item/stories/demo.stories.ts
```diff
@@ -0,0 +1,49 @@
...
+  parameters: {
+    chromatic: {
+      disableSnapshot: true,
+    },
+  },
+} as Meta;
+
+const Template: Story<Props> = ({ cartId, open }): TemplateResult => {
+  return html`
+    <oryx-cart-list-item ?open=${open} cartId=${cartId}></oryx-cart-list-item>
```
tobi-or-not-tobi: this might be the only reason for use to use open property, but I rather do it diff then

libs/domain/cart/list/list.component.ts
```diff
@@ -0,0 +1,26 @@
...
+  protected cartService = resolve(CartService);
+  protected $carts = signal(this.cartService.getCarts());
+
+  protected override render(): TemplateResult | void {
+    const carts = this.$carts();
+
+    return this.renderLayout({
+      template: html`
+        ${repeat(
+          carts ?? [],
```
tobi-or-not-tobi: let's return this render method when there's an emtpy cart, to avoid any layout stuff.

libs/domain/cart/list/list.schema.ts
```diff
@@ -0,0 +1,8 @@
+import { ContentComponentSchema } from '@spryker-oryx/experience';
+import { CartListComponent } from './list.component';
+
+export const cartListComponentSchema: ContentComponentSchema<CartListComponent> =
+  {
+    name: 'List',
+    group: 'Cart',
```
tobi-or-not-tobi, line 8: let's add list icon

libs/domain/cart/list/list.def.ts
```diff
@@ -0,0 +1,6 @@
+import { componentDef } from '@spryker-oryx/utilities';
+
+export const cartListComponent = componentDef({
+  name: 'oryx-cart-list',
+  impl: () => import('./list.component').then((m) => m.CartListComponent),
```
tobi-or-not-tobi, line 5: schema is missing

libs/domain/cart/multi/feature.ts
```diff
@@ -0,0 +1,10 @@
+import { AppFeature } from '@spryker-oryx/core';
+import * as components from './components';
+export * from './components';
+
+export const multiCartComponents = Object.values(components);
+
+export const multiCartFeature: AppFeature = {
```
tobi-or-not-tobi: what about integrate these multi files just in `src/feature.ts`?

libs/domain/cart/totals/src/totals.component.ts
```diff
@@ -1,8 +1,10 @@
 import { ContentMixin } from '@spryker-oryx/experience';
-import { hydrate, signal, signalAware } from '@spryker-oryx/utilities';
-import { html, LitElement, TemplateResult } from 'lit';
+import { Size, hydrate, signal, signalAware } from '@spryker-oryx/utilities';
+import { LitElement, TemplateResult, html } from 'lit';
 import { TotalsController } from '../../src/controllers';
 
+const size = Size.Lg;
```
dunqan: leftover?

libs/domain/cart/totals/src/totals.component.ts
```diff
@@ -13,7 +15,7 @@ export class CartTotalsComponent extends ContentMixin(LitElement) {
   protected override render(): TemplateResult | void {
     if (this.$totals()) {
       return html`<h2>${this.i18n('cart.totals.summary')}</h2>
-        <oryx-composition .uid=${this.uid}></oryx-composition>`;
+        <oryx-composition .uid=${this.uid}></oryx-composition> `;
```
dunqan: 👀

libs/template/presets/storefront/experience/pages/cart-page.ts
```diff
@@ -4,7 +4,17 @@ import { featureVersion } from '@spryker-oryx/utilities';
...
+    {
+      type: 'oryx-content-text',
+      content: {
+        data: {
+          text: 'You have <oryx-link><a href="/carts">3 other carts</a></oryx-link> available.',
+        },
+      },
+    },
+    { type: 'oryx-cart-entries' }
+  );
```
tobi-or-not-tobi: let's drop it as it was here only for initial poc

libs/template/presets/storefront/experience/pages/my-account/carts-page.ts
```diff
@@ -0,0 +1,68 @@
+import { ExperienceComponent } from '@spryker-oryx/experience';
+import { RouteType } from '@spryker-oryx/router';
+import { Size } from '@spryker-oryx/utilities';
+
+export const cartsPage: ExperienceComponent = {
+  id: 'carts-page',
+  type: 'Page',
+  meta: {
+    title: 'Carts',
+    route: '/my-account/carts',
```
tobi-or-not-tobi, line 10: we need to use a `routeType` here, please align with @dunqan

libs/template/presets/storefront/experience/pages/my-account/carts-page.ts
```diff
@@ -0,0 +1,68 @@
+import { ExperienceComponent } from '@spryker-oryx/experience';
+import { RouteType } from '@spryker-oryx/router';
+import { Size } from '@spryker-oryx/utilities';
+
+export const cartsPage: ExperienceComponent = {
```
tobi-or-not-tobi, line 5: We're on are way to move experience data as much as possible to the domains, and only let the presets to add the actual page. Can you follow the example of merchant here?

tobi-or-not-tobi (changes requested):

libs/domain/cart/list-item/list-item.component.ts
Sinucid: Commented for now

dunqan (changes requested):

libs/domain/cart/list-item/list-item.component.ts
Sinucid: It is a public intentionally, the idea was to provide an ability to control collapsible state from outside (to expand all carts with one button, or collapse all, expand only non-empty carts and etc), there is no any sense to make it local

libs/domain/cart/list-item/list-item.component.ts
Sinucid: I would say that we are using `when` almost everywhere and it is a bit long, but much more readable

libs/domain/cart/list-item/list-item.model.ts
Sinucid: It's needed, see other comment ;)

libs/domain/cart/list-item/stories/Static/variants.stories.ts
Sinucid: But this is a static story, we are keeping them in the separate sub-folder in storybook. Or do we have a new rule about this?

libs/domain/cart/list-item/stories/demo.stories.ts
Sinucid: not sure why

libs/domain/cart/totals/src/totals.component.ts
Sinucid: BC

libs/template/presets/storefront/experience/pages/my-account/carts-page.ts
Sinucid: But it has been already added

libs/domain/cart/list-item/list-item.component.ts
Sinucid: Oh, I missed that hardcoded brackets, fixed

libs/domain/cart/multi/feature.ts
Sinucid: The idea was to divide them. Also, using of this feature for b2c can be configurable with env var, that I believe, we should not mix in the single feature

libs/domain/cart/multi/providers.ts
```diff
@@ -0,0 +1,7 @@
+import { Provider } from '@spryker-oryx/di';
+import { provideExperienceData } from '@spryker-oryx/experience';
+import { cartsPage } from './presets';
+
+export const multiCartsProviders: Provider[] = [
+  provideExperienceData(cartsPage),
+];
```
tobi-or-not-tobi: I believe we started this wrong, me actually. Would be better to export the data from the domain, but not provide it by default. That's what we still do in presets.

libs/domain/cart/list-item/list-item.component.ts
tobi-or-not-tobi: but we do not have a case for it, right?

libs/template/labs/src/i18n/translations/en.ts
```diff
@@ -8,6 +8,17 @@ const product = {
...
 
+const carts = {
+  'carts.totals.<count>-carts':
+    'My carts ({count, plural, one {1 cart} other {{count} carts}})',
+  'carts.totals.<count>-items':
+    '({count, plural, 0 {empty} one {1 item} other {{count} items}})',
+  'carts.list.no-cart-entries':
+    'There are no cart entries for this cart available.',
+  'carts.price-mode.gross': '( Gross )',
+  'carts.price-mode.net': '( Net )',
```
tobi-or-not-tobi:
```suggestion
  'carts.price-mode.gross': '(Gross)',
  'carts.price-mode.net': '(Net)',
```

libs/domain/cart/list-item/list-item.component.ts
tobi-or-not-tobi: must be only me then, I don't agree it's better readable

libs/domain/cart/multi/feature.ts
tobi-or-not-tobi: I'd:
- drop multi folder
- keep `multiCartFeature`, just move it to `src/feature.ts`
- avoid exporting `multiCartComponents`
- move multi presets to `src/presets`
- drop providers - see other comment

libs/domain/cart/list-item/list-item.component.ts
Sinucid: not yet, but I believe the default cart should be opened by default for example, but this can be implemented of list-item level. Agree, lets drop it for now

tobi-or-not-tobi (approved):

dunqan (approved):

PR #1012, opened 2023-12-18
feat: adds context to composition

supproduction:


libs/domain/content/src/services/content-context.ts
```diff
@@ -1,3 +1,11 @@
+import { ContentQualifier } from '../models';
+
+declare global {
+  interface ContextValue {
+    [ContentContext.Qualifier]?: ContentQualifier;
+  }
+}
+
 export const enum ContentContext {
```
dunqan, line 9: what about just `export const ContentContext = 'content'` ?

libs/platform/experience/composition/composition.component.ts
```diff
@@ -66,6 +71,51 @@ export class CompositionComponent extends LayoutMixin(
...
+
+    for (const [type, context] of Object.entries(contexts ?? {})) {
+      const isContent = this.contents
+        .flat()
+        .some((_contents: Record<string, { types: string[] }>) =>
+          Object.values(_contents).some((value) => value.types.includes(type))
+        );
+
+      if (isContent) {
+        types.push('content-qualifier');
```
dunqan: do we need special handlign for content here? can't we do it on EB side?

dunqan (dismissed):

tobi-or-not-tobi (approved):

PR #1013, opened 2023-12-18
feat: add image to contentful cms

supproduction:


libs/domain/content/src/services/cms/contentful/normalizers/contentful-field.normalizer.ts
```diff
@@ -1,11 +1,21 @@
...
   value: unknown;
+  assets?: Record<string, ContentfulAssets>;
+}
+
+export interface ContentfulLinkAsset {
+  sys: {
+    id: string;
+    linkType: string;
+    type: string;
+  };
```
tobi-or-not-tobi, line 18: what about moving those to `ContentfulCmsModel`?

tobi-or-not-tobi (approved):

PR #1022, opened 2023-12-20
feat(experience): Added entity link component

tobi-or-not-tobi:
The entity link component can be used to render any link from the data on the screen. When there's a link to an email or phone number, the required protocol will be added. 

closes: [HRZ-90823](https://spryker.atlassian.net/browse/HRZ-90823)

[HRZ-90823]: https://spryker.atlassian.net/browse/HRZ-90823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dunqan (dismissed):

libs/domain/merchant/src/mocks/merchant.service.ts
```diff
@@ -9,6 +9,12 @@ export class MockMerchantService implements Partial<MerchantService> {
       name: 'Merchant 1',
       banner:
         'https://d2s0ynfc62ej12.cloudfront.net/merchant/sonyexperts-banner.png',
+      url: 'https://www.sony.de/experts',
```
Sinucid: Should we use real web address and email for the mocked data?

libs/platform/experience/entity-link/entity-link.component.ts
```diff
@@ -0,0 +1,50 @@
...
+export class EntityLinkComponent extends ContentMixin<EntityLinkOptions>(
+  LitElement
+) {
+  protected entityService = resolve(EntityService);
+
+  protected $data = computed<string | undefined>(() => {
+    const { entity: type, field } = this.$options();
+    return this.entityService
+      .getField<string>({ element: this, type, field })
+      .pipe(catchError(() => of()));
```
Sinucid, line 25: Maybe better to perform error handling on service level to avoid rxjs in components?

libs/platform/experience/entity-link/entity-link.component.ts
```diff
@@ -0,0 +1,50 @@
...
+      </oryx-link>
+    `;
+  }
+
+  /**
+   * Resolves the protocol for the given url, in case it's a phone number or email address.
+   * Phone numbers will get the `tel:` protocol, email addresses the `mailto:` protocol.
+   */
+  protected resolveProtocol(url: string): string {
+    const phoneRegex = /^[+]*[(]{0,1}[0-9]{1,4}[)]{0,1}[-\s./0-9()]*$/im;
```
Sinucid: I believe that these are common regexps, maybe better to keep them in utilities?

libs/platform/experience/entity-link/stories/demo.stories.ts
```diff
@@ -0,0 +1,30 @@
...
+    field: 'contact.phone',
+    icon: IconTypes.Phone,
+    label: 'Call us',
+  },
+  parameters: { chromatic: { disableSnapshot: true } },
+} as Meta;
+
+type Props = EntityLinkOptions;
+
+const Template: Story<Props> = (props: Props): TemplateResult => {
```
Sinucid:
```suggestion
const Template: Story<EntityLinkOptions> = (props): TemplateResult => {
```

Sinucid (changes requested):

libs/platform/experience/entity-link/entity-link.component.ts
Sinucid: What about this one? I saw the same case inside Stan's PR, he replied that handling errors inside services can limit us in covering some specific cases on place, but I'm not sure that this is also a right way, should we move error handling to the components?

tobi-or-not-tobi: Let's leave it here for now. We're not sure if we want to hide it in service layer as of now.

Sinucid (approved):

PR #1024, opened 2023-12-20
feat(experience): link support for entity text

dunqan:
Includes:
- link support for oryx-entity-text component
- ability to resolve entity qualifier from EntityService

closes [HRZ-90347](https://spryker.atlassian.net/browse/HRZ-90347)

[HRZ-90347]: https://spryker.atlassian.net/browse/HRZ-90347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/experience/entity-text/entity-text.component.ts
+++ b/libs/platform/experience/entity-text/entity-text.component.ts
@@ -35,13 +35,12 @@ export class EntityTextComponent extends TextMixin(
     const { entity: type } = this.$options();
 
     return this.entityService.getQualifier({ element: this, type }).pipe(
-      switchMap(({ type, qualifier }) => {
-        console.log(qualifier, type);
-        return this.linkService.get({
+      switchMap(({ type, qualifier }) =>
+        this.linkService.get({
           type,
           qualifier,
-        });
-      }),
+        })
+      ),
       catchError(() => of())
     );
   });
--- a/libs/platform/experience/entity-text/entity-text.component.ts
+++ b/libs/platform/experience/entity-text/entity-text.component.ts
@@ -1,4 +1,4 @@
-import { ContextService, EntityService } from '@spryker-oryx/core';
+import { EntityService } from '@spryker-oryx/core';
 import { resolve } from '@spryker-oryx/di';
 import { ContentMixin } from '@spryker-oryx/experience';
 import { LinkService } from '@spryker-oryx/router';
@@ -17,7 +17,6 @@ export class EntityTextComponent extends TextMixin(
 ) {
   protected entityService = resolve(EntityService);
   protected linkService = resolve(LinkService);
-  protected contextService = resolve(ContextService);
 
   protected $data = computed<string | undefined>(() => {
     const { entity: type, field } = this.$options();
@@ -42,7 +41,8 @@ export class EntityTextComponent extends TextMixin(
           type,
           qualifier,
         })
-      )
+      ),
+      catchError(() => of())
     );
   });
 
--- a/libs/platform/experience/entity-text/entity-text.component.ts
+++ b/libs/platform/experience/entity-text/entity-text.component.ts
@@ -68,7 +68,6 @@ export class EntityTextComponent extends TextMixin(
     | void {
     const text = this.$text();
     const link = this.$link();
-    console.log(link);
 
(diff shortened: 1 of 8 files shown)


libs/platform/experience/entity-text/entity-text.component.ts
```diff
@@ -33,14 +32,18 @@ export class EntityTextComponent extends TextMixin(
...
+    const { entity: type } = this.$options();
 
-    const qualifier = this.contextService.get(this, field);
-    return this.linkService.get({
-      type: entity,
-      qualifier,
-    });
+    return this.entityService.getQualifier({ element: this, type }).pipe(
+      switchMap(({ type, qualifier }) => {
+        console.log(qualifier, type);
```
Sinucid: console.log

libs/platform/experience/entity-text/entity-text.component.ts
```diff
@@ -33,14 +32,18 @@ export class EntityTextComponent extends TextMixin(
...
+        return this.linkService.get({
+          type,
+          qualifier,
+        });
+      }),
+      catchError(() => of())
```
Sinucid: Can we move error handling to the service?

libs/platform/experience/entity-text/entity-text.component.ts
```diff
@@ -33,14 +32,18 @@ export class EntityTextComponent extends TextMixin(
...
 
-    const qualifier = this.contextService.get(this, field);
-    return this.linkService.get({
-      type: entity,
-      qualifier,
-    });
+    return this.entityService.getQualifier({ element: this, type }).pipe(
+      switchMap(({ type, qualifier }) => {
+        console.log(qualifier, type);
+        return this.linkService.get({
```
Sinucid: Can we decrease amount of rxjs in component?

libs/platform/experience/entity-text/entity-text.component.ts
```diff
@@ -68,6 +69,7 @@ export class EntityTextComponent extends TextMixin(
     | void {
     const text = this.$text();
     const link = this.$link();
+    console.log(link);
```
Sinucid: log

Sinucid (changes requested):

libs/platform/experience/entity-text/entity-text.component.ts
dunqan: dropped

libs/platform/experience/entity-text/entity-text.component.ts
dunqan: dropped

libs/platform/experience/entity-text/entity-text.component.ts
dunqan: we can, but I'm not sure we should do it in general, as it means, we are swallowing errors inside service, and we can't implement more complex handling for some specific cases outside of it

libs/platform/experience/entity-text/entity-text.component.ts
dunqan: that's what I'm beating with myself. Alternative is to have two computeds, less efficient (as you convert observable to computed, use it in another observable inside another computed) vs just one stream and one computed. @tobi-or-not-tobi any opinions?

tobi-or-not-tobi: equally not sure, but leaning towards the duplicate computed for now. I'd rather move the complexity elsewhere rather than polluting the component.

libs/platform/experience/entity-text/entity-text.component.ts
Sinucid: ok, then should we know exact what we are handling? Or are we handling any error? For the current case entity service can throw an error that means that some data is wrong and this is importatn, why should we catch it inside component then?

dunqan: In this component we just don't display anything if entity resolving is invalid (or not show the link) and that's the desired behavior here.

tobi-or-not-tobi (approved):

PR #1028, opened 2023-12-21
feat(core): local context fallback

dunqan:
DUPLICATE

Allows to resolve context fallback with current local element.

Additionaly:
- improved merchant context fallback resolving from local product context
- added sold by merchant to cart entry component


dunqan:
dplicate

PR #1031, opened 2023-12-22
refactor: Rename entity components to data

tobi-or-not-tobi:
Rename: 

- `oryx-entity-text` to `oryx-data-text`
- `oryx-entity-image` to `oryx-data-image`
- `oryx-entity-link` to `oryx-data-link`


libs/platform/experience/data-text/data-text.def.ts
```diff
@@ -0,0 +1,14 @@
+import { componentDef } from '@spryker-oryx/utilities';
+import { DataTextComponentOptions } from './data-text.model';
+
+declare global {
+  interface FeatureOptions {
+    'oryx-data-text'?: DataTextComponentOptions;
+  }
+}
+
+export const entityText = componentDef({
```
dunqan:
```suggestion
export const dataText = componentDef({
```

libs/platform/experience/src/feature.ts
```diff
@@ -12,8 +12,8 @@ import {
 export const experienceComponents = [
   components.compositionComponent,
   components.entityText,
```
dunqan:
```suggestion
  components.dataText,
```

dunqan (changes requested):

dunqan (approved):

PR #1041, opened 2024-01-02
chore: remove CMS content

tobi-or-not-tobi:
Some demo content accidentally sneaked into our experience data (only visible if you had set up some CMS integration).

### DIFF
--- a/libs/template/presets/storefront/experience/pages/home-page.ts
+++ b/libs/template/presets/storefront/experience/pages/home-page.ts
@@ -24,27 +24,27 @@ export const homePage: ExperienceComponent = {
   },
   components: [
     featureVersion >= '1.2' ? { ref: 'header' } : {},
-    {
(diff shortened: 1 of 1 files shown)


libs/template/presets/storefront/experience/pages/home-page.ts
```diff
@@ -24,27 +24,27 @@ export const homePage: ExperienceComponent = {
...
-          type: 'oryx-data-text',
-          options: { field: 'content' },
-        },
-        {
-          type: 'oryx-data-image',
-          options: { field: 'picture', rules: [] },
-        },
-      ],
-    },
+    // {
```
Sinucid: If this is still important to keep, leave a comment, please, why this is commented

Sinucid (changes requested):

Sinucid (approved):

PR #1043, opened 2024-01-02
feat(ui): accessible collapsible component

tobi-or-not-tobi:
We've added a clear outline to the summary of the collapsible components. 

closes: [HRZ-90969](https://spryker.atlassian.net/browse/HRZ-90969)

[HRZ-90969]: https://spryker.atlassian.net/browse/HRZ-90969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/base/ui/form/checkbox/src/checkbox.styles.ts
+++ b/libs/base/ui/form/checkbox/src/checkbox.styles.ts
@@ -1,6 +1,6 @@
 import { featureVersion } from '@spryker-oryx/utilities';
 
-import { css } from 'lit';
+import { css, unsafeCSS } from 'lit';
 
 const focusStyles =
   featureVersion >= '1.4'
@@ -11,7 +11,7 @@ const focusStyles =
           border-radius: 1px;
         }
       `
-    : (`` as any);
+    : unsafeCSS(``);
 
 export const checkboxStyles = css`
   :host([disabled]) {
--- a/libs/base/ui/form/checkbox/src/checkbox.styles.ts
+++ b/libs/base/ui/form/checkbox/src/checkbox.styles.ts
@@ -1,4 +1,17 @@
+import { featureVersion } from '@spryker-oryx/utilities';
+
 import { css } from 'lit';
+
+const focusStyles =
+  featureVersion >= '1.4'
+    ? css`
+        ::slotted(input:focus-visible)::after {
+          outline: solid 1px var(--oryx-color-focus);
+          outline-offset: 4px;
+          border-radius: 1px;
+        }
+      `
+    : (`` as any);
 
 export const checkboxStyles = css`
   :host([disabled]) {
@@ -30,6 +43,8 @@ export const checkboxStyles = css`
     inset-block-start: 0;
     inset-inline-start: 0;
   }
+
+  ${focusStyles}
 
   ::slotted(input:not(:disabled))::after {
     cursor: pointer;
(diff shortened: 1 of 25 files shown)


libs/base/ui/structure/collapsible/src/collapsible.component.ts
```diff
@@ -1,20 +1,22 @@
...
+import { collapsibleStyles } from './collapsible.styles';
 import { collapsibleBaseStyle } from './styles';
 
 export class CollapsibleComponent
   extends I18nMixin(LitElement)
   implements CollapsibleAttributes
 {
-  static styles = [collapsibleBaseStyle];
+  static styles =
+    featureVersion >= '1.4' ? collapsibleStyles : [collapsibleBaseStyle];
```
Sinucid, line 19: Is this a new convention?

libs/base/ui/structure/collapsible/src/styles/themes/storefront.styles.ts
```diff
@@ -34,10 +34,9 @@ const blockAppearance = css`
     box-shadow: var(--oryx-elevation-0) var(--oryx-color-elevation);
   }
 
-  :host(${blockSelector}:active),
-  :host(${blockSelector}[open]) {
+  /* :host(${blockSelector}:active) {
```
Sinucid: left over?

libs/base/ui/structure/collapsible/src/styles/themes/storefront.styles.ts
```diff
@@ -47,10 +46,19 @@ const blockAppearance = css`
...
   */
-  summary:focus-visible::after {
-    box-shadow: var(--oryx-box-shadow-focus);
+
+  summary:focus-visible {
+    outline: solid 1px var(--oryx-color-focus);
   }
 
+  :host([open]) summary:focus-visible {
+    /* outline-offset: -4px; */
```
Sinucid: The same question

Sinucid (changes requested):

libs/base/ui/form/checkbox/src/checkbox.styles.ts
```diff
@@ -1,5 +1,18 @@
...
+const focusStyles =
+  featureVersion >= '1.4'
+    ? css`
+        ::slotted(input:focus-visible)::after {
+          outline: solid 1px var(--oryx-color-focus);
+          outline-offset: 4px;
+          border-radius: 1px;
+        }
+      `
+    : (`` as any);
```
Sinucid:
```suggestion
    :  css``;
```

Sinucid (changes requested):

libs/base/ui/structure/collapsible/src/collapsible.component.ts
tobi-or-not-tobi: kind of, but not yet formalised. Few things that cross my mind when making changes For this PR:
- the way we've implemented themes is not so convenient. Say you introduce a new theme, you'd loose all styles for components that are theme driven (e.g. storefront vs backoffice). Going forward, I like our themes to only influence CSS variables. We could still add styles per theme, but the basic components should work properly. 
- keep it simple. I rather have single css file per component, and not an array. Complex CSS is likely an indicator  that the complex is too large anyway.

libs/base/ui/form/checkbox/src/checkbox.styles.ts
tobi-or-not-tobi: won't work as it throws `Unexpected empty source  no-empty-source` in stylelint. This only works if it holds style rules. I've tried many solutions and ended up with this one.

Sinucid (dismissed):

Sinucid (approved):

PR #1044, opened 2024-01-02
fix(ui): notifications's i18n content support

Sinucid:
**i18n** directive uses signals under the hood, that requires first initialisation of the value (usually it is undefined). This handled via life-circle hooks inside components, but for cases like notification center where content for notification is passed to the **push** method inside config, the signal is not re-init the value. To overcome such limitation additional support of I18nContent was added to the ui notification center

- `NotificationContent` interface now supports `I18nContent` ({token?: string, context?: InferI18nContext, raw?: string})
- `I18nContent` interface was added to the '@spryker-oryx/utilities'
- `TextResource` interface of `@spryker-oryx/i18n` package was deprecated. Use `I18nContent` instead
- `Notification` and `NotificationContent` interfaces were moved from `@spryker-oryx/ui/notification` to `'@spryker-oryx/ui/notification-center'`, and deprecated for NotificationComponent

closes: [HRZ-90684](https://spryker.atlassian.net/browse/HRZ-90684)

[HRZ-90684]: https://spryker.atlassian.net/browse/HRZ-90684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/utilities/src/injectables/i18n/i18n.injectable.ts
```diff
@@ -91,3 +91,8 @@ type InferI18nFilterName<F> = F extends I18nContextFilter<infer N, object>
 type InferI18nFilterConfig<F> = F extends I18nContextFilter<string, infer C>
   ? C
   : never;
+
+export interface I18nContent {
```
dunqan: what about possibility to handle all cases with I18nContent, so also raw text (that won't be translated)? something like `raw?: string` ?

libs/base/ui/overlays/notification-center/src/notification-center.component.ts
```diff
@@ -43,6 +45,16 @@ export class NotificationCenterComponent
     `;
   }
 
+  protected resolveContent(
```
dunqan: what about supporting content directly inside i18n directive?

libs/base/utilities/src/injectables/i18n/i18n.injectable.ts
Sinucid: Nice idea. Tbh, we've already had TextResource interface, I forgot about it, I deprecated it and it should be replaced  with I18nContent

libs/base/ui/overlays/notification-center/src/notification-center.component.ts
Sinucid: Added it to the i18nmixin, due to it is components-related stuff

libs/base/utilities/src/directives/i18n/i18n.mixin.ts
```diff
@@ -34,6 +35,25 @@ export function I18nMixin<T extends Type<ReactiveElement>>(
     i18n(...args: Parameters<typeof i18n>) {
       return i18n(...args);
     }
+
+    i18nContent<T extends I18nContent | unknown, A = unknown>(
```
dunqan: I meant deeper integration. 
i18n directive takes token that extends string as first parameter currently. 
It could transparently handle cases, where first (and only) parameter is a I18nContent, so this file would stay as it is, and the change will be in directive itself..

libs/platform/i18n/src/lib/models/text.model.ts
```diff
@@ -1,5 +1,9 @@
 import { I18nContext } from '@spryker-oryx/utilities';
 
+/**
+ * @deprecated since 1.4. Use I18nContent interface
```
dunqan, line 4: why not use TextResource if we already have it? 🤔

libs/base/utilities/src/injectables/i18n/i18n.injectable.ts
dunqan: forgot about it 🤦:)

libs/platform/i18n/src/lib/models/text.model.ts
Sinucid: 1) Due to managing of the dependancies. For current case ui domain (notification center) should not import from platform 
2) It will be still related to i18n, but to the its utilities part, and will be much more logical and convenient to keep it together with directive

libs/base/utilities/src/injectables/i18n/i18n.injectable.ts
```diff
@@ -91,3 +91,13 @@ type InferI18nFilterName<F> = F extends I18nContextFilter<infer N, object>
 type InferI18nFilterConfig<F> = F extends I18nContextFilter<string, infer C>
   ? C
   : never;
+
+export type I18nContent<T extends string | readonly string[] = string> =
+  RequireAtLeastOneProp<
```
dunqan: Just wondering if we couldn't solve this with simple union type?
(values are optional but makes sense only when token is used, and you probably shouldn't use token if raw is used)

dunqan (dismissed):

dunqan (dismissed):

dunqan (approved):

PR #1049, opened 2024-01-04
feat(ui): persisted collapsible open state

tobi-or-not-tobi:
We've added persisted state for collapsible components, which are used inside the facet navigation as well as in the merchant product offers on the PDP. 

closes: [HRZ-90984](https://spryker.atlassian.net/browse/HRZ-90984)

[HRZ-90984]: https://spryker.atlassian.net/browse/HRZ-90984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/base/ui/structure/collapsible/collapsible.component.spec.ts
```diff
@@ -121,4 +121,67 @@ describe('CollapsibleComponent', () => {
...
+      mockFeatureVersion('1.4');
+      element = await fixture(
+        html`<oryx-collapsible syncKey="foo"></oryx-collapsible>`
+      );
+      details = element.shadowRoot?.querySelector('details');
+      element.scrollIntoView = vi.fn();
+    });
+
+    it('should not load the collapsible state from sessionStorage', () => {
+      expect(globalThis.localStorage.getItem).not.toHaveBeenCalled();
```
Sinucid: Are you sure that this is working as expected? If I'm not mistaken need to store reference to the SpyInstance and check its calls, ie:
```
const spy = vi.spyOn(globalThis.sessionStorage.__proto__, 'getItem');
expect(spy).not.toHaveBeenCalled();
```

libs/base/ui/structure/collapsible/collapsible.component.spec.ts
```diff
@@ -121,4 +121,67 @@ describe('CollapsibleComponent', () => {
       expect(element).toContainElement('summary[tabindex="-1"]');
     });
   });
+
+  describe('when a syncKey is provided', () => {
+    vi.spyOn(globalThis.sessionStorage.__proto__, 'getItem');
```
Sinucid: And just wondering, maybe we need to create some service that will persist users actions?

libs/domain/search/facet-value-navigation/facet-value-navigation.component.ts
```diff
@@ -35,9 +38,14 @@ export class SearchFacetValueNavigationComponent
       this.enableClear &&
       (featureVersion >= '1.2' ? this.dirty : this.selectedLength);
 
+    const syncKey = this.$options().syncCollapsibleState
+      ? this.heading?.replace(' ', '-').toLowerCase()
```
Sinucid: Not the best option, heading can change according the locale, need something more stable, for example let component know about the facet's param

libs/platform/experience/src/services/layout/plugins/types/collapsible/collapsible.model.ts
```diff
@@ -6,10 +6,14 @@ declare global {
...
-export interface CollapsibleLayoutProperties {
-  collapsibleOpen?: boolean;
+  export interface LayoutProperty {
+    collapsibleOpen?: boolean;
+    /**
+     * The sync key to use for this collapsible. If set, the collapsible open state
+     * will be stored in the session storage and restored when the component is used
+     * during the same session.
+     */
+    collapsibleSyncKey?: string;
```
Sinucid: What about
```suggestion
    collapsibleStateKey?: string;
```

libs/platform/experience/src/services/layout/plugins/types/collapsible/collapsible.plugin.ts
```diff
@@ -27,10 +27,12 @@ export class CollapsibleLayoutPlugin implements LayoutPlugin {
     if (data.isComposition) return of();
 
     return of({
-      outer: html`<oryx-collapsible ?open=${data?.options?.collapsibleOpen}
-        >${renderLabelSlot(data, 'heading', -1)}
-        ${data.template}</oryx-collapsible
-      >`,
+      outer: html`<oryx-collapsible
+        ?open=${data?.options?.collapsibleOpen}
+        .syncKey=${data.options.collapsibleSyncKey}
```
Sinucid: The same, what about
```suggestion
        .stateKey=${data.options.collapsibleSyncKey}
```
or
```suggestion
        .persistatedStateKey=${data.options.collapsibleSyncKey}
```

Sinucid (changes requested):

libs/base/ui/structure/collapsible/collapsible.component.spec.ts
tobi-or-not-tobi: Seems to work, see other tests where we use `expect(globalThis.localStorage.setItem).toHaveBeenCalledWith(`

libs/base/ui/structure/collapsible/collapsible.component.spec.ts
tobi-or-not-tobi: We do have a storage service, and like to use it here, but we're in a lower level UI component. hence we can't use that service so easily.

libs/base/ui/structure/collapsible/collapsible.component.ts
```diff
@@ -21,10 +21,15 @@ export class CollapsibleComponent
   @property({ reflect: true }) appearance = CollapsibleAppearance.Block;
   @property({ type: Boolean, reflect: true }) open?: boolean;
   @property() heading?: string;
+  @property() persistedStateKey?: string;
   @property({ type: Boolean }) nonTabbable?: boolean;
 
   @query('details') protected details?: HTMLDetailsElement;
 
+  protected firstUpdated(): void {
+    if (!isServer) this.syncState();
```
Sinucid: Btw, this can produce some BCs, better to guard this:
```suggestion
    if (featureVetsion >= '1.4' && !isServer) this.syncState();
```

libs/platform/experience/src/services/layout/plugins/types/collapsible/collapsible.model.ts
```diff
@@ -6,10 +6,13 @@ declare global {
   export interface Layouts {
     collapsible: undefined;
   }
-  // eslint-disable-next-line @typescript-eslint/no-empty-interface
-  export interface LayoutProperty extends CollapsibleLayoutProperties {}
-}
 
-export interface CollapsibleLayoutProperties {
```
Sinucid, line 13: Did we release this interface already, dunno BC or not?

Sinucid (changes requested):

libs/platform/experience/src/services/layout/plugins/types/collapsible/collapsible.model.ts
tobi-or-not-tobi: no this was added earlier this week

libs/base/ui/structure/collapsible/collapsible.component.ts
tobi-or-not-tobi: it's guarded in the method (`if (!this.persistedStateKey) return;`), so I didn't see an issue. It's a new feature that you can use. I'll double check if it's guarded insight the facet components. For the product offer and the new layout, even if you do not opt-in to latest, you should still be able to use it.

libs/base/ui/structure/collapsible/collapsible.component.ts
```diff
@@ -67,8 +77,26 @@ export class CollapsibleComponent
     }
   }
 
-  protected onClick(): void {
-    this.isManuallyOpened = true;
+  protected syncState(store = false): void {
+    if (!this.persistedStateKey) return;
+
+    const uiStorageKey = 'ui';
```
dunqan: this key is too generic for a low level framework component

dunqan (changes requested):

Sinucid (dismissed):

libs/base/ui/structure/collapsible/collapsible.component.ts
```diff
@@ -67,8 +77,26 @@ export class CollapsibleComponent
...
+    const uiState = JSON.parse(sessionStorage.getItem(uiStorageKey) ?? '{}');
+
+    const collapsibleStateKey = 'collapsible';
+    const collapsibleState = uiState[collapsibleStateKey] || {};
+
+    if (store) {
+      if (this.isManuallyOpened) {
+        collapsibleState[this.persistedStateKey] = this.open;
+        uiState[collapsibleStateKey] = collapsibleState;
+        sessionStorage.setItem(uiStorageKey, JSON.stringify(uiState));
```
dunqan: this can still be undefined or throw errors in some other cases than server side (which we did check for)

dunqan (changes requested):

libs/base/ui/structure/collapsible/collapsible.component.ts
tobi-or-not-tobi: `oryx-ui`?

dunqan: what about `oryx-ui.collapsible` as a key? instead of serializing/deserializing one big object for every possible future UI component (may grow)?

tobi-or-not-tobi: changed to `oryx-collapsible`

libs/base/ui/structure/collapsible/collapsible.component.ts
```diff
@@ -67,8 +77,27 @@ export class CollapsibleComponent
...
-  protected onClick(): void {
-    this.isManuallyOpened = true;
+  protected syncState(store = false): void {
+    const storage = globalThis.sessionStorage;
+    if (!storage || !this.persistedStateKey) return;
+
+    const uiStorageKey = 'oryx-collapsible';
+    const collapsibleState = JSON.parse(storage.getItem(uiStorageKey) ?? '{}');
+
+    // const collapsibleStateKey = 'collapsible';
```
Sinucid: left over?

libs/domain/search/facet-range/facet-range.component.ts
```diff
@@ -122,10 +123,11 @@ export class SearchRangeFacetComponent
...
     return html`<oryx-search-facet-value-navigation
-      ?open="${this.open}"
-      ?enableClear="${!this.disableClear}"
-      ?dirty="${this.$isDirty()}"
-      .heading="${this.name}"
+      ?open=${this.open}
+      ?enableClear=${!this.disableClear}
+      ?dirty=${this.$isDirty()}
+      .heading=${this.name}
+      .key=${featureVersion >= '1.4' ? facet.parameter : undefined}
```
Sinucid, line 130: What about using ifDefined directive to make set of attributes a bit cleaner?

`key=${ifDefined(featureVersion >= '1.4' ? facet.parameter : undefined)}`

libs/domain/search/facet-rating/facet-rating.component.ts
```diff
@@ -82,6 +82,7 @@ export class SearchRatingFacetComponent
       ?enableClear="${!this.disableClear}"
       ?dirty=${this.$isDirty()}
       .heading=${this.name}
+      .key=${featureVersion >= '1.4' ? facet.parameter : undefined}
```
Sinucid, line 85: ifDefined?

libs/domain/search/facet/facet.component.ts
```diff
@@ -58,6 +57,7 @@ export class SearchFacetComponent extends I18nMixin(LitElement) {
       .heading=${this.name}
       .selectedLength=${selectedLength}
       .valuesLength=${valuesLength}
+      .key=${featureVersion >= '1.4' ? facet.parameter : undefined}
```
Sinucid, line 60: ifDefined?

Sinucid (changes requested):

libs/domain/search/facet/facet.component.ts
tobi-or-not-tobi: not needed. This is part of property binding anyways, and the attribute is not reflected.

libs/domain/search/facet-range/facet-range.component.ts
tobi-or-not-tobi: how is it cleaner?

libs/base/ui/src/controllers/ui-state.controller.ts
```diff
@@ -0,0 +1,26 @@
...
+export class UiStateController implements ReactiveController {
+  /**
+   * The open state is kept in memory.
+   */
+  static state: { [key: string]: unknown } = {};
+
+  hostConnected?(): void;
+
+  constructor(protected host: LitElement) {
+    this.host.addController(this);
```
dunqan, line 12: I'd drop this line and hostConnected callback, if we are not using it.

libs/base/ui/structure/collapsible/collapsible.component.ts
```diff
@@ -67,8 +80,19 @@ export class CollapsibleComponent
...
   }
 
-  protected onClick(): void {
-    this.isManuallyOpened = true;
+  protected syncState(store = false): void {
+    if (!this.persistedStateKey) return;
+
+    if (store) {
+      if (this.isManuallyOpened) {
+        this.uiController.set(this.persistedStateKey, this.open);
```
dunqan, line 101: shouldn't we prefix key with component name? as the controller could be used in different components, so there is a risk of key collision? (or developer has to care about it)

libs/base/ui/structure/collapsible/collapsible.model.ts
```diff
@@ -21,6 +21,12 @@ export interface CollapsibleAttributes {
    * Prevent focus by keyboard navigation
    */
   nonTabbable?: boolean;
+
+  /**
+   * The key to use for storing the open state. If set, the open state is stored
+   * in session storage and restored when the component is used in the same session.
```
dunqan:
```suggestion
   * in memory and restored when the component is used in the same session.
```

libs/base/ui/src/controllers/ui-state.controller.ts
tobi-or-not-tobi: we start using the host to resolve the component name.

dunqan (approved):

Sinucid (approved):

PR #1054, opened 2024-01-05
feat(cart): remove cart

Sinucid:
Remove cart feature that allow user to delete any cart, that assigned to its own account. The feature is available as a button inside cart-item on carts list page. The confirmation message is shown before deletion. `CartRemoveComponent` with implementations was added to the **multi-cart** extension of cart domain

closes: [HRZ-90661](https://spryker.atlassian.net/browse/HRZ-90661)

[HRZ-90661]: https://spryker.atlassian.net/browse/HRZ-90661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/cart/cart-remove/cart-remove.component.ts
```diff
@@ -0,0 +1,92 @@
...
+    this.isLoading = true;
+
+    const cart = this.$cart();
+    const name = cart?.name;
+    const cartId = cart?.id;
+
+    this.cartService.deleteCart({ cartId }).subscribe({
+      next: () => {
+        this.notificationService.push({
+          type: AlertType.Error,
```
tobi-or-not-tobi:
```suggestion
          type: AlertType.Success,
```

libs/domain/cart/cart-remove/cart-remove.component.ts
```diff
@@ -0,0 +1,92 @@
...
+        .icon=${IconTypes.Trash}
+        @click=${this.onRemove}
+      ></oryx-button>
+      ${this.renderRemoveConfirmationModal()}
+    `;
+  }
+
+  protected onRemove(): void {
+    this.requestsConfirmation = true;
+  }
```
tobi-or-not-tobi:
```suggestion
   protected onRemove(e: PointerEvent): void {
    if (e.altKey) {
      this.onConfirm();
    } else {
      this.requestsConfirmation = true;
    }
  }
```

What do you think about this? It also might inspire to rename `requestsConfirmation` to just `remove`

libs/domain/cart/cart-remove/cart-remove.component.ts
```diff
@@ -0,0 +1,92 @@
...
+        .size=${ButtonSize.Md}
+        .text=${this.i18n('carts.remove.button.remove')}
+        ?loading=${this.isLoading}
+        @click=${this.onConfirm}
+      ></oryx-button>
+    </oryx-modal>`;
+  }
+
+  protected onConfirm(): void {
+    this.isLoading = true;
```
tobi-or-not-tobi:
```suggestion
    this.isRemoving = true;
```

Seems more accurate. It could also be used to start removing the row (animation).

libs/domain/cart/cart-remove/index.ts
```diff
@@ -0,0 +1 @@
+export * from './cart-remove.component';
```
tobi-or-not-tobi, line 1: can we add a schema for this component?

libs/domain/user/address-remove/address-remove.component.ts
```diff
@@ -41,6 +41,7 @@ export class UserAddressRemoveComponent extends I18nMixin(
       open
       .heading=${this.i18n('checkout.address.remove-address')}
       @oryx.close=${this.onClose}
+      @oryx.modal.closed=${this.onClose}
```
tobi-or-not-tobi, line 44: what is this?

libs/domain/cart/cart-remove/stories/demo.stories.ts
```diff
@@ -0,0 +1,29 @@
...
+    customElements.whenDefined('oryx-cart-remove'),
+    customElements.whenDefined('oryx-button'),
+  ]);
+
+  const element =
+    obj.canvasElement.querySelector<LitElement>('oryx-cart-remove');
+  console.log(element?.renderRoot.querySelector<LitElement>('oryx-button'));
+
+  element?.renderRoot.querySelector<LitElement>('oryx-button')?.click();
+};
```
tobi-or-not-tobi: nothing happens locally in my storybook

tobi-or-not-tobi (changes requested):

libs/domain/user/address-remove/address-remove.component.ts
Sinucid: A fix. @tolerants changed events's names for modal since >= 1.4, and oryx.close event is deprecated. Probably he forgot to fix this on some components, me as well. Added new events for all modals for backward compatibility

libs/domain/cart/cart-remove/cart-remove.component.ts
Sinucid: then `inProgress` will be even more clear

libs/domain/cart/cart-remove/cart-remove.component.ts
Sinucid: Nice idea with instant remove, but not sure about renaming of state prop, in general flow we are **requesting confirmation** (ie show the modal), it is not yet exact remove

libs/domain/cart/cart-remove/cart-remove.component.ts
```diff
@@ -34,8 +34,14 @@ export class CartRemoveComponent extends CartComponentMixin(
     `;
   }
 
-  protected onRemove(): void {
-    this.requestsConfirmation = true;
+  protected onRemove(e: PointerEvent): void {
+    console.log(e.altKey);
```
tobi-or-not-tobi: left over

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #1055, opened 2024-01-08
feat: Added dropdown behaviour on hover

tobi-or-not-tobi:
Some dropdown components require the popover to show on hover (and focus) rather than click. For example, when the trigger elements contains a link. 

closes: [HRZ-90988](https://spryker.atlassian.net/browse/HRZ-90988)

[HRZ-90988]: https://spryker.atlassian.net/browse/HRZ-90988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/base/ui/overlays/dropdown/dropdown.component.spec.ts
+++ b/libs/base/ui/overlays/dropdown/dropdown.component.spec.ts
@@ -63,22 +63,32 @@ describe('DropdownComponent', () => {
       expect(element).not.toContainElement('oryx-popover[show]');
     });
 
-    describe('and mouseover event dispatched', () => {
+    describe('and mouseenter event dispatched', () => {
       beforeEach(async () => {
-        element.dispatchEvent(new MouseEvent('mouseover'));
+        element.dispatchEvent(new MouseEvent('mouseenter'));
       });
 
       it('should show the popover', async () => {
         expect(element).toContainElement('oryx-popover[show]');
       });
 
-      describe('and mouseout event dispatched', () => {
+      describe('and mouseleave event dispatched', () => {
         beforeEach(async () => {
-          element.dispatchEvent(new MouseEvent('mouseout'));
+          element.dispatchEvent(new MouseEvent('mouseleave'));
         });
 
         it('should not yet hide the popover', async () => {
           expect(element).toContainElement('oryx-popover[show]');
+        });
+
+        describe('but when 99ms have passed', () => {
+          beforeEach(async () => {
+            await new Promise((resolve) => setTimeout(resolve, 99));
+          });
+
+          it('should still show the popover', async () => {
+            expect(element).toContainElement('oryx-popover[show]');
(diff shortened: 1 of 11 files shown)


libs/base/ui/overlays/dropdown/dropdown.component.spec.ts
```diff
@@ -52,6 +52,48 @@ describe('DropdownComponent', () => {
     });
   });
 
+  describe('when openOnHover property is assigned', () => {
+    beforeEach(async () => {
+      element = await fixture(
+        html`<oryx-dropdown openOnHover></oryx-dropdown>`
```
Sinucid, line 58: maybe just **hover**?
```suggestion
        html`<oryx-dropdown hover></oryx-dropdown>`
```

libs/base/ui/overlays/dropdown/dropdown.component.spec.ts
```diff
@@ -52,6 +52,48 @@ describe('DropdownComponent', () => {
...
+          element.dispatchEvent(new MouseEvent('mouseout'));
+        });
+
+        it('should not yet hide the popover', async () => {
+          expect(element).toContainElement('oryx-popover[show]');
+        });
+
+        describe('but when 100ms have passed', () => {
+          beforeEach(async () => {
+            await new Promise((resolve) => setTimeout(resolve, 100));
```
Sinucid: use vitest fake timers, please

libs/base/ui/overlays/dropdown/dropdown.model.ts
```diff
@@ -44,8 +44,15 @@ export interface DropdownProperties {
...
 
+  /**
+   * Indicates that the popover will be opened on hover. The keyboard user will have a similarly
+   * experience when the focus is moved to the trigger element.
+   */
+  openOnHover?: boolean;
+
   /**
    * Indicates that the popover will be opened on focus.
+   * @deprecated since version 1.4 use openOnHover instead
```
Sinucid, line 55: Why? It is different behaviour

Sinucid (changes requested):

libs/base/ui/overlays/dropdown/dropdown.model.ts
tobi-or-not-tobi: same, but a better name imo. But if we have a name that covers both hover and focus, we could go with that.

libs/base/ui/overlays/dropdown/dropdown.component.spec.ts
tobi-or-not-tobi: hover sounds like an event, rather than a configuration to react to such event.

Sinucid (approved):

PR #1057, opened 2024-01-08
feat(site): Added json-ld for product data

tobi-or-not-tobi:
Added json-ld for product and product offers. 

Product data contains the basic product data and prices. The `AggregateOffer` is provided when the merchant package is installed. 

closes [HRZ-91023](https://spryker.atlassian.net/browse/HRZ-91023)

[HRZ-91023]: https://spryker.atlassian.net/browse/HRZ-91023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/site/src/services/jsonld/jsonld.normalizer.ts
```diff
@@ -0,0 +1,21 @@
+import { Observable } from 'rxjs';
+import { JSONLD } from './jsonld.model';
+
+export interface JsonLdNormalizer<T = any> {
```
dunqan, line 4: What about reusing existing transformer infra for it (what we use for adapter normalizers)?

dunqan (approved):

PR #1058, opened 2024-01-09
test(storefront): visual regression tests + percy integration

Sinucid:
closes [HRZ-3008](https://spryker.atlassian.net/browse/HRZ-3008)

[HRZ-3008]: https://spryker.atlassian.net/browse/HRZ-3008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Ivan-Vidiaiev (changes requested):
Still need to fix Checkout page snapshots and Cart page snapshots

apps/storefront-e2e/src/support/page-objects/landing.page.ts
```diff
@@ -8,7 +8,7 @@ export class LandingPage extends AbstractSFPage {
   }
 
   waitForLoaded(): void {
-    cy.wait('@searchQuery');
+    // cy.wait('@searchQuery');
```
AndreyYevtukhov: if it is not used -> we should remove it completely

apps/storefront-e2e/src/integration/ssr-navigation.cy.ts
```diff
@@ -18,96 +19,108 @@ const footer = new FooterFragment();
 const header = new HeaderFragment();
 const searchbox = new SearchBoxFragment();
 
-describe('SSR suite', { tags: 'smoke' }, () => {
-  if (Cypress.env('isSSR')) {
+const visitOptions = {
+  onBeforeLoad: (win) => {
+    delete win.CSSStyleSheet.prototype.replace;
+  },
+};
```
AndreyYevtukhov, line 26: What about moving it inside https://github.com/spryker/oryx/pull/1058/files#diff-b4224dff4acb276b0959d32afa1d0994ff4e9546b4a2affc96a9f8ea50a7d2afR17 and wrapping it with `isSSREnabled` https://github.com/spryker/oryx/pull/1058/files#diff-2dedf78e428841d9cb523c0d5a92802f7f2523c8895c283478278a1ad3c67760R29 check? 

or even with `isPercyEnabled` if it should be used only for Percy?

AndreyYevtukhov: If we move it into the abstract page class - `visit` calls will be much cleaner everywhere + we will not have to change the method signature

apps/storefront-e2e/src/support/page-objects/checkout.page.ts
```diff
@@ -56,4 +57,35 @@ export class CheckoutPage extends AbstractSFPage {
       .its('response.body.data.attributes.orderReference')
       .as('createdOrderId');
   };
+
+  templateIsReady = () => {
+    //email input is ready
+    this.checkoutAsGuestForm
+      .getEmailInput()
+      .should('be.visible', { timeout: 10000 });
```
AndreyYevtukhov, line 65: I'm not quite sure about these `, { timeout: 10000 }` checks everywhere, they are not 100% reliable and may cause failures

I suggest waiting for API requests to resolve (I'm almost 100% sure our UI is API-data-driven) and only then check the visibility.

By default Cypress wait for 30 seconds for API requests to resolve -> this will be enough for us to wait and then do visibility checks that take 4seconds max (bug 4 seconds should be enough to process response body and render the component)

apps/storefront-e2e/src/integration/ssr-navigation.cy.ts
Sinucid: This code should be called in very specific time, before page start loading scripts and other resources, and the only way that I found was `onBeforeLoad` hook of visit method. To merge this functionality with something else will be a bit tricky. And not all SSR runs require this hack, this is personally for PERCY

apps/storefront-e2e/src/support/page-objects/checkout.page.ts
AndreyYevtukhov: https://docs.cypress.io/api/commands/intercept

API interceptors are easy to create and manage + you can wait for multiple requests to resolve and only then continue https://docs.cypress.io/api/commands/wait#Aliases

AndreyYevtukhov (changes requested):
Looks good, but some parts can be improved. Please, check comments

apps/storefront-e2e/src/support/page-objects/checkout.page.ts
Sinucid: We need to ensure that all elements are properly rendered on the page, and unfortunately this does not depend on api calls. On cypress side checkout page (not locally) takes ~10sec to completely build the checkout page with all forms and elements after all request. Some other pages also had some flakines, but wait for 3s is enough for that.  I can predict that due to limited resources of remote machines.

Setting of timeout gives additional time to make sure that they are rendered, and usuakky this will be les then 10s, that usually will be much faster than just `cy.wait(10000)`

AndreyYevtukhov (approved):

Ivan-Vidiaiev (approved):
Tests look good

PR #1060, opened 2024-01-10
feat(product): Added product discontinued component

tobi-or-not-tobi:
Product discontinued component renders a message on the PDP that the product is no longer available. 

Additionally, the product is no longer indexed (robots meta tag) when the product is discontinued and stock is no longer available. 

closes: [HRZ-91033](https://spryker.atlassian.net/browse/HRZ-91033)
fixes: [HRZ-91042](https://spryker.atlassian.net/browse/HRZ-91042)

[HRZ-91033]: https://spryker.atlassian.net/browse/HRZ-91033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HRZ-91042]: https://spryker.atlassian.net/browse/HRZ-91042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/cart/add/src/add.component.ts
```diff
@@ -79,20 +79,35 @@ export class CartAddComponent extends ProductMixin(
   };
 
   protected $hasStock = computed(() => {
+    const product = this.$product();
+
+    if (product?.discontinued && !product?.availability?.quantity) return false;
```
Sinucid: What about wrap `product?.discontinued && !product?.availability?.quantity` into some getter to avoid duplication (isDiscontinued)? Maybe even on product mixin level

libs/domain/product/discontinued/stories/static.stories.ts
```diff
@@ -0,0 +1,24 @@
...
+export default {
+  title: `${storybookPrefix}/Discontinued`,
+};
+
+const Template: Story = (): TemplateResult => {
+  return html`
+    <h3>Discontinued without note</h3>
+    <oryx-product-discontinued sku="discontinued"></oryx-product-discontinued>
+
+    <h3>Discontinued wit note</h3>
```
Sinucid:
```suggestion
    <h3>Discontinued with note</h3>
```

libs/domain/product/src/models/product.model.ts
```diff
@@ -41,6 +41,8 @@ export interface Product {
   mediaSet?: ProductMediaSet[];
   labels?: ProductLabel[];
   availability?: ProductAvailability;
+  discontinued?: boolean;
+  discontinuedNote?: string;
```
Sinucid, line 45: should it support i18n messages?

Sinucid (changes requested):

libs/domain/product/src/models/product.model.ts
tobi-or-not-tobi: The api has only a single response, but when we request it with different language, it will respond a different message. No need to store various mesaages for each locale.

libs/domain/cart/add/src/add.component.ts
tobi-or-not-tobi: good one, it became a bit messy indeed. I've refactored the min/max logic and simplified it all.

libs/domain/cart/add/src/add.component.ts
```diff
@@ -78,31 +78,42 @@ export class CartAddComponent extends ProductMixin(
     if (this.$product()) this.input?.reset?.();
   };
 
-  protected $hasStock = computed(() => {
```
Sinucid, line 81: Sounds like BC

libs/domain/cart/add/src/add.component.ts
```diff
@@ -78,31 +78,42 @@ export class CartAddComponent extends ProductMixin(
...
+   * The available quantity is the sum of the product's availability quantity
+   * and the quantity of all entries in the cart with the same sku.
+   * If the product is discontinued, the available quantity is 0.
+   * If the product is never out of stock, the available quantity is Infinity.
+   */
+  protected $availableQuantity = computed(() => {
+    const { availability, sku, discontinued } = this.$product() ?? {};
+    const { quantity = 0, isNeverOutOfStock } = availability ?? {};
+
+    if ((discontinued && !quantity) || (!quantity && !isNeverOutOfStock)) {
```
Sinucid:
```suggestion
    if (!quantity && (discontinued || !isNeverOutOfStock)) {
```

Sinucid (changes requested):

Sinucid (approved):

PR #1066, opened 2024-01-12
chore: lazy-loading for merchant lib

dunqan:
closes: [HRZ-90974](https://spryker.atlassian.net/browse/HRZ-90974)

[HRZ-90974]: https://spryker.atlassian.net/browse/HRZ-90974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/domain/merchant/src/product/index.ts
```diff
@@ -1,3 +1,3 @@
+export * from '../../services/src/product/product.normalizer';
```
Sinucid:
```suggestion
export * from '@spryker-oryx/merchant/services';
```

Sinucid: And why do we need this re-exports?

tobi-or-not-tobi (dismissed):

libs/domain/merchant/src/product/index.ts
dunqan: left over, thanks!

Sinucid (approved):

PR #1069, opened 2024-01-15
fix(router): same routes duplication

Sinucid:
`navigate` method of the router service is pushing the state to the browser history on each call, even if it pushes the same route multiple times. The fix adds the check that prevent navigation when prev and current routes are the same

closes: [HRZ-90436](https://spryker.atlassian.net/browse/HRZ-90436)

[HRZ-90436]: https://spryker.atlassian.net/browse/HRZ-90436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/platform/router/src/services/default-router.service.ts
+++ b/libs/platform/router/src/services/default-router.service.ts
@@ -51,13 +51,7 @@ export class DefaultRouterService implements RouterService {
     const queryParams =
       extras?.queryParams ?? this.getURLSearchParams(url[1]) ?? {};
 
-    if (
-      this.router$.value === url[0] &&
-      JSON.stringify(this.urlSearchParams$.value) ===
-        JSON.stringify(queryParams)
-    ) {
-      return;
-    }
+    if (this.isSameRoute(route, extras)) return;
 
     this.storageService.set(
       PREVIOUS_PAGE,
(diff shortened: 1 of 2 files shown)


tolerants (dismissed):

dunqan (dismissed):

libs/platform/router/src/services/default-router.service.ts
```diff
@@ -51,13 +51,7 @@ export class DefaultRouterService implements RouterService {
...
       extras?.queryParams ?? this.getURLSearchParams(url[1]) ?? {};
 
-    if (
-      this.router$.value === url[0] &&
-      JSON.stringify(this.urlSearchParams$.value) ===
-        JSON.stringify(queryParams)
-    ) {
-      return;
-    }
+    if (this.isSameRoute(route, extras)) return;
```
tobi-or-not-tobi: we can move it up before initializing `queryParams`

libs/platform/router/src/services/default-router.service.ts
```diff
@@ -77,6 +71,8 @@ export class DefaultRouterService implements RouterService {
       return;
     }
 
+    if (this.isSameRoute(route)) return;
```
tobi-or-not-tobi: Can we do this before checking on not found? Not only more efficient but also more logical I can imagine.

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #1073, opened 2024-01-16
fix: adjust buckets parsing

supproduction:
adjust buckets parsing in order to show them in EB


libs/platform/experience/src/models/composition.model.ts
```diff
@@ -4,6 +4,8 @@ import { Breakpoint } from '@spryker-oryx/utilities';
 export interface CompositionProperties {
   rules?: StyleRuleSet[];
   context?: ContextValue & Record<string, unknown>;
+  bucket?: boolean;
+  bucketType?: 'main' | 'label';
```
tobi-or-not-tobi: could be any string

tobi-or-not-tobi (changes requested):

tobi-or-not-tobi (approved):

PR #1080, opened 2024-01-24
feat: glue package

dunqan:
Move all Glue specific adapters to `@spryker-oryx/glue` library


dunqan:
decided to take approach from #1083

PR #1085, opened 2024-02-14
AE-57 FOSS dependency

gerner-spryker:
- Added `package-aggregator` to enable storing all frontend package dependencies.
- Saved the current state of project frontend dependencies to `package-aggregated.txt`.

closes: https://spryker.atlassian.net/browse/AE-57 (https://release.spryker.com/release-groups/view/5294)


package-aggregator
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
tolerants, line 1: @gerner-spryker Why it should be a part of the repo and not external utility?

package-aggregated.txt
```diff
@@ -0,0 +1,159 @@
+@clack/prompts:^0.6.3
```
tolerants: It would be great to understand why we need it.

package-aggregated.txt
```diff
@@ -0,0 +1,159 @@
...
+@rollup/plugin-commonjs:^24.0.0
+@rollup/plugin-json:^6.0.0
+@rollup/plugin-node-resolve:^15.0.1
+@rollup/plugin-typescript:^11.0.0
+@spryker-oryx/application:1.4.0
+@spryker-oryx/auth:1.4.0
+@spryker-oryx/cart:1.4.0
+@spryker-oryx/checkout:1.4.0
+@spryker-oryx/content:1.4.0
+@spryker-oryx/core:*
```
tolerants: Do we really need a `*` in case we also have a specific version of the same package?

